PR #3, opened 2025-08-08
Implement server-side registry for Abilities API 

### LINKED ISSUE #1: Implement server-side registry for Abilities API
A typical WordPress site might have dozens of plugins, each adding unique functionality. But there’s no standardized way for these components to express their capabilities. An AI assistant has no systematic way to discover that your backup plugin can create snapshots, your SEO plugin can analyze content, or your e-commerce solution can generate reports. This fragmentation limits what we can build.

The Abilities API establishes a common pattern that the entire ecosystem can adopt. It may look something like this:

```php
// Any plugin can register its abilities with the `init` hook.
wp_register_ability( 'my-seo-plugin/analyze-content-seo', [
    'label'       => __( 'Analyze Content SEO', 'my-seo-plugin' ),
    'description' => __( 'Analyzes post content for SEO improvements.', 
'my-seo-plugin' ),
    'thinking_message' => __('Reviewing your content now!'),
    'success_message' => __('Content reviewed successfully.'),
    'execute_callback'    => [ 'MySEOPlugin', 'analyze_content' ],
    'input_schema'  => [
        'type' => 'object',
        'properties' => [
            'post_id' => [ 'type' => 'integer', 'description' => __( 'The post identifier.', 'my-seo-plugin' ), 'required' => true ],
        ],
        'additionalProperties' => false,
    ],
    'output_schema' => ]
        'type' => 'number',
        'description' => __( 'The score for the content in percentage.', 'my-seo-plugin' ),
        'required' => true,
    ],
    'permission_callback'  => 'edit_posts',
] );
```

More details in the announcement post:

https://make.wordpress.org/ai/2025/07/17/abilities-api/

See also Feature API as a reference:

- https://github.com/Automattic/wp-feature-api/blob/trunk/includes/wp-feature.php
- https://github.com/Automattic/wp-feature-api/blob/trunk/includes/class-wp-feature-registry.php
- https://github.com/Automattic/wp-feature-api/blob/trunk/includes/class-wp-feature.php

gziolo:
Closes https://github.com/WordPress/abilities-api/issues/1.

PR in WordPress core, which validates compatibility and whether tests pass:
- https://github.com/WordPress/wordpress-develop/pull/9410

### Registering an ability

```php
function example_register_add_numbers_ability() {
	wp_register_ability(
		'example/add-numbers',
		[
			'label'               => 'Add numbers',
			'description'         => 'Calculates the result of adding two numbers.',
			'input_schema'        => [
				'type'                 => 'object',
				'properties'           => [
					'a' => [
						'type'        => 'number',
						'description' => 'First number.',
						'required'    => true,
					],
					'b' => [
						'type'        => 'number',
						'description' => 'Second number.',
						'required'    => true,
					],
				],
				'additionalProperties' => false,
			],
			'output_schema'       => [
				'type'        => 'number',
				'description' => 'The result of adding the two numbers.',
				'required'    => true,
			],
			'execute_callback'    => function ( array $input ): int {
				return $input['a'] + $input['b'];
			},
			'permission_callback' => function (): bool {
				return true;
			},
			'meta'                => [
				'category' => 'math',
			],
		]
	);
}

add_action( 'abilities_api_init', 'example_register_add_numbers_ability' );
```

### Unregistering an ability

```php
wp_unregister_ability( 'example/add-numbers' );
```

### Retrieving a specific ability

```php
```

### Retrieving all abilities

```php
```

### Operations on abilities

```php
$ability->has_permission( [ 'a' => 2, 'b' => 3 ] ); // true
echo $ability->execute( [ 'a' => 2, 'b' => 3 ] ); // 5
```

### DIFF
--- a/src/abilities-api.php
+++ b/src/abilities-api.php
@@ -1,0 +1,87 @@
+<?php declare( strict_types = 1 );
+
+/**
+ * Abilities API
+ *
+ * Defines functions for managing abilities in WordPress.
+ *
+ * @package WordPress
+ * @subpackage Abilities API
+ * @since 0.1.0
+ */
+
+/**
+ * Registers a new ability using Abilities API.
+ *
+ * Note: Do not use before the {@see 'abilities_api_init'} hook.
+ *
+ * @see WP_Abilities_Registry::register()
+ *
+ * @since 0.1.0
+ *
+ * @param string|WP_Ability $name       The name of the ability, or WP_Ability instance. The name must be a string
+ *                                      containing a namespace prefix, i.e. `my-plugin/my-ability`. It can only
+ *                                      contain lowercase alphanumeric characters, dashes and the forward slash.
+ * @param array             $properties Optional. An associative array of properties for the ability. This should
+ *                                      include `label`, `description`, `input_schema`, `output_schema`,
+ *                                      `execute_callback`, `permission_callback`, and `meta`.
+ * @return ?WP_Ability An instance of registered ability on success, null on failure.
+ */
+function wp_register_ability( $name, array $properties = array() ): ?WP_Ability {
+	if ( ! did_action( 'abilities_api_init' ) ) {
+		_doing_it_wrong(
+			__FUNCTION__,
+			sprintf(
+				/* translators: 1: abilities_api_init, 2: string value of the ability name. */
+				esc_html__( 'Abilities must be registered on the %1$s action. The ability %2$s was not registered.' ),
+				'<code>abilities_api_init</code>',
+				'<code>' . esc_attr( $name ) . '</code>'
+			),
+			'0.1.0'
+		);
+		return null;
+	}
+
+	return WP_Abilities_Registry::get_instance()->register( $name, $properties );
+}
+
+/**
+ * Unregisters an ability using Abilities API.
+ *
+ * @see WP_Abilities_Registry::unregister()
+ *
+ * @since 0.1.0
+ *
+ * @param string $name The name of the registered ability, with its namespace.
+ * @return ?WP_Ability The unregistered ability instance on success, null on failure.
+ */
+function wp_unregister_ability( string $name ): ?WP_Ability {
+	return WP_Abilities_Registry::get_instance()->unregister( $name );
+}
+
+/**
+ * Retrieves a registered ability using Abilities API.
+ *
+ * @see WP_Abilities_Registry::get_registered()
+ *
+ * @since 0.1.0
+ *
+ * @param string $name The name of the registered ability, with its namespace.
+ * @return ?WP_Ability The registered ability instance, or null if it is not registered.
+ */
+function wp_get_ability( string $name ): ?WP_Ability {
+	return WP_Abilities_Registry::get_instance()->get_registered( $name );
+}
+
+/**
+ * Retrieves all registered abilities using Abilities API.
+ *
+ * @see WP_Abilities_Registry::get_all_registered()
+ *
+ * @since 0.1.0
+ *
+ * @return WP_Ability[] The array of registered abilities.
+ */
+function wp_get_abilities(): array {
+	return WP_Abilities_Registry::get_instance()->get_all_registered();
+}
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,24 @@
+# This file is for unifying the coding style for different editors and IDEs
+# editorconfig.org
+
+# WordPress Coding Standards
+# https://make.wordpress.org/core/handbook/coding-standards/
+
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = tab
+
+[*.yml]
+indent_style = space
+indent_size = 2
+
+[*.md]
+trim_trailing_whitespace = false
+
+[{*.txt,wp-config-sample.php}]
+end_of_line = crlf
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phpunit
+		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/9.2/phpunit.xsd"
+		bootstrap="tests/bootstrap.php"
+		backupGlobals="false"
+		colors="true"
+		beStrictAboutTestsThatDoNotTestAnything="true"
+		beStrictAboutOutputDuringTests="true"
+		convertErrorsToExceptions="true"
+		convertWarningsToExceptions="true"
+		convertNoticesToExceptions="true"
+		convertDeprecationsToExceptions="true"
+		>
+    <testsuites>
+        <testsuite name="unit">
+            <directory suffix="Test.php">tests/unit</directory>
+        </testsuite>
+    </testsuites>
+</phpunit>
--- a/src/class-wp-abilities-registry.php
+++ b/src/class-wp-abilities-registry.php
@@ -232,7 +232,16 @@ final class WP_Abilities_Registry {
 	 * @return WP_Ability|null The registered ability instance, or null if it is not registered.
 	 */
 	public function get_registered( $name ): ?WP_Ability {
-		return $this->registered_abilities[ $name ] ?? null;
+		if ( ! $this->is_registered( $name ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				/* translators: %s: Ability name. */
+				sprintf( esc_html__( 'Ability "%s" not found.' ), esc_attr( $name ) ),
+				'0.1.0'
+			);
+			return null;
+		}
+		return $this->registered_abilities[ $name ];
 	}
 
 	/**
--- a/src/class-wp-abilities-registry.php
+++ b/src/class-wp-abilities-registry.php
@@ -29,7 +29,7 @@ final class WP_Abilities_Registry {
 	 * Container for the main instance of the class.
 	 *
 	 * @since 0.1.0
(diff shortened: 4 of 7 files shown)


gziolo:
The proposed initial implementation for server-side registry with testing code coverage is ready to receive the initial feedback. What's still missing is the unit tests configuration in a standalone repository, as I developed and tested everything with the WordPress core testing infrastructure available.

gziolo:
> The proposed initial implementation for server-side registry with testing code coverage is ready to receive the initial feedback. What's still missing is the unit tests configuration in a standalone repository, as I developed and tested everything with the WordPress core testing infrastructure available.

To better illustrate that I opened a PR in WordPress Core to ensure all the CI checks pass and all unit tests work:
- https://github.com/WordPress/wordpress-develop/pull/9410

CI for WP core still validates against PHP 7.2 and 7.3, so that might enforce some syntax changes as these jobs fail ... It looks like unit tests pass on 7.4 and higher as expected.

PHPCS doesn't like the followign things:
- Typed properties are not supported in PHP 7.3 or earlier
- Short array syntax is not allowed
- Trailing comma's are not allowed in function calls in PHP 7.2 or earlier

I'm inclined to address these issues to avoid a compatibility mismatch.

justlevine:
> I'm inclined to address these issues to avoid a compatibility mismatch.

IMHO dont waste the time rn

PHP 7.4 is coming to core in either 6.9 or 7.0, so either way in time for this to be merged, but more important once I set up vipcs locally most these code standard fails will autofix.

src/abilities-api.php
justlevine, line 1: Would recommend using `inc/` instead of `src/`. I've got a feeling we're gonna need use some `@wordpress/*` packages that need to be built down the road.

justlevine:
Should we use a namespace `WordPress/Abilities` instead of the root `WP_*` ? Gutenberg doesnt, Performance does. 

My gut says we _should_, because like GB our goal here is for experimentation/extenders and we don't want to have to worry about breaking back-compat when we merge into core. (Unlike individual performance plugins which arent meant to be extended or rly used in 3rd-party code

src/abilities-api.php
gziolo: Initially, I wanted to replicate https://github.com/WordPress/php-ai-client and went with `src` for PHP, but I agree that it's hard to use one pattern because we expect both PHP and JS code in this repo. Gutenberg uses `packages` and `lib`. Here we also need to take into account a need for a composer package, and an npm package. It isn't so simple to pick right folder names 😅

gziolo:
> Should we use a namespace WordPress/Abilities instead of the root WP_* ? Gutenberg doesnt, Performance does.
>
> My gut says we _should_, because like GB our goal here is for experimentation/extenders and we don't want to have to worry about breaking back-compat when we merge into core. 

Gutenberg doesn't use namespaces, so I'm a little bit confused. Did you mean we shouldn't? 

Overall, in WP core you can find some libs like SimplePie, PhpMailer, Sodium that use namespaces, but they were forked from preexisting projects. I don't know whether there is enough support to use namespaces for other parts of codebase in WP core. From the composer package perspective, using namespaces would only help, but there is the backward compatibility question as part of that once this code gets included in WP core.

justlevine:
> > Should we use a namespace WordPress/Abilities instead of the root WP_* ? Gutenberg doesnt, Performance does.
> > My gut says we _should_, because like GB our goal here is for experimentation/extenders and we don't want to have to worry about breaking back-compat when we merge into core.
> 
> Gutenberg doesn't use namespaces, so I'm a little bit confused. Did you mean we shouldn't?
> 

FML This is why I shouldn't do code review after midnight 🤦 I meant to ask if we want to _prefix or suffix_.

For example: [`WP_Duotone_Gutenberg`](https://github.com/WordPress/gutenberg/blob/d2fce222ebbbef8dbc56eee1badcfe4ae0df04b0/lib/class-wp-duotone-gutenberg.php#L40) becomes [`WP_Duotone`](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-duotone.php) in core.

Although I guess my subconscious mind was suggesting that namespaces could solve the same problem, but make the eventual migration even easier, e.g:

```diff
// If the final API doesnt change, then the user only needs to remove the `use` statement
- use SomeNamespace\wp_get_ability;
...
wp_get_ability( ...$args );

/// versus search-replace in the code logic
- some_prefix_get_ability( ...$args );
+ wp_get_ability( ...$args );
```

src/abilities-api.php
justlevine: Naming things is always the hardest part 😬

I don’t think GB (JS-first monorepo) or php-ai-client (standalone Composer lib) are quite the right references for a feature (or canonical) plugin.

FWIW, heres what other official WordPress/* projects are doing:

- [WordPress/TwoFactor](https://github.com/WordPress/two-factor) and Performance Plugins ([e.g.](https://github.com/WordPress/performance/tree/trunk/plugins/view-transitions) use `includes` for PHP and scatter JS (`assets`, root-level `css/js` folders, `src`)
- [Health Check](https://github.com/WordPress/health-check/tree/trunk/HealthCheck): PHP in `HealthCheck` namespace, build-step assets in `src`.
- [PCP](https://github.com/WordPress/plugin-check) uses `includes` for PHP and `assets` for JS/CSS`.
- [WordPress Importer](https://github.com/WordPress/wordpress-importer) uses `src` for PHP but it's legacy and has no js/css.

----
Semantically it seems the general pattern is
- `includes` (or `inc` but that might be more of an a8c thing, havn't seen it in `WordPress/*`) = core PHP (no build step)
- `src` are plugin assets that need a build step. OR sometimes they're co-located `assets`. (OR sometimes non-buildable PHP)
- `packages` reusable (and usually releasable) JS/TS libraries.
- havnt found any consistent use of `lib` in the org.

tl;dr I think `includes` or `inc` is best for clarity (unlike `src` we _know_ PHP is expected to be there). Then, if/when we need buildable assets for the admin we can see whether `assets` vs buildable `src` (or a ts `package`, but that seems unlikely in this repo) make sense without being boxed in by semantics.

gziolo: Interesting find is that there are best practices for WP plugins at https://developer.wordpress.org/plugins/plugin-basics/best-practices/#folder-structure. I see `includes` there. `src` for JavaScript code makes sense, too. The remaining question is where we locate the composer package and npm package that we will need here allowing early access to server-side registry, REST API controller, and the client-side package with the client-side registry that also integrates the server-side registry through REST API call.

gziolo: I implemented previously `wp-scripts plugin-zip` command and this is what I found there today:

https://github.com/WordPress/gutenberg/blob/fc937dccee20e7b0e6dfa671f6afac03f4bf6ef5/packages/scripts/scripts/plugin-zip.js#L32-L46

Let's go with `includes` to have an easy way to integrate that. For JS output, we can use either `build` or `public`.

gziolo:
> For example: [WP_Duotone_Gutenberg](https://github.com/WordPress/gutenberg/blob/d2fce222ebbbef8dbc56eee1badcfe4ae0df04b0/lib/class-wp-duotone-gutenberg.php#L40) becomes [WP_Duotone](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-duotone.php) in core.

This is used that way only when Gutenberg wants to override the preexisting class in WP Core. If the class or function doesn't exist then in Gutenberg the strategy is to polyfiil the functionality using check if such thing exists to prevent double declaration.

gziolo:
For the sake of illustration, I fixed all issues reported in https://github.com/WordPress/wordpress-develop/pull/9410 regarding PHP 7.2 and coding standards compatibility with [1975276](https://github.com/WordPress/abilities-api/pull/3/commits/1975276c582ca136185538ef91147cc4010a3dac). In particular, the changes around types might need to be reverted when we opt to go with PHP 7.4 as the minimum required version, but the rest might be fine to leave as is.

justlevine:
> > For example: [WP_Duotone_Gutenberg](https://github.com/WordPress/gutenberg/blob/d2fce222ebbbef8dbc56eee1badcfe4ae0df04b0/lib/class-wp-duotone-gutenberg.php#L40) becomes [WP_Duotone](https://github.com/WordPress/wordpress-develop/blob/trunk/src/wp-includes/class-wp-duotone.php) in core.
> 
> This is used that way only when Gutenberg wants to override the preexisting class in WP Core. If the class or function doesn't exist then in Gutenberg the strategy is to polyfiil the functionality using check if such thing exists to prevent double declaration.

I know better than to attempt to Gutenberg-splain to you 🙇 Thanks for the correction, I must be having some mandela effect moment about not needing to worry about my Fonts API or Interactivity API experiments clashing on merge until I renamed them 🤦

So with **absolutely 0 WordPress prior art** the question becomes more generically: "do we want to take some preventative measures (prefixing _or_ namespacing) to allow for immediate 3rd party adoption/use of the Abilities API without worrying about carrying tech-debt into core with us or breaking the early-ecosystem (both of which I think we can agree _do_ have prior art in WordPress 😅).

gziolo:
> do we want to take some preventative measures (prefixing or namespacing) to allow for immediate 3rd party adoption/use of the Abilities API without worrying about carrying tech-debt into core with us or breaking the early-ecosystem

We definitely should provide some recommendations on how to safely use the package in a way that will eventually get replaced by the version that ships in WP core. I'm not entirely sure what that will be, but that's part of the story we need to design. @jonathanbossenger plans to work on the documentation as outlined in #5, so we definitely need to coordinate with him.

gziolo:
Both #4 and #6 depend on this PR, so let's land it first. I tested everything against WordPress core CI, see https://github.com/WordPress/wordpress-develop/pull/9410. We can follow up with the tooling afterwards, while in parallel extend the logic with REST API controllers.

src/abilities-api.php
gziolo: I figured out it's best to tackle it in #4 after we land the initial code. This way it's going to be much easier to handle every code change while the CI is running.

PR #6, opened 2025-08-12
Add REST API controllers for Abilities API

### LINKED ISSUE #2: Implement REST API controller that exposes registered abilities
Depends on:
- #1 

See Feature API implementation as a reference:

https://github.com/Automattic/wp-feature-api/blob/trunk/includes/rest-api/class-wp-rest-feature-controller.php

emdashcodes:
Closes https://github.com/WordPress/abilities-api/issues/2.

This PR adds initial REST API endpoints to the Abilities API with support for both resource (GET) and tool (POST) type abilities.

## How

Implements the following REST API endpoints:

- `WP_REST_Abilities_List_Controller` - Handles listing abilities and retrieving individual abilities
  - `GET /wp-json/wp/v2/abilities` - List all registered abilities
  - `GET /wp-json/wp/v2/abilities/{id}` - Get a specific ability by ID
  
- `WP_REST_Abilities_Run_Controller` - Handles executing abilities
  - `GET/POST /wp-json/wp/v2/abilities/{id}/run` - Execute an ability (method depends on ability type)
    - Resources use `GET`
    - Tools use `POST`

## Testing Steps

1. Make sure all unit tests pass

If you enable the Ability API as a plugin:

1. Test listing abilities: `GET /wp/v2/abilities`
3. Test getting a specific ability: `GET /wp/v2/abilities/{id}`
4. Test executing a resource ability: `GET /wp/v2/abilities/{id}/run`
5. Test executing a tool ability: `POST /wp/v2/abilities/{id}/run`


abilities-api.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @package Abilities_API
+ */
+
+// Prevent direct access
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+// TODO: https://github.com/WordPress/abilities-api/pull/4/ also contains this
```
emdashcodes: I know this is already getting added in another PR, but I'm adding a temporary version to make testing easier.

gziolo:
I landed #3, which triggered GitHub automations, and accidentally closed this PR after the target branch got deleted 😅 

I'll try to fix it.

gziolo:
Reopened and changed the base branch, but it will require a rebase to remove my commits from the list of changes.

abilities-api.php
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @since 0.1.0
+ */
+function abilities_api_init() {
+	/**
+	 * Fires when the Abilities API is ready to register abilities.
+	 *
+	 * @since 0.1.0
+	 */
+	do_action( 'abilities_api_init' );
```
gziolo: Why was it needed here? It should be enough to run the ability registration inside a callback set for the `abilities_api_init` action.

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,292 @@
...
+		register_rest_route(
+			$this->namespace,
+			'/' . $this->rest_base . '/(?P<id>[a-zA-Z0-9\-\/]+)',
+			array(
+				'args' => array(
+					'id' => array(
+						'description' => __( 'Unique identifier for the ability.' ),
+						'type'        => 'string',
+						'pattern'     => '^[a-zA-Z0-9\-\/]+$',
+					),
```
gziolo: I went with the `name` property so we need to decide whether `id` or `name` should be used consistently. See
https://github.com/WordPress/abilities-api/blob/e7249f2b55bfbc5323cb93251f9b3b13370d1aa5/src/class-wp-ability.php#L23-L30

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,292 @@
...
+				),
+				array(
+					'methods'             => WP_REST_Server::READABLE,
+					'callback'            => array( $this, 'get_item' ),
+					'permission_callback' => array( $this, 'get_items_permissions_check' ),
```
gziolo: I see that `get_items_permissions_check` is used for both routes, so maybe renaming it to something more general like `get_permissions_check` would work better here.

abilities-api.php
gziolo: That's fine. It's helpful for the time being. It will also allow us to land this PR sooner if the other one is not finalized.

src/rest/class-wp-rest-abilities-init.php
```diff
@@ -0,0 +1,35 @@
...
+
+		$run_controller = new WP_REST_Abilities_Run_Controller();
+		$run_controller->register_routes();
+
+		$list_controller = new WP_REST_Abilities_List_Controller();
+		$list_controller->register_routes();
+	}
+}
+
+add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ), 120 );
```
gziolo: Some notes as I compare with other repos. I double-checked in Gutenberg, and all callbacks are registered there at the regular priority. In WordPress Core, all endpoints get required unconditionally in [one place](https://github.com/WordPress/wordpress-develop/blob/94cab031c85eaf86e5e6975421359b0be3b994b7/src/wp-settings.php#L289-L329). Then all routes get registered using [create_initial_rest_routes](https://github.com/WordPress/wordpress-develop/blob/94cab031c85eaf86e5e6975421359b0be3b994b7/src/wp-includes/rest-api.php#L265) method.

It gets executed at priority 99 [here](https://github.com/WordPress/wordpress-develop/blob/94cab031c85eaf86e5e6975421359b0be3b994b7/src/wp-includes/default-filters.php#L531), so I reason this ensures that plugins can use other registered routes.

@emdashcodes, can you confirm my reasoning? 

How would it be consumed inside the composer package? For WordPress core I anticipate we won't offer `WP_REST_Abilities_Init` class but inline the code in some of the places listed above in the correct order.

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,292 @@
...
+			'meta'          => $ability->get_meta(),
+		);
+
+		$context = ! empty( $request['context'] ) ? $request['context'] : 'view';
+		$data = $this->add_additional_fields_to_object( $data, $request );
+		$data = $this->filter_response_by_context( $data, $context );
+
+		$response = rest_ensure_response( $data );
+
+		$links = array(
```
gziolo: I see that in some places links are added when the following conditions are met:

```php
$fields = $this->get_fields_for_response( $request );
if ( rest_is_field_included( '_links', $fields ) || rest_is_field_included( '_embedded', $fields ) ) {
```

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,292 @@
...
+	 *
+	 * @return array Item schema data.
+	 */
+	public function get_item_schema() {
+		$schema = array(
+			'$schema'    => 'http://json-schema.org/draft-04/schema#',
+			'title'      => 'ability',
+			'type'       => 'object',
+			'properties' => array(
+				'id' => array(
```
gziolo: `id` (or `name`), `label` and `description` should be marked as `required`.

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,292 @@
...
+	/**
+	 * Retrieves all abilities.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function get_items( $request ) {
+		$abilities = wp_get_abilities();
```
gziolo, line 100: I see that it's a common practice in WP core to bail early for the `HEAD` method:

```php
if ( $request->is_method( 'HEAD' ) ) {
	// Return early as this handler doesn't add any response headers.
	return new WP_REST_Response( array() );
}
```

gziolo: I see that for example for comments, it stil would make sense to send the number of items and page with HEAD method:

```php
$response = $is_head_request ? new WP_REST_Response( array() ) : rest_ensure_response( $comments );
$response->header( 'X-WP-Total', $total_comments );
$response->header( 'X-WP-TotalPages', $max_pages );
```

[Reference](https://github.com/WordPress/wordpress-develop/blob/94cab031c85eaf86e5e6975421359b0be3b994b7/src/wp-includes/rest-api/endpoints/class-wp-rest-comments-controller.php#L316-L318).

That's something we can handle separately as it requires some more planning and it's definitely usage dedicated for performance optimizations.

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,292 @@
...
+
+		// Handle pagination.
+		$page     = $request['page'];
+		$per_page = $request['per_page'];
```
gziolo: Should there be some handling if these params are not included? I'm not entirely sure whether these default values from the schema are automatically propagated here.

tests/unit/REST/WPRESTAbilitiesListControllerTest.php
```diff
@@ -0,0 +1,397 @@
...
+	 */
+	public function test_get_items(): void {
+		$request  = new WP_REST_Request( 'GET', '/wp/v2/abilities' );
+		$response = $this->server->dispatch( $request );
+
+		$this->assertEquals( 200, $response->get_status() );
+
+		$data = $response->get_data();
+		$this->assertIsArray( $data );
+		$this->assertNotEmpty( $data );
```
gziolo, line 195: Would it be possible to assert if the item count is 50?

gziolo (commented):
I reviewed all the code and tests for the list controller and left my feedback inline. I still need to check the run controller. Everything is looking great, my feedback is mostly about minor things or I had some questions/notes where I wasn't entirely sure how it works. 

My only high-level, important feedback is around using a single form: `id` or `name` to uniquely identify the ability.

src/rest/class-wp-rest-abilities-run-controller.php
```diff
@@ -0,0 +1,311 @@
...
+			$input = $request->get_query_params();
+			unset( $input['_locale'], $input['context'], $input['page'], $input['per_page'] );
+		} else {
+			$input = $request->get_json_params() ?: array();
+		}
+
+		$input_validation = $this->validate_input( $ability, $input );
+		if ( is_wp_error( $input_validation ) ) {
+			return $input_validation;
+		}
```
gziolo, line 144: This part is not needed as it's covered inside `$ability->execute( $input )`.

src/rest/class-wp-rest-abilities-run-controller.php
```diff
@@ -0,0 +1,311 @@
...
+				'rest_ability_execution_failed',
+				__( 'Ability execution failed. Please check permissions and input parameters.' ),
+				array( 'status' => 500 )
+			);
+		}
+
+		$output_validation = $this->validate_output( $ability, $result );
+		if ( is_wp_error( $output_validation ) ) {
+			return $output_validation;
+		}
```
gziolo, line 167: This part is also not needed, as it is implemented inside `$ability->execute( $input )`.

src/rest/class-wp-rest-abilities-run-controller.php
gziolo: The only difference would be that validation on the ability returns `false` while logging the reason with `_doing_it_wrong`. We could consider returning the `WP_Error` instead.

src/rest/class-wp-rest-abilities-run-controller.php
gziolo: Similar to the input. The only difference would be that validation on the ability returns `false` while logging the reason with `_doing_it_wrong`. We could consider returning the `WP_Error` instead.

src/rest/class-wp-rest-abilities-run-controller.php
```diff
@@ -0,0 +1,311 @@
...
+			return new WP_Error(
+				'rest_ability_not_found',
+				__( 'Ability not found.' ),
+				array( 'status' => 404 )
+			);
+		}
+
+		if ( 'GET' === $request->get_method() ) {
+			$input = $request->get_query_params();
+			unset( $input['_locale'], $input['context'], $input['page'], $input['per_page'] );
```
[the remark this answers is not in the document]
gziolo: Would it be possible to send input parameters for the ability in a more reliable way? Here, if someone changes query params through hooks, they might interfere with input schema validation. What issues would pop if it's rewritten as:

```php
$input = $request->get_query_params()['input'];
```

src/rest/class-wp-rest-abilities-list-controller.php
emdashcodes: I updated this to use `name`.

src/rest/class-wp-rest-abilities-init.php
emdashcodes: @gziolo I've updated this to use default priority. 

Good point about the Init class potentially being inlined. I think for now having a separate class makes it easier to test as a plugin, but we can refactor when integrating into core.

> How would it be consumed inside the composer package? 

The Feature API does some version handling to only register the API once.. we might need to do something similar.

src/rest/class-wp-rest-abilities-list-controller.php
emdashcodes: I think this is a good idea. I added a TODO for it.

src/rest/class-wp-rest-abilities-list-controller.php
emdashcodes: They are propagated but it does seem like good defensive move to still include.

src/rest/class-wp-rest-abilities-run-controller.php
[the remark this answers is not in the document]
emdashcodes: Updated this 👍

src/rest/class-wp-rest-abilities-run-controller.php
emdashcodes: I added a TODO for now. The issue is that the false doesn't give us more detailed information about why it failed.

Like you noted,  if we update the main method to return WP_Error, we can use that instead.

I think that might make the most sense, and folks can catch it if needed.

src/rest/class-wp-rest-abilities-run-controller.php
emdashcodes: See above.

phpstan.neon.dist
```diff
@@ -29,3 +29,14 @@ parameters:
 			analyseAndScan:
 				- node_modules (?)
 
+		# Ignore specific errors
+		ignoreErrors:
+			# WP_REST_Request is not actually a generic class in WordPress core.
+			# PHPStan's WordPress stubs appear to define it as generic for type checking,
+			# but WordPress itself doesn't use generics. This seems to be an incompatibility
```
emdashcodes, line 36: @justlevine I'm not sure if there is a better way to handle this... but I was having trouble typing this in a way that worked. Any ideas?

emdashcodes:
I've addressed the static analysis and PHPCS issues from this branch. The remaining ones seem to be on `trunk`.

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,315 @@
...
+
+declare( strict_types = 1 );
+
+/**
+ * REST API: WP_REST_Abilities_List_Controller class
+ *
+ * @package WordPress
+ * @subpackage REST_API
+ * @since 0.1.0
+ */
```
justlevine: Looks like this got duped by the rebase. Only need one file header (_above_ the `declare()`).

emdashcodes: Thanks! Fixed

phpstan.neon.dist
justlevine: This issue is that WP_REST_Request expects the generic of args, e.g. `\WP_REST_Request<array:{name:string}>` for example on `WP_REST_Abilities_List_Controller::get_item()`.

If it's not important we can use `WP_REST_Request<array<string,mixed>>`, though it might also go away if we just use `{@inheritDoc}`, but im not by my desk rn to check.

(Either way probably should be {@inheritDoc} in most of those functions).

Might make sense to disable `SlevomatCodingStandard.TypeHints.*.MissingNativeTypeHint` on those classes though since the types are only hinted by the parent, not enforced.

src/rest/class-wp-rest-abilities-list-controller.php
```diff
@@ -0,0 +1,307 @@
...
+	/**
+	 * Retrieves a specific ability.
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param \WP_REST_Request $request Full details about the request.
+	 * @return \WP_REST_Response|\WP_Error Response object on success, or WP_Error object on failure.
+	 */
+	public function get_item( $request ) {
+		$ability = wp_get_ability( $request['name'] );
```
justlevine, line 150: Is there a reason not to use `WP_REST_Request->get_param()` (/ `->get_params()` elsewhere )?

We lose the `rest_request_parameter_order` filter by relying on `\ArrayAccess`

phpcs.xml.dist
```diff
@@ -193,6 +193,22 @@
 			<property name="enableStandaloneNullTrueFalseTypeHints" value="false" /><!-- Only available in PHP 8.2+ -->
 		</properties>
 	</rule>
+	
+	<!-- 
+		Exclude REST controllers from type hint requirements.
```
emdashcodes, line 198: I had some issues going around in circles with the type hinting not matching core and want the linter wants, so temporarily disabled until I can look at this more.

src/rest/class-wp-rest-abilities-list-controller.php
gziolo: Cool!

src/rest/class-wp-rest-abilities-list-controller.php
gziolo: Makes perfect sense.

src/rest/class-wp-rest-abilities-run-controller.php
gziolo: Good idea with TODO and follow up refactor 👍🏻

gziolo (approved):
Thank you for rebasing with `trunk` and addressing all feedback. I see the increased unit tests coverage added in 9b071cb9e5969c51d3e795c05c646424f3a27fce 👏🏻 

Let's land it in the current shape and address the code quality configuration challenges seperately. I also plan to open follow-up issues for TODO items included in the codebase.

src/rest/class-wp-rest-abilities-list-controller.php
gziolo: I see both forms used in REST API controllers included in WP core, sometimes even in the same handler 😅

phpstan.neon.dist
gziolo: I landed this PR without addressing it. Let's figure out the strategy for type hinting seperately. I didn't want this larger PR to get blocked on code quality concerns.

src/rest/class-wp-rest-abilities-run-controller.php
gziolo: Issue created:
- https://github.com/WordPress/abilities-api/issues/14

phpstan.neon.dist
gziolo: Tracked in https://github.com/WordPress/abilities-api/issues/13.

phpcs.xml.dist
gziolo: Tracked in https://github.com/WordPress/abilities-api/issues/13.

justlevine: Resolved in #15

src/rest/class-wp-rest-abilities-list-controller.php
gziolo: Now tracked in https://github.com/WordPress/abilities-api/issues/16.

src/rest/class-wp-rest-abilities-list-controller.php
justlevine: @gziolo I think that's a sign of tech debt, not an endorsement of pattern 😅

gziolo: We can update to use `get_param()` 👍🏻 

Will you take care of it?

phpstan.neon.dist
swissspidy: See https://github.com/swissspidy/media-experiments/blob/4aace63141fbd57cbece2d574129adfde3d6793d/inc/class-rest-attachments-controller.php for an example on how to properly do this in PHPStan. It's totally doable and there is definitely no need to ignore the errors.

justlevine: @swissspidy the question isnt whether it's doable, its whether it's justifiable when we need to strip out all the non-WPCS syntax in a few months... 

Though tbh I'm more jealous of the short array syntax than even the types 😅

swissspidy: It doesn't need to be stripped out though.

PR #8, opened 2025-08-13
“Fixes #5” so GitHub links the PR to the issue.

vinay070403:
Fixes #5

PR #10, opened 2025-08-13
Adding initial developer documentation

### LINKED ISSUE #5: Create developer documentation
Create documentation for developers on how to register and use Abilities.

See [Feature API docs](https://github.com/Automattic/wp-feature-api/tree/trunk/docs) for reference.

We can probably reuse what we need from he Feature API docs, but at a minimum we probably need:

1. [Intro](https://github.com/Automattic/wp-feature-api/blob/trunk/docs/1.intro.md)
2. [Getting started](https://github.com/Automattic/wp-feature-api/blob/trunk/docs/2.getting-started.md)
3. [Registering Abilities](https://github.com/Automattic/wp-feature-api/blob/trunk/docs/3.registering-features.md)
4. [Using Abilities](https://github.com/Automattic/wp-feature-api/blob/trunk/docs/4.using-features.md)
5. [REST API endpoints](https://github.com/Automattic/wp-feature-api/blob/trunk/docs/5.rest-api-endpoints.md)

The repo readme should also be updated to include a "quick start" example, as well as links to the rest of the docs.

jonathanbossenger:
Adds developer documentation for the Abilities API. This is still a work in progress.

Fixes #5


jonathanbossenger:
This documentation was AI-generated, based on the original Features API docs. 

My next step will be to review the generated docs manually and update them accordingly.

vinay070403:
Thanks for the update! I see that #8 is already in progress. I’ll close my PR to avoid duplication. Happy to help with other issues in the future!

gziolo:
I updated the branch with GitHub UI to the latest changes from `trunk` while testing the repo configuration. We are still working on tooling, so there is no need to worry about coding the failing checks with code analysis.

docs/1.intro.md
```diff
@@ -0,0 +1,66 @@
...
+
+The WordPress Abilities API provides a standardized way to register and discover distinct units of functionality within a WordPress site. These units, called "Abilities," represent specific actions or capabilities that components can perform, with clearly defined inputs, outputs, and permissions.
+
+It acts as a central registry, making it easier for different parts of WordPress, third-party plugins, themes, and external systems (like AI agents) to understand and interact with the capabilities available on a specific site.
+
+**Core Concepts**
+
+- **Ability:** A distinct piece of functionality with a unique ID following the `namespace/ability-name` pattern. Each ability has a human-readable name and description, input/output definitions (using JSON Schema), optional permissions, and an associated callback function for execution.
+- **Registry:** A central, singleton object (`WP_Abilities_Registry`) that holds all registered abilities. It provides methods for registering, unregistering, finding, and querying abilities.
+- **Callback:** The PHP function or method executed when an ability is called via `wp_execute_ability()`.
```
jonathanbossenger: **Callback:** The PHP function or method executed when an ability is called via `WP_Ability::do_execute`.

docs/1.intro.md
```diff
@@ -0,0 +1,66 @@
...
+
+**What is the Abilities API?**
+
+The WordPress Abilities API provides a standardized way to register and discover distinct units of functionality within a WordPress site. These units, called "Abilities," represent specific actions or capabilities that components can perform, with clearly defined inputs, outputs, and permissions.
+
+It acts as a central registry, making it easier for different parts of WordPress, third-party plugins, themes, and external systems (like AI agents) to understand and interact with the capabilities available on a specific site.
+
+**Core Concepts**
+
+- **Ability:** A distinct piece of functionality with a unique ID following the `namespace/ability-name` pattern. Each ability has a human-readable name and description, input/output definitions (using JSON Schema), optional permissions, and an associated callback function for execution.
```
jonathanbossenger: Each registered Ability is an instance of the `WP_Ability` class.

jonathanbossenger: unique name

docs/1.intro.md
```diff
@@ -0,0 +1,66 @@
...
+- **Registry:** A central, singleton object (`WP_Abilities_Registry`) that holds all registered abilities. It provides methods for registering, unregistering, finding, and querying abilities.
+- **Callback:** The PHP function or method executed when an ability is called via `wp_execute_ability()`.
+- **Schema:** JSON Schema definitions for an ability's expected input (`input_schema`) and its returned output (`output_schema`). This allows for validation and helps agents understand how to use the ability.
+- **Permission Callback:** An optional function that determines if the current user can execute a specific ability. Falls back to input validation if not provided.
+- **Namespace:** The first part of an ability ID (before the slash), typically matching the plugin or component name that registers the ability.
```
jonathanbossenger: The first part of an ability name

docs/1.intro.md
```diff
@@ -0,0 +1,66 @@
...
+- **AI Integration:** Allow AI agents to discover and interact with site capabilities
+- **Plugin Interoperability:** Enable plugins to discover and use each other's functionality
+- **Automation Tools:** Provide programmatic access to site features
+- **API Documentation:** Self-documenting capabilities with schema validation
+- **Developer Tools:** Standardized way to expose plugin functionality
+
+**Registration Example**
+
+```php
+add_action( 'abilities_api_init', function() {
```
jonathanbossenger: @gziolo, do we have a specific requirement for anonymous callback functios over named functions in documentation examples? 

I know that using named functions allows developers to unhook callbacks, but I don't think this is a problem for registering Abilities.

jonathanbossenger: Follow-up question: Should we standardize example code using traditional syntax vs more modern PHP syntax, example using the `array` keyword vs the short array notation `[]`?

docs/2.getting-started.md
```diff
@@ -0,0 +1,56 @@
+# 2. Getting Started
+
+## Installation Options
+
+(To be determined)
```
jonathanbossenger: Update this to include current installation instructions as a plugin, but prepare to add composer instructions when the composer package is available.

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,272 @@
+### 3. Registering Abilities (`wp_register_ability`)
+
+The primary way to add functionality to the Abilities API is by using the `wp_register_ability()` function, typically hooked into the `abilities_api_init` action.
+
+**Function Signature**
+
+```php
+wp_register_ability( string $id, array $args ): bool
```
jonathanbossenger: Update instances of $id with $name

jonathanbossenger: wp_register_ability returns an instance of registered ability on success, null on failure.

docs/4.using-abilities.md
```diff
@@ -0,0 +1,297 @@
...
+    echo 'Ability Name: ' . esc_html( $ability->get_name() ) . "\n";
+    echo 'Label: ' . esc_html( $ability->get_label() ) . "\n";
+    echo 'Description: ' . esc_html( $ability->get_description() ) . "\n";
+    echo "---\n";
+}
+```
+
+**Executing an Ability (`$ability->execute()`)**
+
+Once you have a `WP_Ability` object (usually from `wp_get_ability`), you execute it using the `execute()` method.
```
jonathanbossenger, line 61: do_execute() method

docs/4.using-abilities.md
```diff
@@ -0,0 +1,297 @@
...
+$ability = wp_get_ability( 'my-plugin/some-ability' );
+
+if ( ! $ability ) {
+    // Ability not registered
+    echo 'Ability not found';
+    return;
+}
+
+try {
+    $result = $ability->execute( $input );
```
jonathanbossenger: do_execute()

docs/4.using-abilities.md
```diff
@@ -0,0 +1,297 @@
...
+    // Success - use the result
+    // Process $result based on the ability's output schema
+    
+} catch ( Exception $e ) {
+    // Handle exceptions thrown by the execute callback
+    echo 'Exception during execution: ' . $e->getMessage();
+}
+```
+
+**Practical Usage Examples**
```
jonathanbossenger: This entire section is probably unnessecary

jonathanbossenger (commented):
Initial manual review complete. I just want to test all the code examples.

docs/4.using-abilities.md
gziolo: `do_execute` is a protected method, so I would keep `execute` everywhere in the docs.

docs/1.intro.md
gziolo: From https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#closures-anonymous-functions 

> Closures should not be passed as filter or action callbacks, as removing these via remove_action() / remove_filter() is complex (at this time) (see [#46635](https://core.trac.wordpress.org/ticket/46635) for a proposal to address this).

However, this applies mainly to WordPress core, as this gives the control to the extenders to modify the behavior with ease. In the case of abilities, you still can unregistered them by name, so that might be less of an issue for plugins. 

> Follow-up question: Should we standardize example code using traditional syntax vs more modern PHP syntax, example using the array keyword vs the short array notation []?

I'm not entirely sure what the best practices are in the dev notes and on the Developers blogs, but I would follow with that. I don't have strong preferences.

justlevine: My 2C

> do we have a specific requirement for anonymous callback functions over named functions in documentation examples?

No requirement to be anonymous it's just an action as far as the code cares. Users can deregister abilities without unhooking their action using their `$name)`, so while general hook hygiene is nice it's nonessential. (although to be pedantic, if we're keeping it anonymous then at least make it `static function()`.)

> Should we standardize example code using traditional syntax vs more modern PHP syntax, example using the array keyword vs the short array notation []?

Personally think any usage examples (i.e. not core code itself) should use modern PHP syntax and encourage best practices, but `plugin-check` uses full WPCS i.e. `array()` ([e.g.](https://github.com/WordPress/plugin-check/blob/trunk/docs/creating-a-static-check.md)) if we're looking at prior art and Chesterton Fences

docs/3.registering-abilities.md
```diff
@@ -5,12 +5,12 @@ The primary way to add functionality to the Abilities API is by using the `wp_re
...
 
 ```php
-wp_register_ability( string $id, array $args ): bool
+wp_register_ability( string $id, array $args ): ?\WP_Ability
 ```
 
 -   `$id` (`string`): A unique identifier for the ability.
 -   `$args` (`array`): An array of arguments defining the ability configuration.
--   **Return:** (`bool`) `true` if the ability was successfully registered, `false` on failure (e.g., invalid arguments, duplicate ID).
+-   **Return:** (`?\WP_Ability`) An instance of the registered ability if the it was successfully registered, `null` on failure (e.g., invalid arguments, duplicate ID).
```
gziolo, line 13: This might change after we land:
- https://github.com/WordPress/abilities-api/pull/23

gziolo: Ah wait, it's a different use case. I will walk myself out 😄

docs/1.intro.md
jonathanbossenger: Thanks, I'll formalize to stick to WPCS for now then, on both counts.

docs/4.using-abilities.md
jonathanbossenger: :facepalm: my apologies, I missed the public method.

jonathanbossenger:
Updated all the factual errors. Last step is to test and verify the code examples, but I'm marking this as ready to review so long.

docs/1.intro.md
```diff
@@ -0,0 +1,67 @@
+# 1. Introduction & Overview
+
+**What is the Abilities API?**
+
+The WordPress Abilities API provides a standardized way to register and discover distinct units of functionality within a WordPress site. These units, called "Abilities," represent specific actions or capabilities that components can perform, with clearly defined inputs, outputs, and permissions.
```
gziolo: Typo: 
```suggestion
The WordPress Abilities API provides a standardized way to register and discover distinct units of functionality within a WordPress site. These units, called "Abilities", represent specific actions or capabilities that components can perform, with clearly defined inputs, outputs, and permissions.
```

docs/1.intro.md
```diff
@@ -0,0 +1,67 @@
...
+
+It acts as a central registry, making it easier for different parts of WordPress, third-party plugins, themes, and external systems (like AI agents) to understand and interact with the capabilities available on a specific site.
+
+**Core Concepts**
+
+- **Ability:** A distinct piece of functionality with a unique name following the `namespace/ability-name` pattern. Each ability has a human-readable name and description, input/output definitions (using JSON Schema), optional permissions, and an associated callback function for execution. Each registered Ability is an instance of the `WP_Ability` class.
+- **Registry:** A central, singleton object (`WP_Abilities_Registry`) that holds all registered abilities. It provides methods for registering, unregistering, finding, and querying abilities.
+- **Callback:** he PHP function or method executed when an ability is called via `WP_Ability::execute()`.
```
gziolo: Typo:
```suggestion
- **Callback:** the PHP function or method executed when an ability is called via `WP_Ability::execute()`.
```

docs/1.intro.md
```diff
@@ -0,0 +1,67 @@
...
+- **Schema:** JSON Schema definitions for an ability's expected input (`input_schema`) and its returned output (`output_schema`). This allows for validation and helps agents understand how to use the ability.
+- **Permission Callback:** An optional function that determines if the current user can execute a specific ability. Falls back to input validation if not provided.
```
gziolo: I wouldn't say it falls back, but it's only the phrasing. In my mind, input needs to be valid only because you can't execute the ability if it has invalid input. Since the input is also passed to the permission check so we need to set the same level of scrutiny.

docs/1.intro.md
```diff
@@ -0,0 +1,67 @@
...
+- **Extensibility:** Simple registration pattern allows any plugin or theme to expose their capabilities.
+- **AI-Friendly:** Machine-readable format enables intelligent automation and AI agent interactions.
+
+**Use Cases**
+
+- **AI Integration:** Allow AI agents to discover and interact with site capabilities
+- **Plugin Interoperability:** Enable plugins to discover and use each other's functionality
+- **Automation Tools:** Provide programmatic access to site features
+- **API Documentation:** Self-documenting capabilities with schema validation
+- **Developer Tools:** Standardized way to expose plugin functionality
```
gziolo: Nit consistent styling:

```suggestion
- **AI Integration:** Allow AI agents to discover and interact with site capabilities.
- **Plugin Interoperability:** Enable plugins to discover and use each other's functionality.
- **Automation Tools:** Provide programmatic access to site features.
- **API Documentation:** Self-documenting capabilities with schema validation.
- **Developer Tools:** Standardized way to expose plugin functionality.
```

docs/2.getting-started.md
```diff
@@ -0,0 +1,61 @@
+# 2. Getting Started
+
+## Installation Options
+
+Currently, the Abilities API is available as a WordPress plugin. You can install it by cloning the repository into your `wp-content/plugins` directory.
```
gziolo, line 5: It would be nice to clarify the installation options. I asked about it in https://github.com/WordPress/abilities-api/pull/24#issuecomment-3199427983. For now, it's valid, but I'm not entirely sure what the expected end goal is.

docs/1.intro.md
```diff
@@ -0,0 +1,67 @@
...
+
+**Registration Example**
+
+```php
+add_action( 'abilities_api_init', `my_plugin_register_ability`);
+function my_plugin_register_ability(){
+    wp_register_ability( 'my-plugin/create-post', array(
+        'label' => 'Create Post',
+        'description' => 'Creates a new blog post with the provided content',
```
gziolo: This could use i18n helpers, as in another example.

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,276 @@
...
+
+The `$args` array accepts the following keys:
+
+-   `label` (`string`, **Required**): A human-readable name for the ability. Used for display purposes. Should be translatable.
+-   `description` (`string`, **Required**): A detailed description of what the ability does, its purpose, and its parameters or return values. This is crucial for AI agents to understand how and when to use the ability. Should be translatable.
+-   `input_schema` (`array`, **Required**): A JSON Schema definition describing the expected input parameters for the ability's execute callback. Used for validation and documentation.
+-   `output_schema` (`array`, **Required**): A JSON Schema definition describing the expected format of the data returned by the ability. Used for validation and documentation.
+-   `execute_callback` (`callable`, **Required**): The PHP function or method to execute when this ability is called.
+    -   The callback receives one argument: an associative array of validated input parameters.
+    -   The callback should return the result of the ability's operation or throw an exception on failure.
```
gziolo: It should return `WP_Error` rather than throwing an exception on failure.

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,276 @@
...
+-   `permission_callback` (`callable`|`null`, **Optional**): A callback function to check if the current user has permission to execute this ability.
+    -   Should return `true` if the user has permission, `false` or a `WP_Error` object otherwise.
```
gziolo: We have just changed that in fb953007e265cfa0e2c23bad8f9857a3b6e51682 to return `true` or `WP_Error`.

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,276 @@
...
+    -   If not provided, the ability will only validate input parameters before execution.
+    -   Use `'__return_true'` for publicly accessible abilities.
```
gziolo: It technically isn't needed as it defaults to `true`.

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,276 @@
...
+-   `meta` (`array`, **Optional**): An associative array for storing arbitrary additional metadata about the ability.
+
+**Ability ID Convention**
+
+The `$id` parameter must follow the pattern `namespace/ability-name`:
+-   **Format:** Must contain only lowercase alphanumeric characters (`a-z`, `0-9`), hyphens (`-`), and one forward slash (`/`) for namespacing.
+-   **Convention:** Use your plugin slug as the namespace, like `my-plugin/ability-name` or `woocommerce/get-product`.
+-   **Examples:** `my-plugin/update-settings`, `contact-form/send-message`, `analytics/track-event`
```
gziolo: Nit, since there are examples listed, maybe this would work better:
```suggestion
-   **Convention:** Use your plugin slug as the namespace, like `my-plugin/ability-name`.
-   **Examples:** `my-plugin/update-settings`, `woocommerce/get-product`, `contact-form/send-message`, `analytics/track-event`
```

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,276 @@
...
+        ),
+        'execute_callback' => function( $input ) {
+            $sent = wp_mail(
+                $input['to'],
+                $input['subject'],
+                $input['message']
+            );
+            
+            if ( ! $sent ) {
+                throw new Exception( 'Failed to send email' );
```
gziolo: Taking into account the current REST API implementation, it would be better to return `WP_Error` so it can be automatically handled. See:

https://github.com/WordPress/abilities-api/blob/fb953007e265cfa0e2c23bad8f9857a3b6e51682/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php#L146-L150

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,276 @@
...
+
+**Usage Examples**
+
+```php
+// Get a registered ability
+$ability = wp_get_ability( 'my-plugin/get-site-info' );
+
+if ( $ability ) {
+    // Execute the ability
+    try {
```
gziolo: Assuming we use `WP_Error` everywhere, this example would have to get adjusted.

docs/3.registering-abilities.md
```diff
@@ -0,0 +1,276 @@
...
+            return array( 'sent' => true );
+        },
+        'permission_callback' => function() {
+            return current_user_can( 'publish_posts' );
+        }
+    ));
+}
+```
+
+**Usage Examples**
```
gziolo: By the way, the next document is about usage, so probably this section should get moved there.

docs/3.registering-abilities.md
gziolo: Well, we can discuss enhancement seperately from this effort. The point is that it will fail with a generic 500 HTTP status code when executing the ability through the REST API endpoint.

docs/1.intro.md
jonathanbossenger: I wasn't sure about this one myself. Maybe it's just better to leave that line out entirely?

docs/2.getting-started.md
jonathanbossenger: Agreed. I've made a note to keep an eye on this and subscribed to #24. Once a call has been made I'll submit an update to the docs.

docs/3.registering-abilities.md
jonathanbossenger: I'll update to "return an instance of WP_Error on failure" for now, we can update if things change.

jonathanbossenger:
- Updated PR based on review feedback (props @gziolo)
- Updated Using Abilities examples to rely on WP_Error object return on failure

gziolo:
Would it make sense to land the initial version and tackle the subsequent iterations in follow-up targeted PRs? I approved PR, so with all checks passing, you should be able to merge at any point.

gziolo (approved):

jonathanbossenger:
> Would it make sense to land the initial version and tackle the subsequent iterations in follow-up targeted PRs? I approved PR, so with all checks passing, you should be able to merge at any point.

Yup, that was my plan as well. Going to merge here in a bit.

docs/1.intro.md
gziolo: Works for me 👍🏻

PR #4, opened 2025-08-08
chore: scaffold plugin entrypoint and tooling

justlevine:
## What

This PR scaffolds the plugins tooling and development/test environment. More specifically:

- `phpcs` with VIPCS + Slevomat + Plugin Check rulesets
- `phpstan` level 8
- `phpunit` running on `wp-env` with codecoverage.
- `prettier` running via `@wordpress/scripts`

CI/CD has been set up:
- `test.yml` - for PR linting/testing. Includes annotations + artifacts for human + 🤖 review
- `copilot-setup-steps.yml` for agentic developer environment.

Additionally, the root-level `./abilities-api.php` was created with the requisite plugin headers for activation, as well as some other standard config files.

> [!NOTE]
> This PR is based on #3 solely so there is code to test this PR against. However, we can easily rebase and apply this first if we want to get our tooling in place.
> As a result there are some incongruencies with decision made in #3. Those differences are not important **unless explicitly mentioned later**

### Minimum PHP v7.4
Regardless of [when](https://core.trac.wordpress.org/ticket/62622) it comes to WordPress. PHP 7.4 gives us access to better and easier tooling, and [a prereq for most other things we wanna do](https://github.com/WordPress/php-ai-client/issues/3). Considering we're going the canonical route anyway until merge, it seems pointless to handcuff us at this point.

> [!NOTE]
> For illustration, PHPCS and PHPStan are ~currently set to use PHP 7.2 as a minimum~, just to highlight what we're already violating by writing code like we're not in the aughts.
> Edit 13 Aug: Bumped to 7.4

### Strict PHPCS / PHPStan Rulesets

Crossed sourced from existing `WordPress/*` feature/canonical plugins, with the goal to be:

- **Strict**: Enables both humans and 🤖 to write quality code w.o. slowing down a senior developer on the basic/unimportant.
- **Autofixable**: Non WPCS-required and other opinionated checks should save time and cognitive load.
- **Enforcable**: Runs on CI/CD with annotations for users and 🤖, so issues arent missed. As committers we can always decide to ignore a smell or instance thereof.

That said, beyond the strictness and auto-enforcement, I don't feel too strongly about the formatting smells. Figure it's easier to remove/disable a rule when we see its effects on real-world code.

### V0.0.1

[...]

PR #11, opened 2025-08-13
chore: php remediations

justlevine:
## What

This is a first pass of PHPStan/PHPCS remediations. Most were auto-fixed, a couple (like moving the file headers above the `declare_strict()` ), 

Some tweaks to the `phpcs.xml.dist` as well, due to some accidental conflicts when I was deduping rulesets.

Follow up to #4 
---

Remaining smells are:

-  `WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedFunctionFound` - do we want a namespace?
- `WordPress.WP.I18n.MissingArgDomain` - do we want this translatable before merged into core?
- Missing array shapes - I just didnt have time right now to review the APIs and add them it but this is easy to do.

And then some things that need actual remediation (using `WP_Ability` as a string) or a decision about if we want to suppress/fix/turn off the rule.

## Why

1. In its current shape, it incrementally remediates making the "noise smaller" (and the merge conflicts fewer)
2. Lets us visually deliberate about rules we may or may not care about.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -58,7 +58,10 @@
 		<!-- Deprecated: @todo remove in PHPCS 4.0 -->
 		<exclude name="WordPressVIPMinimum.JS" />
 	</rule>
-	<rule ref="WordPress-Extra" />
+	<rule ref="WordPress-Extra">
+		<!-- Needed to typehint, see: https://github.com/WordPress/WordPress-Coding-Standards/issues/403 -->
+		<exclude name="Generic.Commenting.DocComment.MissingShort" />
+	</rule>
 
 	<!--
 		Load PHPCompatibility & PHPCompatibilityWP
@@ -70,7 +73,8 @@
 		Checks from PCP
 		@see https://github.com/wordpress/plugin-check
 	-->
-	<rule ref="./vendor/wpackagist-plugin/plugin-check/vendor/plugin-check/phpcs-sniffs/PluginCheck/ruleset.xml" />
+	<rule
+		ref="./vendor/wpackagist-plugin/plugin-check/vendor/plugin-check/phpcs-sniffs/PluginCheck/ruleset.xml" />
 	<rule ref="./vendor/wpackagist-plugin/plugin-check/phpcs-rulesets/plugin-check.ruleset.xml" />
 
 	<!-- Check for superfluous whitespace - These get suppressed in VIP-GO.-->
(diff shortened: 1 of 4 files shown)


phpcs.xml.dist
```diff
@@ -58,7 +58,10 @@
 		<!-- Deprecated: @todo remove in PHPCS 4.0 -->
 		<exclude name="WordPressVIPMinimum.JS" />
 	</rule>
-	<rule ref="WordPress-Extra" />
+	<rule ref="WordPress-Extra">
+		<!-- Needed to typehint, see: https://github.com/WordPress/WordPress-Coding-Standards/issues/403 -->
+		<exclude name="Generic.Commenting.DocComment.MissingShort" />
+	</rule>
```
justlevine: To test without the needs-decision noise, add



```suggestion
	<rule ref="WordPress-Extra">
		<exclude name="WordPress.WP.I18n.MissingArgDomain" />
		<exclude name="WordPress.NamingConventions.PrefixAllGlobals" />
		<!-- Needed to typehint, see: https://github.com/WordPress/WordPress-Coding-Standards/issues/403 -->
		<exclude name="Generic.Commenting.DocComment.MissingShort" />
	</rule>
```

And then you'll just have the array-shape docblock notices left to review

gziolo: I'm tempted to make these changes initially.

src/abilities-api.php
```diff
@@ -35,7 +36,7 @@ function wp_register_ability( $name, array $properties = array() ): ?WP_Ability
 				/* translators: 1: abilities_api_init, 2: string value of the ability name. */
 				esc_html__( 'Abilities must be registered on the %1$s action. The ability %2$s was not registered.' ),
 				'<code>abilities_api_init</code>',
-				'<code>' . esc_attr( $name ) . '</code>'
+				'<code>' . esc_html( $name ) . '</code>'
```
gziolo:
```suggestion
				'<code>' . esc_html( $name instanceof WP_Ability ? $name->get_name() : $name ) . '</code>'
```

gziolo (approved):
That's a nice improvement. We can address the rest of reported issues in follow-up PRs. I like how it helps to catch some small things in the code 💯

gziolo:
> And then some things that need actual remediation (using WP_Ability as a string) or a decision about if we want to suppress/fix/turn off the rule.

This was an actual error that I fixed in [f762155](https://github.com/WordPress/abilities-api/pull/11/commits/f7621556868fdf9152036c7748d27f868870e59a).

Issue #13, opened 2025-08-14
Scaffold repository tooling


gziolo:
Initial setup for the repository was added with:
- https://github.com/WordPress/abilities-api/pull/4

Let’s coordinate remaining tasks in this issue:
- [x] Prepare the composer package for publication and consumption by WordPress plugins.
- [x] Decide whether to use namespaces in the plugin for Abilities API, replicate that in WP Core and how to use it in plugins with future compatibility in mind.
- [x] Decide on the minimum required PHP version (7.4 vs 7.2).
- [x] Fix remaining coding standards issues reported by CI (https://github.com/WordPress/abilities-api/pull/4#issuecomment-3188060840).
- [x] Ensure CI checks run when landing PRs in `trunk` (https://github.com/WordPress/abilities-api/pull/4#issuecomment-3188104309).
- [x] Audit `ignoreErrors` added to PHPStan (https://github.com/WordPress/abilities-api/pull/6#discussion_r2274466039).
- [x] Audit excludes added for REST controllers in PHPCS (https://github.com/WordPress/abilities-api/pull/6#discussion_r2274631674).
- [x] Upload the code coverage report for base (example message https://github.com/WordPress/abilities-api/pull/6#issuecomment-3184260517).
- [x] Ensure that CI jobs fail when unit tests fail (see example https://github.com/WordPress/abilities-api/pull/19#issuecomment-3195288405).

justlevine:
> * Fix remaining coding standards issues reported by CI (https://github.com/WordPress/abilities-api/pull/4#issuecomment-3188060840).
> * Audit excludes added for REST controllers in PHPCS ([Add REST API controllers for Abilities API #6 (comment)](https://github.com/WordPress/abilities-api/pull/6#discussion_r2274631674)).

Advocating removing them entirely:  https://github.com/WordPress/abilities-api/pull/15/files has PHPCS passing with intent.

justlevine:
> * Ensure CI checks run when landing PRs in `trunk` ([chore: scaffold plugin entrypoint and tooling #4 (comment)](https://github.com/WordPress/abilities-api/pull/4#issuecomment-3188104309)).

https://github.com/WordPress/abilities-api/pull/17 should fix this

justlevine:
Going a bit out of technical order for logical cohesion:

> * Audit ignoreErrors added to PHPStan

I think the broader question here is do we want to take advantage or turn off `missingType.generics` - not specific to REST.

[...]

PR #15, opened 2025-08-14
tools(phpcs): remove conflicting TypeHint rules from PHPCS ruleset.

justlevine:
## What

This PR removes unnecessary and conflicting PHPCS rules regarding Type Hints and signatures.

## Why

Part of https://github.com/WordPress/abilities-api/issues/13

I was over optimistic at seeing them used in [`WordPress/performance`](https://github.com/WordPress/performance/blob/123450cb2b56e5e1f8ec0accc14de9cdf1f38183/tools/phpcs/phpcs.ruleset.xml#L94-L123) , but as it's getting in the way there's no need to include them.

What's important can be caught in code review or in follow up PRs before merge.

### DIFF
--- a/phpcs.xml.dist
+++ b/phpcs.xml.dist
@@ -59,6 +59,7 @@
 		<exclude name="WordPressVIPMinimum.JS" />
 	</rule>
 	<rule ref="WordPress-Extra">
+		<!-- @todo: Determine a strategy on namespacing/merging into core. -->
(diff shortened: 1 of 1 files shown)


gziolo (approved):
Makes sense to start with less strict rules.

gziolo:
All checks are required now. I enabled auto-merge 🎉

justlevine:
>  I enabled auto-merge 🎉 

Love that we can but ps this makes me want to prefix/namespace even more 😁

PR #19, opened 2025-08-16
dev: use `WP_Abilities_Registry::$instance` instead of adding a new `global`

justlevine:
## What

This PR changes `WP_Abilities_Registry` to store it's instance on `::$instance`, instead of on a new `global $wp_abilities` variable.

(Note: I'm just code-reviewing and leaving feedback. If you think this warrants a wider discussion I can open a matching issue, otherwise I think a discussion here following by [accept | reject] is enough)

## Why

Global variables are hard-to-type-and-test antipatterns. Just because legacy WP code used them before there were better patterns doesn't mean we need to introduce new ones.  

Using a private instance gives core contributors _and_ end-devs more insight into usage, and as developers working on the Abilities API, we get a much smaller surface for breaking changes to worry about.

### DIFF
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -19,6 +19,14 @@ declare( strict_types = 1 );
  */
 final class WP_Abilities_Registry {
 	/**
+	 * The singleton instance of the registry.
+	 *
+	 * @since 0.1.0
+	 * @var ?self
+	 */
+	private static ?self $instance = null;
+
+	/**
 	 * Holds the registered abilities.
 	 *
 	 * @since 0.1.0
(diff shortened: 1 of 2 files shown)


gziolo:
One of the unit tests is failing with these changes:

Screenshot 2025-08-18 at 08 22 43

Unfortuntalty, the existing CI check doesn't propagate test failures and all jobs still pass.

gziolo (changes requested):
I don't have a strong preference here. WordPress uses globals for many fundamental singletones like `$post`, `$wp_scripts`, `$wp_rest_server`. However, we need to make it fully testable and be able to reset the instance between test runs.

justlevine:
> Unfortuntalty, the existing CI check doesn't propagate test failures and all jobs still pass.

Also unfortunately, our PHPUnit CI is passing when the tests itself fail - I'll take a look at that too! 😅

gziolo (approved):
Using reflection makes the test pass 👍

PR #18, opened 2025-08-14
chore: align reqs with WP6.9 merge target

justlevine:
## What

This PR

- Drops the PHPCS/PHPStan min PHPVersion check to 7.2
- Bumps the min WordPress check to 6.8
- Removes the use of text-domains
- Updates the allowed prefixes
- Aligns the initial version number to 0.1.0

in accordance with a planned WordPress 6.9 merge.

## Why

We still benefit from the tooling by leaving the composer requirement at 7.4, but keep the codebase mergeable for 6.9 by linting against PHP 7.2.

Part of #13

PR #20, opened 2025-08-16
code quality: add PHPStan array shape for ability `$properties`

justlevine:
## What

This PR adds array shapes for the `$properties` array used by 

- `wp_register_ability()`
- `WP_Abilities_Registry::register()`
- `WP_Ability::__construct()`

## Why

To ensure both we and consumers are using the correct array shapes.

## Additional Notes

- Generics are _not_ used e.g. to hint the schema/meta shapes. Until core supports generics, that extra level of DX would fall to a 3rd party library like `php-stubs/wordpress-stubs`.
- As a result of the array shapes we could in theory make the param less verbose, but human contribs might not be familiar with the syntax, so I left it out.

### DIFF
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -27,6 +27,17 @@ declare( strict_types = 1 );
  *                                        include `label`, `description`, `input_schema`, `output_schema`,
  *                                        `execute_callback`, `permission_callback`, and `meta`.
  * @return ?\WP_Ability An instance of registered ability on success, null on failure.
+ *
+ * @phpstan-param array{
+ *   label?: string,
+ *   description?: string,
+ *   input_schema?: array<string,mixed>,
+ *   output_schema?: array<string,mixed>,
+ *   execute_callback?: callable( array<string,mixed> $input): (mixed|\WP_Error),
+ *   permission_callback?: callable( ?array<string,mixed> $input ): bool,
+ *   meta?: array<string,mixed>,
+ *   ...<string, mixed>
+ * } $properties
  */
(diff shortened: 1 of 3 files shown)


justlevine:
I don't have perms to rerun the failing CI, but since it's for WP6.7 which we're removing in #18 🤷

includes/abilities-api.php
```diff
@@ -27,6 +27,17 @@
...
+ *
+ * @phpstan-param array{
+ *   label?: string,
+ *   description?: string,
+ *   input_schema?: array<string,mixed>,
+ *   output_schema?: array<string,mixed>,
+ *   execute_callback?: callable( array<string,mixed> $input): (mixed|\WP_Error),
+ *   permission_callback?: callable( ?array<string,mixed> $input ): bool,
+ *   meta?: array<string,mixed>,
+ *   ...<string, mixed>
```
gziolo, line 39: Everything else gets ignored, do we need to list it?

gziolo (approved):
That's very useful addition 👍🏻

includes/abilities-api.php
justlevine: That's exaclty why it's here -  this is the long-overdue syntax for partially sealed arrays.

Otherwise if users don't clean up their arrays (or more "justifiable" real world scenario: if extensions hook in their own custom args) they'll get errors about unsupported array values, when it doesn't matter because we only keep the ones we need.

gziolo: I don’t think it adds any value at the moment because it’s impossible to successfully pass more settings. It’ll rather change in the future when we add WP hooks.

PR #22, opened 2025-08-16
dev: use `$request->get_params()` in REST controllers

justlevine:
## What

This PR updates our WP_REST_* controller classes to use `$request->get_params()` or `$request->get_param( $param_name )`, instead of using `ArrayAccess` to get the $request parameters directly.

## Why

This opts the params into the full `WP_Rest*` lifecycle, so [the `rest_request_parameter_order` filter](https://github.com/WordPress/wordpress-develop/blob/6.8.2/src/wp-includes/rest-api/class-wp-rest-request.php#L397C26-L397C54) etc are applied consistently.


gziolo (approved):
Nice, thank you for a follow up 👍🏻

PR #23, opened 2025-08-16
dev!: return `WP_Error` from `WP_Ability` methods

### LINKED ISSUE #14: Extend error handling for ability schema validation
Follow-up for:
- https://github.com/WordPress/abilities-api/pull/6

REST API needs detailed error messages with HTTP status codes. While `WP_Ability::execute()` validates internally, it only returns `false` and logs with `_doing_it_wrong()`, which doesn't provide capturable error messages. Related code that handles it today:

https://github.com/WordPress/abilities-api/blob/trunk/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php#L137L144

https://github.com/WordPress/abilities-api/blob/dd94fce1a6ba8930d39f2011fa1e6147f9759ad8/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php#L164L169

Consider updating `WP_Ability` to return `WP_Error` for better error handling which would then simplify and unify the implementation in REST API controller.

justlevine:
## What

This PR updates various `WP_Ability` methods to return an explicit `WP_Error` object instead of logging a `_doing_it_wrong()` and returning `null`

## Why

Refactor for #14 

## How

- Now that we can return `WP_Error`, `null` is a valid response from `execute()`, and the test has been updated accordingly. If we want, we could map `null` to an `ability_unknown_execution_error` or something.
- The only WP_Error that's _not_ bubbled out to the WP_REST_Response is the `WP_Ability::get_permissions()` error, so as not to leak a potentially sensitive error to a user without the proper permissions.

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -242,7 +242,7 @@ class WP_Ability {
 	 * @since 0.1.0
 	 *
 	 * @param mixed $output The output data to validate.
-	 * @return true|\WP_Error Returns true if valid, false if validation fails.
+	 * @return true|\WP_Error Returns true if valid, or a WP_Error object if validation fails.
 	 */
 	protected function validate_output( $output ) {
 		$output_schema = $this->get_output_schema();
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -180,32 +180,17 @@ class WP_Ability {
 	 * @since 0.1.0
 	 *
 	 * @param array<string,mixed> $input Optional. The input data to validate.
-	 * @return bool Returns true if valid, false if validation fails.
+	 * @return true|\WP_Error Returns true if valid or the WP_Error object if validation fails.
 	 */
-	protected function validate_input( array $input = array() ): bool {
+	protected function validate_input( array $input = array() ) {
 		$input_schema = $this->get_input_schema();
 		if ( empty( $input_schema ) ) {
 			return true;
 		}
 
 		$valid_input = rest_validate_value_from_schema( $input, $input_schema );
-		if ( is_wp_error( $valid_input ) ) {
-			_doing_it_wrong(
-				__METHOD__,
-				esc_html(
-					sprintf(
-						/* translators: %1$s ability name, %2$s error message. */
-						__( 'Invalid input provided for ability "%1$s": %2$s.' ),
-						$this->name,
-						$valid_input->get_error_message()
-					)
-				),
-				'0.1.0'
-			);
-			return false;
-		}
 
-		return true;
+		return is_wp_error( $valid_input ) ? $valid_input : true;
 	}
 
 	/**
@@ -216,11 +201,12 @@ class WP_Ability {
 	 * @since 0.1.0
 	 *
 	 * @param array<string,mixed> $input Optional. The input data for permission checking.
-	 * @return bool Whether the ability has the necessary permission.
+	 * @return true|\WP_Error Whether the ability has the necessary permission.
 	 */
-	public function has_permission( array $input = array() ): bool {
-		if ( ! $this->validate_input( $input ) ) {
-			return false;
+	public function has_permission( array $input = array() ) {
+		$is_valid = $this->validate_input( $input );
+		if ( is_wp_error( $is_valid ) ) {
+			return $is_valid;
 		}
 
 		if ( ! is_callable( $this->permission_callback ) ) {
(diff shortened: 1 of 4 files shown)


justlevine:
Looks like my IDE `phpcbf`d our tests file. Since all the other code PRs right now are mine and I don't mind eating any conflicts, im'a leave it as is unless told otherwise.

includes/abilities-api/class-wp-ability.php
```diff
@@ -180,32 +180,17 @@ public function get_meta(): array {
...
+	 * @return true|\WP_Error Returns true if valid or the WP_Error object if validation fails.
 	 */
-	protected function validate_input( array $input = array() ): bool {
+	protected function validate_input( array $input = array() ) {
 		$input_schema = $this->get_input_schema();
 		if ( empty( $input_schema ) ) {
 			return true;
 		}
 
 		$valid_input = rest_validate_value_from_schema( $input, $input_schema );
```
gziolo, line 201: It looks like it could return the value here because [rest_validate_value_from_schema](https://developer.wordpress.org/reference/functions/rest_validate_value_from_schema/) returns `true` or `WP_Error`.

gziolo (commented):
I left some initial feedback. Thank you for taking the time to refactor the code.

gziolo:
@emdashcodes, this could use some sanity check from you.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
```diff
@@ -146,19 +146,7 @@ public function run_ability( $request ) {
...
-		if ( is_null( $result ) ) {
-			return new \WP_Error(
-				'rest_ability_execution_failed',
-				__( 'Ability execution failed. Please check permissions and input parameters.', 'abilities-api' ),
-				array( 'status' => 500 )
-			);
+			return $result;
 		}
 
 		$output_validation = $this->validate_output( $ability, $result );
```
gziolo, line 152: Both `$this->validate_input()` and `$this->validate_output()` calls should no longer be needed as they are handled internally inside `$ability->execute()` with similar `WP_Errors`. We would need to differentiate mostly the error names to the proper `status` code in the response: 400 vs 500.

tests/unit/abilities-api/wpRegisterAbility.php
```diff
@@ -160,80 +160,90 @@ public function test_register_ability_no_permissions(): void {
...
 		$result = wp_register_ability( self::$test_ability_name, self::$test_ability_properties );
 
-		$this->assertNull(
-			$result->execute(
-				array(
-					'a'       => 2,
-					'b'       => 3,
-					'unknown' => 1,
-				)
+		$this->setExpectedIncorrectUsage( 'WP_Ability::execute' );
```
gziolo, line 185: Noting that `@expectedIncorrectUsage WP_Ability::execute` annotation plays the same role.

includes/abilities-api/class-wp-ability.php
gziolo: Thinking a bit more about it after seeing the updated unit tests, we should remap these WP_Errors so the names are more tied to the abilty, for example:

`rest_invalid_type` -> `ability_input_invalid_type` 

Could we replace the prefix?

includes/abilities-api/class-wp-ability.php
```diff
@@ -259,32 +242,17 @@ protected function do_execute( array $input ) {
...
-						$valid_output->get_error_message()
-					)
-				),
-				'0.1.0'
-			);
-			return false;
-		}
 
-		return true;
+		return is_wp_error( $valid_output ) ? $valid_output : true;
```
gziolo, line 265: Similar feedback as for input handling since we use `rest_validate_value_from_schema` which happens to be tied to much to REST API at this level.

tests/unit/abilities-api/wpRegisterAbility.php
justlevine: Specifically on this one test, the annotation way kept claiming there wasn't any `_doing_it_wrong()` coming from `::execute()` while continuing to warn me about the 'unexpected' `_doing_it_wrong`

gziolo: Interesting, it's fine as is 👍🏻

gziolo:
> Looks like my IDE phpcbfd our tests file. Since all the other code PRs right now are mine and I don't mind eating any conflicts, im'a leave it as is unless told otherwise.

I noticed similar issues when running `composer format` on `wordpress-develop`. I opened new PR to address that subset of changes:
- https://github.com/WordPress/abilities-api/pull/25

gziolo:
Resolving merge conflicts on mobile is hard but I think I made it 😆

I plan to land changes and look into my feedback separately. This PR largely addresses the issue but we still can add some final touches to reuse the builtin schema validation inside `execute` method inside the ability ti prevent double processing.

gziolo (approved):

Issue #32, opened 2025-08-20
Add Node.js version requirement to CONTRIBUTING.md


rollybueno:
The project has a .nvmrc file specifying Node.js version 20, but the CONTRIBUTING.md file doesn't mention the minimum Node.js version requirement or how to use the .nvmrc file.

justlevine:
Hey @rollybueno and thanks for flagging this, and for the accompanying PR!

The Node 20 requirement is mentioned in the [Local Setup: Prerequisites section](https://github.com/WordPress/abilities-api/blob/trunk/CONTRIBUTING.md#prerequisites), and I see from your PR you found the `nvm use` reference. 

I'm not sure how much handholding we need beyond that (for example, we don't offer steps for installing node 20 _without_ NVM), but as long as it doesn't confuse anything I guess it can't hurt.

Will leave some more specific feedback on the PR itself 🙇

jonathanbossenger:
As the node version requirement is included in the requirements section, I don't see that it needs to be repeated in the installation steps.

gziolo:
Closed with https://github.com/WordPress/abilities-api/pull/33.

PR #33, opened 2025-08-20
Documentation: Include specific nvm version needed

rollybueno:
**Current issue**: The nvm use command will fail if Node.js 20.10.0+ isn't already installed, causing setup errors for new contributors.

**Solution**: Specify the minimum version requirement and provide a single command that handles both installation and activation.

Issue: https://github.com/WordPress/abilities-api/issues/32

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,8 +49,8 @@ You can use Docker and the `wp-env` tool to set up a local development environme
 2. Change into the project folder and install the development dependencies:
 
    ```bash
-   ## Install Node.js 20.10.0 or higher if you don't have it
-   nvm install 20 && nvm use
+   ## If you're using nvm, make sure to use the correct Node.js version:
+   nvm install && nvm use
 
    ## Then install the NPM dependencies:
    npm install
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,8 +49,8 @@ You can use Docker and the `wp-env` tool to set up a local development environme
 2. Change into the project folder and install the development dependencies:
 
    ```bash
-   ## If you're using NVM, make sure to use the correct Node.js version:
-   nvm use
+   ## Install Node.js 20.10.0 or higher if you don't have it
+   nvm install 20 && nvm use
 
    ## Then install the NPM dependencies:
    npm install


CONTRIBUTING.md
```diff
@@ -49,8 +49,8 @@ You can use Docker and the `wp-env` tool to set up a local development environme
 2. Change into the project folder and install the development dependencies:
 
    ```bash
-   ## If you're using NVM, make sure to use the correct Node.js version:
-   nvm use
+   ## Install Node.js 20.10.0 or higher if you don't have it
```
justlevine: What is the goal with this comment change?
To me it implies NVM is no longer optional, and makes the focus on what Node version to install, when the whole point of NVM (if users _choose it_) is to not have to care about the node version.

CONTRIBUTING.md
```diff
@@ -49,8 +49,8 @@ You can use Docker and the `wp-env` tool to set up a local development environme
...
+   nvm install 20 && nvm use
```
justlevine: Wherever this goes, it should 
1. be `nvm install` (to use .nvmrc)
2. exist separately from `nvm use`.

justlevine (commented):
Hey @rollybueno 

Not entirely sure I understand your concern here ( `npm install` will fail if Node.js is not installed too, regardless of if someone uses NVM), so take the following feedback with a grain of salt, but I have misgivings about the solution in this PR.

Unlike `nvm install` which is meant to install the needed version _once_, `nvm use` is meant to be used _every time_ you spin up a local setup (pull/switch branches etc). 
That's why the node version is listed as a prereq  (like Composer and Docker), while `nvm use` is included in the instructions alongside `npm install`.

If there's a need to clarify how to use NVM to install node versions, I strongly suggest we find a different spot for it and not inside this bash script (similar to how we dont include a `cd path/to/abilities-api`) in this particular part of the docs..

Smaller feedback on the changes in the diff

CONTRIBUTING.md
gziolo: > The nvm use command will fail if Node.js 20.10.0+ isn't already installed, causing setup errors for new contributors.

That's the reason to include `nvm install 20` here. However, there is `.nvmrc` so it could be simplified to:

```bash
nvm install && nvm use
```

This way it will both install and switch to the version set in the config.

justlevine: Not asking about the goal of using `nvm install` here (I flag that else where), asking specifically about the clarity and intent of the comment itself.

gziolo: Right, the existing comment seems correct as long as we update the command that runs to install the version specified in `.nvmrc` if it is missing.

CONTRIBUTING.md
gziolo: @justlevine, is this how you would go about it:
```suggestion
   ## If you're using nvm, make sure to use the correct Node.js version:
   nvm install && nvm use
```

CONTRIBUTING.md
rollybueno: My fault. The comment in fact is confusing

rollybueno:
Sorry I've been away for a while and now just got back. I have encountered an error when I first run the original command since I don't have v20+ in my other laptop. I'm aware we have it under prerequisite, but the idea is to have a single command to install all dependencies in one go.. 

It's not a real blocker but a very small "improvement" for new contributors. If we think that the info under prerequisite is good enough and let the contributors to use existing command, we can close this.

CONTRIBUTING.md
rollybueno: I think he wanted to have it on 2 separate commands though single command should also work and more convenient.

justlevine: @gziolo The `[MD](link)` syntax wont work in the syntax highlighting, otherwise it's fine. (as @rollybueno , id personally prefer it two lines, but since our goal isnt to educate users on NVM, it doesn't matter)

gziolo (approved):
Let's land this small tweak that might help with contributing 👍🏻

PR #35, opened 2025-08-21
Chore: Configure the composer package for publishing

### LINKED ISSUE #13: Scaffold repository tooling

gziolo:
- Closes https://github.com/WordPress/abilities-api/issues/13.

Prepare the composer package for publication and consumption by WordPress plugins.

For complete transparency, I reused some of the configurations from:
- https://github.com/WordPress/php-ai-client/blob/trunk/composer.json
- https://github.com/WordPress/php-ai-client/blob/trunk/.gitattributes
- https://github.com/WordPress/gutenberg/blob/trunk/.gitattributes

I also used AI (GitHub Copilot coding agent) extensively to configure everything so it works both as a WP plugin and shares only `includes` through Composer. I avoided using namespaces to keep things simplified and used `autoload.files` to bootstrap all necessary files.

## Testing

Some checks that I executed

### Validate composer config

```bash
composer validate --no-check-publish
```

Result contained deprecation notices and a success message:

```
./composer.json is valid
```

### Files included in the git archive

```bash
echo "Total tracked files:" && git ls-files | wc -l && echo "\nFiles in git archive (respecting .gitattributes export-ignore):" && git archive --format=tar HEAD | tar -tf - | wc -l && echo "\nFiles excluded from archive (tracked but export-ignored):" && comm -23 <(git ls-files | sort) <(git archive --format=tar HEAD | tar -tf - | sort) | sed -n '1,200p' && echo "\nFiles explicitly in archive (sample):" && git archive --format=tar HEAD | tar -tf - | sed -n '1,200p'
```

Result

```
Total tracked files:
      40

Files in git archive (respecting .gitattributes export-ignore):
      22

Files excluded from archive (tracked but export-ignored):
.editorconfig
.gitattributes
.github/workflows/copilot-setup-steps.yml
.github/workflows/release.yml
.github/workflows/test.yml
.gitignore
.nvmrc
.prettierignore
.prettierrc.js
.wordpress-org/.gitkeep
.wp-env.json
composer.lock
package-lock.json
package.json
phpcs.xml.dist
phpstan.neon.dist
phpunit.xml.dist
tests/_output/.gitkeep
tests/bootstrap.php
tests/unit/abilities-api/wpAbilitiesRegistry.php
tests/unit/abilities-api/wpRegisterAbility.php
tests/unit/rest-api/wpRestAbilitiesInit.php
tests/unit/rest-api/wpRestAbilitiesListController.php
tests/unit/rest-api/wpRestAbilitiesRunController.php

Files explicitly in archive (sample):
.wordpress-org/
CONTRIBUTING.md
LICENSE.md
README.md
abilities-api.php
composer.json
docs/
docs/1.intro.md
docs/2.getting-started.md
docs/3.registering-abilities.md
docs/4.using-abilities.md
includes/
includes/abilities-api.php
includes/abilities-api/
includes/abilities-api/class-wp-abilities-registry.php
includes/abilities-api/class-wp-ability.php
includes/bootstrap.php
includes/rest-api/
includes/rest-api/class-wp-rest-abilities-init.php
includes/rest-api/endpoints/
includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -6,7 +6,6 @@
 
 # Tests and test fixtures
 /tests/ export-ignore
-/tests/** export-ignore
 
 # PHP unit / static analysis configs.
 /phpcs.xml.dist export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,45 @@
+# Automatically normalize line endings.
+* text=auto
+
+# Files and directories to exclude from the generated archive (export-ignore).
+# These will be omitted from GitHub/Packagist ZIP archives to keep packages small.
+
+# Tests and test fixtures
+/tests/ export-ignore
+/tests/** export-ignore
+
+# PHP unit / static analysis configs.
+/phpcs.xml.dist export-ignore
+/phpstan.neon.dist export-ignore
+/phpunit.xml.dist export-ignore
+
+# Continuous Integration configs.
+/.github/ export-ignore
+/.github/** export-ignore
+
+# Node and build artifacts.
+node_modules/ export-ignore
+node_modules/** export-ignore
+package.json export-ignore
+package-lock.json export-ignore
+
+# Composer/vendor and dev tooling.
+/vendor/ export-ignore
+/vendor/** export-ignore
+/composer.lock export-ignore
+
+# Don't include the dotfiles for distribution.
+/.editorconfig export-ignore
+/.gitattributes export-ignore
+.gitignore export-ignore
+.gitkeep export-ignore
+/.nvmrc export-ignore
+/.prettierignore export-ignore
+/.prettierrc.js export-ignore
+/.wp-env.json export-ignore
+
+# Mark docs as documentation for GitHub Linguist.
+/docs/** linguist-documentation
+/CONTRIBUTING.md linguist-documentation
+/LICENSE.md linguist-documentation
+/README.md linguist-documentation
--- a/.gitattributes
+++ b/.gitattributes
@@ -15,31 +15,17 @@
 
 # Continuous Integration configs.
 /.github/ export-ignore
-/.github/** export-ignore
 
 # Node and build artifacts.
-node_modules/ export-ignore
-node_modules/** export-ignore
 package.json export-ignore
 package-lock.json export-ignore
-
-# Composer/vendor and dev tooling.
-/vendor/ export-ignore
-/vendor/** export-ignore
 /composer.lock export-ignore
 
 # Don't include the dotfiles for distribution.
-/.editorconfig export-ignore
-/.gitattributes export-ignore
-.gitignore export-ignore
-.gitkeep export-ignore
-/.nvmrc export-ignore
-/.prettierignore export-ignore
-/.prettierrc.js export-ignore
-/.wp-env.json export-ignore
+/.* export-ignore
 
 # Mark docs as documentation for GitHub Linguist.
-/docs/** linguist-documentation
-/CONTRIBUTING.md linguist-documentation
+/docs/ linguist-documentation export-ignore
+/CONTRIBUTING.md linguist-documentation export-ignore
 /LICENSE.md linguist-documentation
 /README.md linguist-documentation
--- a/abilities-api.php
+++ b/abilities-api.php
@@ -26,17 +26,8 @@
  */
 define( 'WP_ABILITIES_API_DIR', plugin_dir_path( __FILE__ ) );
 
-/**
- * Prefer Composer autoloader when available. If present, it will load
- * `includes/bootstrap.php` (registered in composer.json) which makes the
- * procedural functions and classes available. Otherwise fall back to the
- * local includes bootstrap to keep compatibility with non-Composer installs.
- */
-if ( file_exists( WP_ABILITIES_API_DIR . 'vendor/autoload.php' ) ) {
-	require_once WP_ABILITIES_API_DIR . 'vendor/autoload.php';
-} else {
-	require_once WP_ABILITIES_API_DIR . 'includes/bootstrap.php';
-}
+
+require_once WP_ABILITIES_API_DIR . 'includes/bootstrap.php';
 
 if ( function_exists( 'add_action' ) ) {
 	add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
--- a/abilities-api.php
+++ b/abilities-api.php
@@ -27,26 +27,17 @@
 define( 'WP_ABILITIES_API_DIR', plugin_dir_path( __FILE__ ) );
 
 /**
- * Version of the plugin.
+ * Prefer Composer autoloader when available. If present, it will load
+ * `includes/bootstrap.php` (registered in composer.json) which makes the
+ * procedural functions and classes available. Otherwise fall back to the
+ * local includes bootstrap to keep compatibility with non-Composer installs.
  */
-define( 'WP_ABILITIES_API_VERSION', '0.1.0' );
+if ( file_exists( WP_ABILITIES_API_DIR . 'vendor/autoload.php' ) ) {
+	require_once WP_ABILITIES_API_DIR . 'vendor/autoload.php';
+} else {
+	require_once WP_ABILITIES_API_DIR . 'includes/bootstrap.php';
+}
 
-/**
- * First the WP_Ability class that users can extend.
- */
-require_once WP_ABILITIES_API_DIR . 'includes/abilities-api/class-wp-ability.php';
-
-/**
- * Then the WP_Abilities_Registry class that manages the abilities.
- */
-require_once WP_ABILITIES_API_DIR . 'includes/abilities-api/class-wp-abilities-registry.php';
-
-/**
- * Then the public access functions that users can use to interact with the abilities.
- */
-require_once WP_ABILITIES_API_DIR . 'includes/abilities-api.php';
-
-/**
- * Initialize REST API controllers.
- */
-require_once WP_ABILITIES_API_DIR . 'includes/rest-api/class-wp-rest-abilities-init.php';
+if ( function_exists( 'add_action' ) ) {
+	add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
+}
--- a/composer.json
+++ b/composer.json
@@ -1,33 +1,44 @@
 {
 	"name": "wordpress/abilities-api",
-	"description": "AI Abilities for WordPress",
+	"description": "AI Abilities for WordPress.",
 	"license": "GPL-2.0-or-later",
 	"type": "wordpress-plugin",
-	"version": "0.1.0",
 	"keywords": [
 		"ai",
+		"api",
 		"llm",
 		"abilities",
-		"abilities-api",
 		"wordpress"
 	],
-	"homepage": "https://github.com/WordPress/abilities-api/issues",
+	"authors": [
+		{
+			"name": "WordPress AI Team",
+			"homepage": "https://make.wordpress.org/ai/"
+		}
+	],
+	"homepage": "https://github.com/WordPress/abilities-api",
 	"support": {
-		"issues": "https://github.com/WordPress/abilities-api/issues"
+		"issues": "https://github.com/WordPress/abilities-api/issues",
+		"source": "https://github.com/WordPress/abilities-api"
+	},
+	"minimum-stability": "dev",
+	"prefer-stable": true,
+	"autoload": {
+		"files": [
+			"includes/bootstrap.php"
+		]
 	},
 	"config": {
-		"sort-packages": true,
-		"platform": {
-			"php": "7.4"
-		},
 		"allow-plugins": {
 			"dealerdirect/phpcodesniffer-composer-installer": true,
 			"phpstan/extension-installer": true,
 			"composer/installers": true
-		}
+		},
+		"platform": {
+			"php": "7.4"
+		},
+		"sort-packages": true
 	},
-	"minimum-stability": "dev",
-	"prefer-stable": true,
 	"repositories": [
 		{
 			"type": "composer",
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
 	"name": "wordpress/abilities-api",
 	"description": "AI Abilities for WordPress.",
 	"license": "GPL-2.0-or-later",
-	"type": "wordpress-plugin",
+	"type": "library",
 	"keywords": [
 		"ai",
 		"api",
--- a/composer.lock
+++ b/composer.lock
@@ -5169,9 +5169,9 @@
     "platform": {
         "php": "^7.4 | ^8"
     },
-    "platform-dev": {},
+    "platform-dev": [],
     "platform-overrides": {
         "php": "7.4"
     },
-    "plugin-api-version": "2.6.0"
+    "plugin-api-version": "2.2.0"
 }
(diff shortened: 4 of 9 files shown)


abilities-api.php
```diff
@@ -27,26 +27,17 @@
...
- * Then the public access functions that users can use to interact with the abilities.
- */
-require_once WP_ABILITIES_API_DIR . 'includes/abilities-api.php';
-
-/**
- * Initialize REST API controllers.
- */
-require_once WP_ABILITIES_API_DIR . 'includes/rest-api/class-wp-rest-abilities-init.php';
+if ( function_exists( 'add_action' ) ) {
+	add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
```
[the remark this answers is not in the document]
gziolo: The file is loaded earlier, so we can skip it.

composer.json
```diff
@@ -1,33 +1,44 @@
...
-		"sort-packages": true,
-		"platform": {
-			"php": "7.4"
-		},
 		"allow-plugins": {
 			"dealerdirect/phpcodesniffer-composer-installer": true,
 			"phpstan/extension-installer": true,
 			"composer/installers": true
-		}
+		},
```
gziolo, line 36: I moved a few things around, so the order is similar to https://github.com/WordPress/php-ai-client/blob/trunk/composer.json. That made auditing easier for me.

.gitattributes
```diff
@@ -0,0 +1,45 @@
...
+/phpstan.neon.dist export-ignore
+/phpunit.xml.dist export-ignore
+
+# Continuous Integration configs.
+/.github/ export-ignore
+/.github/** export-ignore
+
+# Node and build artifacts.
+node_modules/ export-ignore
+node_modules/** export-ignore
```
justlevine: Nit: there shouldn't be a need to exclude paths that are already excluded from the repo.

.gitattributes
```diff
@@ -0,0 +1,45 @@
...
+/.nvmrc export-ignore
+/.prettierignore export-ignore
+/.prettierrc.js export-ignore
+/.wp-env.json export-ignore
+
+# Mark docs as documentation for GitHub Linguist.
+/docs/** linguist-documentation
+/CONTRIBUTING.md linguist-documentation
+/LICENSE.md linguist-documentation
+/README.md linguist-documentation
```
justlevine: 1. Are we intentionally including docs?
Usually Composer libraries leave it out of the dist.

2. Either way it should be the same as `package.json:files`, so we should either add `CONTRIBUTING.md` there or also `export-ignore` it above in this file.

.gitattributes
```diff
@@ -0,0 +1,45 @@
...
+/tests/ export-ignore
+/tests/** export-ignore
+
+# PHP unit / static analysis configs.
+/phpcs.xml.dist export-ignore
+/phpstan.neon.dist export-ignore
+/phpunit.xml.dist export-ignore
+
+# Continuous Integration configs.
+/.github/ export-ignore
```
justlevine, line 16: we should either `export-ignore` `.wordpress-org` too (or just delete the dir entirely, since we're not planning to release this as a plugin so don't need that folder to store assets)

abilities-api.php
```diff
@@ -27,26 +27,17 @@
 define( 'WP_ABILITIES_API_DIR', plugin_dir_path( __FILE__ ) );
 
 /**
- * Version of the plugin.
+ * Prefer Composer autoloader when available. If present, it will load
+ * `includes/bootstrap.php` (registered in composer.json) which makes the
+ * procedural functions and classes available. Otherwise fall back to the
+ * local includes bootstrap to keep compatibility with non-Composer installs.
  */
```
justlevine: I'm thinking we can undo this (and the extra `includes/bootstrap.php`), as we don't need to use Composer's autoloader _internally_ in order for users to require/autoload it in their Composer projects.

(Might even be explicitly better _not_ to autoload and instead have the user `if ( ! class_exists('WP_Ability) ) { require_once 'path/to/vendor/wordpress/abilities-api/abilities-api.php'; }` to avoid conflicts.)

Will test downstream and report back (if nobody beats me to it).

.gitattributes
```diff
@@ -0,0 +1,45 @@
+# Automatically normalize line endings.
+* text=auto
+
+# Files and directories to exclude from the generated archive (export-ignore).
+# These will be omitted from GitHub/Packagist ZIP archives to keep packages small.
+
+# Tests and test fixtures
+/tests/ export-ignore
+/tests/** export-ignore
```
jeffpaul: I think the line above will cover for the second (aka I think we can remove this second line of `/tests/** export-ignore`)

.gitattributes
```diff
@@ -0,0 +1,45 @@
...
+
+# PHP unit / static analysis configs.
+/phpcs.xml.dist export-ignore
+/phpstan.neon.dist export-ignore
+/phpunit.xml.dist export-ignore
+
+# Continuous Integration configs.
+/.github/ export-ignore
+/.github/** export-ignore
```
jeffpaul: I think the line above will cover for the second (aka I think we can remove this second line of `/.github/** export-ignore`)

.gitattributes
jeffpaul: I think the line above will cover for the second (aka I think we can remove this second line of `/node_modules/** export-ignore`)

.gitattributes
```diff
@@ -0,0 +1,45 @@
...
+
+# Node and build artifacts.
+node_modules/ export-ignore
+node_modules/** export-ignore
+package.json export-ignore
+package-lock.json export-ignore
+
+# Composer/vendor and dev tooling.
+/vendor/ export-ignore
+/vendor/** export-ignore
```
jeffpaul: I think the line above will cover for the second (aka I think we can remove this second line of `/vendor/** export-ignore`)

.gitattributes
```diff
@@ -0,0 +1,45 @@
...
+/composer.lock export-ignore
+
+# Don't include the dotfiles for distribution.
```
jeffpaul, line 23: I think you can simplify this section to:
`/.* export-ignore`

jeffpaul:
The `composer.json` changes look ok to me, left some comments on the `.gitattributes` file where I think we can simplify it in a couple places.

justlevine (commented):
Quick test of `git archive` shows we're getting the desired outcome.

Some nits in the gitattributes that only matter if someone wants to use the file as prior art, but considering the 6.9 merge and that anybody using a release version and not a `dev-*` branch will get the GH release .zip, they can probably be ignored.

```sh
git archive --format=tar HEAD -o repo.tar && tar -tf repo.tar


.wordpress-org/
CONTRIBUTING.md
LICENSE.md
README.md
abilities-api.php
composer.json
docs/
docs/1.intro.md
docs/2.getting-started.md
docs/3.registering-abilities.md
docs/4.using-abilities.md
includes/
includes/abilities-api.php
includes/abilities-api/
includes/abilities-api/class-wp-abilities-registry.php
includes/abilities-api/class-wp-ability.php
includes/bootstrap.php
includes/rest-api/
includes/rest-api/class-wp-rest-abilities-init.php
includes/rest-api/endpoints/
includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
```

.gitattributes
justlevine: I dont think we need to export-ignore anything that's already `.gitignore`d

.gitattributes
justlevine: if it doesn't, then moving the `.gitkeep export-ignore` higher in the file should do the trick

felixarntz (commented):
@gziolo Overall this looks solid to me, though I have a high level question about the approach in https://github.com/WordPress/abilities-api/issues/13#issuecomment-3212006337 - why do we want to publish this on Packagist (vs having it as a plugin on .org and thus also on wpackagist.org)?

.gitattributes
gziolo: I will delete `.wordpress-org` for now.

abilities-api.php
gziolo: Updated with [98d19b0](https://github.com/WordPress/abilities-api/pull/35/commits/98d19b09aed031075697ec2782bd8c751dd8aca5).

.gitattributes
gziolo: Done in [7fd57a3](https://github.com/WordPress/abilities-api/pull/35/commits/7fd57a3b7f7f9a41490e0771fdc132272e1593ae).

.gitattributes
gziolo: I thought it would be helpful to include docs, but I'm happy to ignore the folder. We definitely don't want the want about contributing guidelines. Updated with [7fd57a3](https://github.com/WordPress/abilities-api/pull/35/commits/7fd57a3b7f7f9a41490e0771fdc132272e1593ae).

.gitattributes
gziolo: Make sense, updated with [7fd57a3](https://github.com/WordPress/abilities-api/pull/35/commits/7fd57a3b7f7f9a41490e0771fdc132272e1593ae).

.gitattributes
gziolo: It worked with [7fd57a3](https://github.com/WordPress/abilities-api/pull/35/commits/7fd57a3b7f7f9a41490e0771fdc132272e1593ae) 👍🏻

gziolo:
>  why do we want to publish this on Packagist (vs having it as a plugin on .org and thus also on wpackagist.org)?

I responded in the issue. The short version is that, for now, I was willing to replicate how WP Feature API was structured for distribution, and it turns out they offered a Composer library. I updated the config to reflect it in [55fab93](https://github.com/WordPress/abilities-api/pull/35/commits/55fab93e8b06c9a77022f8541b3f8917e5cfe48e).

I addressed all the feedback. This is what I see in the archive fter executing `git archive --format=tar HEAD -o repo.tar && tar -tf repo.tar`:

Screenshot 2025-08-22 at 11 34 18

Should we also exclude the top-level `abilities-api.php`, which is an entry point for the WP plugin?

composer.json
```diff
@@ -1,33 +1,44 @@
 {
 	"name": "wordpress/abilities-api",
-	"description": "AI Abilities for WordPress",
+	"description": "AI Abilities for WordPress.",
 	"license": "GPL-2.0-or-later",
-	"type": "wordpress-plugin",
-	"version": "0.1.0",
+	"type": "library",
```
justlevine, line 5: Strongly disagree with this change. Nothing prevents anybody from using a `wordpress-plugin` as a library. But if anybody (including MCP Adapter or Experiments) tries to test this using Composer as a plugin dependency, they can't if it's a `library`.

justlevine:
> Should we also exclude the top-level abilities-api.php, which is an entry point for the WP plugin?

Please give me time to test using `abilities-api.php` _instead_ of adding `bootstrap.php`, and some of the other changes here that IMO should be necessary for this to be used as a Composer library, don't make sense in our context (versus the WP Feature API prior art), and might even be detrimental to adoption and usage both in the next 3 months and in the post-6.9-merge cleanup.

emdashcodes:
> why do we want to publish this on Packagist (vs having it as a plugin on .org and thus also on wpackagist.org)?

> I responded in the issue. The short version is that, for now, I was willing to replicate how WP Feature API was structured for distribution, and it turns out they offered a Composer library. I updated the config to reflect it in [55fab93](https://github.com/WordPress/abilities-api/pull/35/commits/55fab93e8b06c9a77022f8541b3f8917e5cfe48e).

Our main goal for this in the Feature API was to make it possible for plugins to bundle the Feature API as a dependency, without the end user needing to install a separate plugin. This was the reason for the version loader as well.

justlevine:
Hey @gziolo et-al please take a look at https://github.com/WordPress/abilities-api/pull/42 which reverts the `"type": "library"` to `type:wordpress-plugin` and includes documentation showing how it can still be including as a Composer library (even if the Feature API considerations may not necessarily apply here).

composer.json
justlevine: https://github.com/WordPress/abilities-api/pull/42/files#r2294029457

Jameswlepage:
Felix, Jeff, and I had a chat about the wordpress-plugin vs library type discussion. Let's go with the library type to unblock things, mainly so other plugins like Elementor/Woo/Hosting MU etc can include this as a traditional package. This decision is made in part with the understanding that this will go core in 6.9, so we’re treating this as a transitional setup.

We recognize the concern that calling it a library may limit some Composer use cases (e.g., plugin dependencies). However, since wordpress-plugin can still be consumed as a library but not always vice versa, we’ll document usage patterns clearly so consumers understand the intended workflows. If needed, we can revisit the package type once the 6.9 merge has landed and the distribution model stabilizes.

For now, prioritizing unblock and compatibility with large plugin ecosystems outweighs the downside, and the configuration should make adoption easier in the short term while keeping a path open for refinement later.

justlevine:
>Let's go with the library type to unblock things, mainly so other plugins like Elementor/Woo/Hosting MU etc can include this as a traditional package. 

This is a false premise. Elementor/Woo/Hosting MU can all use something with a `"type": "wordpress-plugin"` same way they can if it's type "library". 

> However, since wordpress-plugin can still be consumed as a library but not always vice versa, we’ll document usage patterns clearly so consumers understand the intended workflows. 

To me this reads like a justification for _using_ `wordpress-plugin`. If it does everything that a `library` does, why go with the intentionally limiting one that requires additional documentation?

> For now, prioritizing unblock and compatibility with large plugin ecosystems outweighs the downside, and the configuration should make adoption easier in the short term while keeping a path open for refinement later.

This also seems like a false premise. What becomes "unblocked" or "compatible" by using "library" over "wordpress-plugin" , and how does the configuration make adoption easier in the short term?

gziolo:
I rebased this branch and applied a couple of small fixes based on the work shared by @justlevine in https://github.com/WordPress/abilities-api/pull/42. Taking into account the decision to go with type `library` in Composer, is there anything missing in this PR before we land it and create a git tag with `v0.1.0` release?

justlevine (approved):
> is there anything missing in this PR before we land it 

lgtm 🚀

> and create a git tag with v0.1.0 release?

Not strictly necessary for a v0.1.0, but do you want me to backport the relevant doc changes from #42 into a new PR?

gziolo:
> Not strictly necessary for a v0.1.0, but do you want me to backport the relevant doc changes from https://github.com/WordPress/abilities-api/pull/42 into a new PR?

That would be very helpful!

PR #37, opened 2025-08-21
Propose filters to use with the registered ability

### LINKED ISSUE #39: Proposal: Provide extensibility for individual abilities
## Filters

It should be possible to allow extenders to modify the behavior of registered abilities. I propose we make the following extensible:

- input schema – allow modifying the schema of the input passed to the action
- output schema – allow modifying the schema of the produced output
- permissions check – allow changing the condition for executing the ability
- execution result – allow modifying the output result

Example for how to revoke access to the ability:

```php
$filter_cb = static function ( $permission, $ability_name ) {
	if ( 'test/filter-ability' !== $ability_name ) {
		return $permission;
	}

	return false;
};

add_filter( 'ability_permission_result', $filter_cb, 10, 2 );
```

## Actions

It should also be possible to fire action before and after ability execution to allow developers to hook into the ability lifecycle for monitoring, logging, or extending functionality.

Example for how to log the succesful ability execution:

```php
$action_cb = static function ( $ability_name, $input, $result ) {
	if ( 'test/action-ability' !== $ability_name ) {
		return;
	}

	my_plugin_log_ability_result( $ability, $input, $result );
		
};

add_action( 'executed_ability', $action_cb, 10, 3 );
```

gziolo:
Closes #39.

This PR introduces filter support for the `WP_Ability` class, allowing developers to customize ability behavior through WordPress hook filters. The main purpose is to provide extensibility points for input/output schemas, permission checks, and execution results.

- Adds four new WordPress filters: `ability_input_schema`, `ability_output_schema`, `ability_permission_result`, and `ability_execute_result`.
- Includes comprehensive test coverage for all filter integration scenarios.
- Updates phpcs configuration to accommodate the new hook prefixes.

Example for how to revoke access to the ability:

```php
$filter_cb = static function ( $permission, $ability_name ) {
	if ( 'test/filter-ability' !== $ability_name ) {
		return $permission;
	}

	return false;
};

add_filter( 'ability_permission_result', $filter_cb, 10, 2 );
```

Example for how to change the output value:

```php
$filter_cb = static function ( $result, $ability_name ) {
	if ( 'test/filter-ability' !== $ability_name ) {
		return $result;
	}

	return 'modified-' . $result;
};

add_filter( 'ability_execute_result', $filter_cb, 10, 2 );
```

## Testing instructions

6 new unit tests were added to cover possible scenarios. You can run them with

`npm run test:php`

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -316,10 +316,10 @@ class WP_Ability {
 		 *
 		 * @since 0.1.0
 		 *
-		 * @param mixed               $result       The raw result from the execute callback (can be any type or \WP_Error).
+		 * @param mixed|\WP_Error     $result       The raw result from the execute callback (can be any type or \WP_Error).
 		 * @param string              $ability_name The ability name.
 		 * @param array<string,mixed> $input        The input passed to the ability.
-		 * @return mixed The filtered result.
+		 * @return mixed|\WP_Error The filtered result.
 		 */
 		return apply_filters( 'ability_execute_result', $result, $this->name, $input );
 	}
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -107,7 +107,7 @@ class WP_Ability {
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
 	 *   execute_callback: callable( array<string,mixed> $input): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( ?array<string,mixed> $input ): bool,
+	 *   permission_callback?: ?callable( ?array<string,mixed> $input ): true|\WP_Error,
 	 *   meta?: array<string,mixed>,
 	 *   ...<string, mixed>,
 	 * } $properties
@@ -177,7 +177,16 @@ class WP_Ability {
 	 * @return array<string,mixed> The input schema for the ability.
 	 */
 	public function get_input_schema(): array {
-		return $this->input_schema;
+		/**
+		 * Filters the input schema for a specific ability.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param array<string,mixed> $input_schema The input schema.
+		 * @param string              $ability_name The ability name.
+		 * @return array<string,mixed> The filtered input schema.
+		 */
+		return (array) apply_filters( 'ability_input_schema', $this->input_schema, $this->name );
 	}
 
 	/**
@@ -188,7 +197,16 @@ class WP_Ability {
 	 * @return array<string,mixed> The output schema for the ability.
 	 */
 	public function get_output_schema(): array {
-		return $this->output_schema;
+		/**
+		 * Filters the output schema for a specific ability.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param array<string,mixed> $output_schema The output schema.
+		 * @param string              $ability_name  The ability name.
+		 * @return array<string,mixed> The filtered output schema.
+		 */
+		return (array) apply_filters( 'ability_output_schema', $this->output_schema, $this->name );
 	}
 
 	/**
@@ -248,11 +266,26 @@ class WP_Ability {
 			return $is_valid;
 		}
 
-		if ( ! is_callable( $this->permission_callback ) ) {
-			return true;
+		$permission = true;
+		if ( is_callable( $this->permission_callback ) ) {
+			$permission = call_user_func( $this->permission_callback, $input );
 		}
 
-		return call_user_func( $this->permission_callback, $input );
+		/**
+		 * Filters the permission result for a specific ability.
+		 *
+		 * Allows plugins to override or short-circuit the permission check for an ability.
+		 * The filter receives the current permission result (true|\WP_Error), the
+		 * ability name, and the input provided for the permission check.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param true|\WP_Error      $permission   The current permission result.
+		 * @param string              $ability_name The ability name.
+		 * @param array<string,mixed> $input        The input for the ability.
+		 * @return true|\WP_Error The filtered permission result.
+		 */
+		return apply_filters( 'ability_permission_result', $permission, $this->name, $input );
 	}
 
 	/**
@@ -337,6 +370,21 @@ class WP_Ability {
 		}
 
 		$result = $this->do_execute( $input );
+		/**
+		 * Filters the raw result returned by the ability execute callback.
+		 *
+		 * Allows plugins to modify or short-circuit the execute callback result
+		 * before output validation occurs. The filter receives the raw result,
+		 * the ability name and the input provided to the ability.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param mixed               $result       The raw result from the execute callback (can be any type or \WP_Error).
+		 * @param string              $ability_name The ability name.
+		 * @param array<string,mixed> $input        The input passed to the ability.
+		 * @return mixed The filtered result.
+		 */
+		$result = apply_filters( 'ability_execute_result', $result, $this->name, $input );
 		if ( is_wp_error( $result ) ) {
 			return $result;
 		}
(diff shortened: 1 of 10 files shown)


tests/unit/abilities-api/wpAbilityFilters.php
```diff
@@ -0,0 +1,194 @@
...
+			}
+
+			return 'modified-' . $result;
+		};
+
+		add_filter( 'ability_execute_result', $filter_cb, 10, 2 );
+
+		$output = $ability->execute( array( 'value' => 'ok' ) );
+
+		remove_filter( 'ability_execute_result', $filter_cb, 10 );
```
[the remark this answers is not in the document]
gziolo: No need to pass the default priority. There is no 4th arg in the function.

tests/unit/abilities-api/wpAbilityFilters.php
```diff
@@ -0,0 +1,194 @@
...
+			}
+
+			return new \WP_Error( 'filtered_error', 'Filtered out' );
+		};
+
+		add_filter( 'ability_execute_result', $filter_cb, 10, 2 );
+
+		$output = $ability->execute( array( 'value' => 'ok' ) );
+
+		remove_filter( 'ability_execute_result', $filter_cb, 10 );
```
gziolo:
```suggestion
		remove_filter( 'ability_execute_result', $filter_cb );
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -254,7 +287,23 @@ protected function do_execute( array $input ) {
...
+		/**
+		 * Filters the raw result returned by the ability execute callback.
+		 *
+		 * Allows plugins to modify or short-circuit the execute callback result
+		 * before output validation occurs. The filter receives the raw result,
+		 * the ability name and the input provided to the ability.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @param mixed               $result       The raw result from the execute callback (can be any type or \WP_Error).
```
gziolo:
```suggestion
		 * @param mixed|\WP_Error     $result       The raw result from the execute callback (can be any type or \WP_Error).
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -254,7 +287,23 @@ protected function do_execute( array $input ) {
...
+		 * @param string              $ability_name The ability name.
+		 * @param array<string,mixed> $input        The input passed to the ability.
+		 * @return mixed The filtered result.
```
gziolo:
```suggestion
		 * @return mixed|\WP_Error The filtered result.
```

gziolo:
@jonathanbossenger, this is something to keep on the radar for inclusion in the docs in case we agree that this level of extensibility is expected.

justlevine:
@gziolo is there an expectation for people to be calling public WP_Ability methods more than once per lifecycle?

My two concerns are
1. Because this breaks SRP, folks extending the class (e.g. #21 ) now either need more boilerplate (or not implement the hooks).
2. The possibility of hooks being triggered multiple times introduces perf and behavior unpredictability.

gziolo:
> The possibility of hooks being triggered multiple times introduces perf and behavior unpredictability.

That's the reality in the WordPress land:

> "With the great power (of extensibility) comes great responsibility"

Some recent examples:

https://github.com/WordPress/wordpress-develop/blob/55644dc608d3a9c9d2541379d8a5b48d4b2c3d99/src/wp-includes/class-wp-block-type.php#L600-L617

https://github.com/WordPress/wordpress-develop/blob/55644dc608d3a9c9d2541379d8a5b48d4b2c3d99/src/wp-includes/class-wp-block-type.php#L619-L636

> Because this breaks SRP, folks extending the class (e.g. https://github.com/WordPress/abilities-api/pull/21 ) now either need more boilerplate (or not implement the hooks).

There are different audiences here:
- developers implementing abilities
- plugin developers and site admins who want to customize these abilities

If the author of the ability uses a custom class that extends `WP_Ability`, it's their responsibility to use parent methods to keep these hooks. However, they might not want that for some reason that we can't anticipate. This works both ways, so I think it's fine to leave these considerations to implementers. WordPress core will never register abilities that skip these hooks, making things unpredictable.

justlevine:
@gziolo my question wasn't critique, I was looking for some direction for code review 🙇

---

I don't think we disagree in premise. The nuance here is our extremely short timetable before merging. To use your example of `get_variations()`

- Variations were used in core long before that hook was introduced.
- The hook was introduced _intentionally_ and to meet specific needs: https://core.trac.wordpress.org/ticket/59969

Just because we don't have time to evolve it holistically doesn't mean that we can't design our initial API with intention and around specific use cases. And if we don't want to wait to merge these until we have a specific use case from e.g. MCP Adapter or AI Experiments, then we should at least take a bit here to consider the use cases/ code flows for these proposed hooks (instead of designing code to conform with non-rubberducked hooks _after the fact_). 


--- 

So with that context, and putting yourself into the mind of an "implementer" for a minute - how do _you_ envision these hooks to be used?

- Are there `public *` methods that are obviously gonna be called multiple times (where if the filter is repeated there will be detrimental). 
- Are the methods that contain these hooks the most likely going to be overloaded.
- And _conversely_, are there hooks that we _want_ to be triggered every time people use, or consider it a _good think_ if they need to actively take steps to reinclude the hook in their downstream `extends *`.

gziolo:
> So with that context, and putting yourself into the mind of an "implementer" for a minute - how do you envision these hooks to be used?

I'm mostly concerned with this proposal about making it possible to customize the abilities that will get registered through WordPress Core.

> Are there public * methods that are obviously gonna be called multiple times (where if the filter is repeated there will be detrimental).

Can you provide a real-life example of how often you anticipate the same hooks being called multiple times?

In the frontend context, WordPress core won't even process individual abilities because they are never consumed by default, so only the `abilitis_api_init` hooks get registered.

Inside REST API context, they should be called once per ability, depending on the endpoint type:
- get all would call input and out schema hooks to produce the result
- get single would call input and output schema hooks to produce the result
- execute, would call all 4 hooks, it could be that input and output schema hooks, when defined, would get called twice - during permissions check and during the execution

WP Admin is less predictable at this point, as we don't know how wide the usage will become. That said, WP hooks are everywhere in the codebase. I would be surprised learning that the abilities registry would have a larger impact than the rest of the codebase. What's the part that worries you the most?

justlevine:
> 
> > Are there public * methods that are obviously gonna be called multiple times (where if the filter is repeated there will be detrimental).
> 
> Can you provide a real-life example of how often you anticipate the same hooks being called multiple times?

The one that prompted my question is `$ability->has_permissions()` e.g

```php
add_filter( 'ability_permission_result', static function( $result ) {
  if ( $result instanceof \WP_Error ) {
    return $result;
  }
  
  return MyCustom::check_user_permissions_in_enterprise_sso( wp_get_current_user() );
}

... 

add_filter( 'ability_execute_result', static function ( $result, string $ability_name, array $input ) {
  if ( 'my/complex-tool-with-state-and-context' !== $ability_name ) {
    return $result;
  }
  
  $prev_ability = wp_get_ability( $input['prevAbilityId'] );
  if ( $prev_ability->has_permissions() ) {
    ...
  }
}
```


> What's the part that worries you the most?

Solely the timeline to core merge has us theorycrafting these, nothing blocking.

The combination of (a.) `public` getters that (b.) don't just "get" but break SRP to _do_ something, while (c.) the filters run on each call instead of caching and (d) we already have people hankering to `extends WP_Ability` makes this class a recipe for refactor (IMO), but it's hard to justify the extra complexity of "doing it right" ( e.g. separate `prepare_*()` methods, applying filters once etc ) in an initial release.

gziolo:
All valid concerns. We don’t have to rush introducing these filters, and wait for feedback from folks what’s limiting them. 

Aside, caching and filters is often incompatible in WordPress reality because there are two primary challenges:
- filters can be added or removed at different stages of page rendering, I saw many examples where filter gets added before executing the function and immediately removed after
- result from filters and functions are not guaranteed to be idempotent as they often depend on thr current global state

gziolo:
This is possible in two different ways:
- https://github.com/WordPress/abilities-api/pull/74 – by filtering the args passed during registration
- https://github.com/WordPress/abilities-api/pull/21 – by using `ability_class` to extend the functionality in the base `WP_Ability` class

Let's close this one as not planned for now.

PR #40, opened 2025-08-22
Remove `ignoreErrors` from PHPStan

gziolo:
- Part of https://github.com/WordPress/abilities-api/issues/13.

I audited `ignoreErrors` added to PHPStan (https://github.com/WordPress/abilities-api/pull/6#discussion_r2274466039). I decided to remove it and fix the PHPDoc to make PHPStan happy.

### DIFF
--- a/includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
+++ b/includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
@@ -89,7 +89,7 @@ class WP_REST_Abilities_List_Controller extends WP_REST_Controller {
 	 *
 	 * @since 0.1.0
 	 *
-	 * @param \WP_REST_Request $request Full details about the request.
+	 * @param \WP_REST_Request<array<string,mixed>> $request Full details about the request.
 	 * @return \WP_REST_Response Response object on success.
 	 */
 	public function get_items( $request ) {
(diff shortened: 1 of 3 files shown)


justlevine (approved):
@gziolo if you don't mind the extra work stripping them out before [merge into WP Core](https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop+WP_REST_Request&type=code&p=4) then go ahead and merge 🤷

(my 2c: We're not using nor relying on the real array shape, so this is just suppressing the external `phpstan-wordpress` error in the file by marking the generic they introduce as `mixed` so they don't get type-checked. tl;dr all this does is hide the error suppression inside the file implementation, instead of centrally declaring that we're not using the (3rd-party) generics for that file. ).

gziolo:
> if you don't mind the extra work stripping them out before [merge into WP Core](https://github.com/search?q=repo%3AWordPress%2Fwordpress-develop+WP_REST_Request&type=code&p=4) then go ahead and merge 🤷

I want to make sure all files are linted for now. I will figure out next week what to do if CI on wordpress-develop is largely unhappy. As an alternative, we might consider silencing only the errors related to generics.

justlevine:
> As an alternative, we might consider silencing only the errors related to generics.

FTR that _was_ the pre-merge behavior... only generics from `WP_REST_Request` in those `wp-rest/*` files were being ignored.

- https://github.com/WordPress/abilities-api/blob/7736905fc5254327fd354f17f00ce8fa5d47b2b7/phpstan.neon.dist#L34-L41

gziolo:
True. I will revisit next week and revert if that makes more sense 👍🏻

gziolo:
I tested in https://github.com/WordPress/wordpress-develop/pull/9410 all the latest changes from `trunk`, and we can safely use more complex types in PHPDoc.

PR #42, opened 2025-08-22
dev: Composer polish

justlevine:
## What

This PR polishes up #35 , making it usable as _both_ a WordPress plugin and a Composer library.

## How

1. Library type switched back to `wordpress-plugin`, allowing for plugin installation using `composer.json:extra`.
2. Removed the `composer.json:autoload` prop in favor of manually `require_once`, depending on if the user wants the REST API endpoints (and the rest of the plugin) or just the API classes and functions.
3. Restored the `WP_ABILITIES_API_VERSION` const for version targeting, moved into `bootstrap.php` 
4. Added usage docs for both plugin and dependency, composer and non-composer usage.

More specific comments inline.

## Additional notes

I'm not entirely sure the use case of wanting to `require_once` the classes but not hook the REST API endpoints. _Without it_, we dont need a separate "library vs plugin" bootstrap, devs can can just `require_once` the same root `abilities-api.php` as WordPress core would do if it was located in `wp-content/plugins` instead of `my-plugin/vendor/*`.

Left it in this PR to respect the Chesterton fence (though I'm like 95% sure it's just an artifact of from the prior art and not relevant here).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,33 +1,44 @@
 {
 	"name": "wordpress/abilities-api",
-	"description": "AI Abilities for WordPress",
+	"description": "AI Abilities for WordPress.",
 	"license": "GPL-2.0-or-later",
 	"type": "wordpress-plugin",
-	"version": "0.1.0",
 	"keywords": [
 		"ai",
+		"api",
 		"llm",
 		"abilities",
-		"abilities-api",
 		"wordpress"
 	],
-	"homepage": "https://github.com/WordPress/abilities-api/issues",
+	"authors": [
+		{
+			"name": "WordPress AI Team",
+			"homepage": "https://make.wordpress.org/ai/"
+		}
+	],
+	"homepage": "https://github.com/WordPress/abilities-api",
 	"support": {
-		"issues": "https://github.com/WordPress/abilities-api/issues"
+		"issues": "https://github.com/WordPress/abilities-api/issues",
+		"source": "https://github.com/WordPress/abilities-api"
+	},
+	"minimum-stability": "dev",
+	"prefer-stable": true,
+	"autoload": {
+		"files": [
+			"includes/bootstrap.php"
+		]
 	},
 	"config": {
-		"sort-packages": true,
-		"platform": {
-			"php": "7.4"
-		},
 		"allow-plugins": {
 			"dealerdirect/phpcodesniffer-composer-installer": true,
 			"phpstan/extension-installer": true,
 			"composer/installers": true
-		}
+		},
+		"platform": {
+			"php": "7.4"
+		},
+		"sort-packages": true
 	},
-	"minimum-stability": "dev",
-	"prefer-stable": true,
 	"repositories": [
 		{
 			"type": "composer",
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
 	"name": "wordpress/abilities-api",
 	"description": "AI Abilities for WordPress.",
 	"license": "GPL-2.0-or-later",
-	"type": "wordpress-plugin",
+	"type": "library",
 	"keywords": [
 		"ai",
 		"api",
--- a/composer.json
+++ b/composer.json
@@ -2,7 +2,7 @@
 	"name": "wordpress/abilities-api",
 	"description": "AI Abilities for WordPress.",
 	"license": "GPL-2.0-or-later",
-	"type": "library",
+	"type": "wordpress-plugin",
 	"keywords": [
 		"ai",
 		"api",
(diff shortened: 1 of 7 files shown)


docs/2.getting-started.md
```diff
@@ -1,12 +1,153 @@
 # 2. Getting Started
 
-## Installation Options
+## Installation
```
justlevine, line 3: Personally I'd make this a separate `INSTALLATION.md` doc, but because the doc-files are `#` prefixed, I didnt want to blow up the diff here.

docs/2.getting-started.md
```diff
@@ -1,12 +1,153 @@
...
 
-Currently, the Abilities API is available as a WordPress plugin. You can install it by cloning the repository into your `wp-content/plugins` directory.
+Until the Abilities API is merged into WordPress core, it must be installed before it can be used.
+
+### As a plugin
+
+The best and easiest way to try and use the Abilities API is to install it as a plugin by downloading the latest release from the [GitHub Releases page](https://github.com/WordPress/abilities-api/releases/latest).
```
justlevine, line 9: This requires us to cut a release (but so does composer). Can use the path to trunk in the interim, I just didn't want to rewrite something that's anyway needs to be rewritten by a competent docs writer

docs/2.getting-started.md
```diff
@@ -1,12 +1,153 @@
...
+
+```bash
+composer require wordpress/abilities-api
+```
+
+### As a dependency
+
+Plugin authors and developers may wish to rely on the Abilities API as a dependency in their own projects, before it is merged into core. You can do that in one of the following ways.
+
+#### As a Plugin Dependency (Recommended)
```
justlevine, line 68: Even without it being on .org (and therefore available to autoinstall), I believe this should _still_  be `Recommended` , since we want to ensure people are testing with the latest versions for the next few months, and not some potentially-deconflicted library.

But if you disagree, drop it.

docs/2.getting-started.md
```diff
@@ -1,12 +1,153 @@
...
+  "repositories": [
+    {
+      "type": "vcs",
+      "url": "https://github.com/WordPress/abilities-api.git"
+    }
+    // ... other repositories.
+  ],
+  // ... rest of your composer.json.
+}
+```
```
justlevine, line 101: Removed once we're on packagist.

includes/bootstrap.php
justlevine, line 1: Mainly just cleaning up the composer specific comments.

includes/bootstrap.php
```diff
@@ -1,16 +1,27 @@
...
-// Load core classes if they are not already defined (for non-Composer installs or direct includes).
+declare( strict_types = 1 );
+
+/**
+ * Plugin version.
+ *
+ * Useful to confirm you're using the latest version of the Abilities API.
+ */
+if ( ! defined( 'WP_ABILITIES_API_VERSION' ) ) {
+	define( 'WP_ABILITIES_API_VERSION', '0.1.0' );
```
justlevine, line 21: Not sure if this was intentionally removed, but we need it for MCP Adapter / Experiments for the next 3 months if we want to be able to allow this library to evolve.

docs/2.getting-started.md
```diff
@@ -1,12 +1,153 @@
...
+      "type": "vcs",
+      "url": "https://github.com/WordPress/abilities-api.git"
+    }
+    // ... other repositories.
+  ],
+  "extra": {
+    "installer-paths": {
+      // This should match your WordPress+Composer setup.
+      "wp-content/plugins/{$name}/": [
+          "type:wordpress-plugin"
```
justlevine, line 49: This is what we lose by making `type: library` instead of `wordpress-plugin`. We gain nothing.

composer.json
justlevine, line 1: Much better prior art for what we're doing here: https://github.com/wordpress/plugin-check

includes/bootstrap.php
```diff
@@ -1,16 +1,27 @@
...
- * Keep this file simple and side-effect free except for requiring the functions
- * file.
+ * This file will not be needed when the Abilities API is merged in Core.
+ *
+ * @package abilities-api
+ * @since 0.1.0
  */
 
-// Load core classes if they are not already defined (for non-Composer installs or direct includes).
+declare( strict_types = 1 );
```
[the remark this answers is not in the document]
justlevine: Said who?

docs/2.getting-started.md
justlevine, line 1: NOTE: This is only here for reviewers because there seems to be some confusion about WordPress Plugins and composer usage (also in MCP Adapter... ) so I wanted clear illustration I could point to. Would recommend @jonathanbossenger or someone else better than I to review before final merge.

composer.json
gziolo: That's helpful 👍🏻

docs/2.getting-started.md
gziolo: We can update that part after the package is available on Packagist.

PR #43, opened 2025-08-22
Update an example in Introduction & Overview document

jonathanbossenger:
Updates the example with something simpler


gziolo (approved):
Works for me 👍

PR #47, opened 2025-08-27
Add code quality improvements to code comments

gziolo:
Code quality improvements extracted from https://github.com/WordPress/abilities-api/pull/37.

CI checks should pass here.

CI checks should pass also for the code tested agains WordPress core in https://github.com/WordPress/wordpress-develop/pull/9410.

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -65,7 +65,7 @@ class WP_Ability {
 	 * The ability execute callback.
 	 *
 	 * @since 0.1.0
-	 * @var callable
+	 * @var callable( array<string,mixed> $input): (mixed|\WP_Error)
 	 */
 	protected $execute_callback;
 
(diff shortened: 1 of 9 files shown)


includes/abilities-api/class-wp-ability.php
```diff
@@ -240,7 +240,7 @@ protected function validate_input( array $input = array() ) {
 	 * @since 0.1.0
 	 *
 	 * @param array<string,mixed> $input Optional. The input data for permission checking.
-	 * @return true|\WP_Error Whether the ability has the necessary permission.
+	 * @return bool|\WP_Error Whether the ability has the necessary permission.
```
justlevine, line 243: Is this an intentional change (to allow users to return `false` instead of an explicit WP_Error)?

gziolo: Yes, I forgot to note that I was checking some usage and it’s a nice shortcut to use an existing WP functions for capabilities verification. Many of them use internally [current_user_can()](https://developer.wordpress.org/reference/functions/current_user_can/) which returns `bool`. That’s why I wanted to make it compatible.

justlevine (approved):
lgtm

(To accommodate the dual falsy `false|WP_Error` return values, we're gonna want to make sure that usage and docs use `if ( true === $ability->has_permissions() ) ` and not e.g. the shorthand `if ( $ability->has_permissions() )` which will also return true for a `WP_Error`.)

PR #48, opened 2025-08-28
docs: add installation steps for release

justlevine:
## What

This PR documents the installation steps for the v0.1.0 release on GH + Composer

## How
Cherry picked from #42 and altered to reflect the latest `trunk` after #35

### DIFF
--- a/docs/2.getting-started.md
+++ b/docs/2.getting-started.md
@@ -1,12 +1,93 @@
 # 2. Getting Started
 
-## Installation Options
+## Installation
 
-Currently, the Abilities API is available as a WordPress plugin. You can install it by cloning the repository into your `wp-content/plugins` directory.
+Until the Abilities API is merged into WordPress core, it must be installed before it can be used.
+
+### As a plugin
+
+The easiest way to try and use the Abilities API is to install it as a plugin by downloading the latest release from the [GitHub Releases page](https://github.com/WordPress/abilities-api/releases/latest).
+
+#### With WP-CLI
 
 ```bash
-cd wp-content/plugins
-git clone https://github.com/WordPress/abilities-api
+wp plugin install https://github.com/WordPress/abilities-api/releases/latest/download/abilities-api.zip
+```
+
+#### With WP-Env
+
+```jsonc
+// .wp-env.json
+{
+  "$schema": "https://schemas.wp.org/trunk/wp-env.json",
+  // ... other config ...
+  "plugins": [
+    "WordPress/abilities-api",
+    // ... other plugins ...
+  ],
+  // ... more config ...
+}
+```
+
+### As a dependency
+
+Plugin authors and developers may wish to rely on the Abilities API as a dependency in their own projects, before it is merged into core. You can do that in one of the following ways.
+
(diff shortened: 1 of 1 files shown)


docs/2.getting-started.md
```diff
@@ -1,12 +1,93 @@
...
+
+```diff
+# my-plugin.php
+/*
+ *
+ * Plugin Name:       My Plugin
+ * Plugin URI:        https://example.com/plugins/the-basics/
+ * Description:       Handle the basics with this plugin.
+ * {all the other plugin header fields...}
++ * Requires Plugins:  abilities-api
```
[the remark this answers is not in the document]
justlevine: Great contextual learning copilot 🤦

docs/2.getting-started.md
```diff
@@ -1,12 +1,93 @@
...
+ */
+```
+
+While this is enough to ensure the Abilities API is loaded before your plugin, if you need to ensure specific version requirements or provide users guidance on installing the plugin, you can use the methods described [later on](#checking-availability-with-code)
+
+#### As a Composer dependency
+
+```bash
+composer require wordpress/abilities-api
+```
```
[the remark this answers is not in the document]
justlevine: Didn't bother to read the PR description this is literally the reason for the PR 🤦

docs/2.getting-started.md
```diff
@@ -1,12 +1,93 @@
...
 
-## Installation Options
+## Installation
 
-Currently, the Abilities API is available as a WordPress plugin. You can install it by cloning the repository into your `wp-content/plugins` directory.
+Until the Abilities API is merged into WordPress core, it must be installed before it can be used.
+
+### As a plugin
+
+The easiest way to try and use the Abilities API is to install it as a plugin by downloading the latest release from the [GitHub Releases page](https://github.com/WordPress/abilities-api/releases/latest).
```
jonathanbossenger, line 9: I assume we will have a release available soon, because right now this is not available.

docs/2.getting-started.md
```diff
@@ -1,12 +1,93 @@
...
+
+While this is enough to ensure the Abilities API is loaded before your plugin, if you need to ensure specific version requirements or provide users guidance on installing the plugin, you can use the methods described [later on](#checking-availability-with-code)
+
+#### As a Composer dependency
+
+```bash
+composer require wordpress/abilities-api
+```
+
+### Checking availability with code
```
jonathanbossenger, line 60: One note on these checks, if the code that's doing the checking is inside of a plugin that loads after wp-abilities-api (eg wp-my-plugin) then this might not work. It's on the lower end of the scale of the realm of possibility, but still possible enough to warrant a note about it. I experienced this recently when setting up my workshop examples after recent updates to the mcp-adapter.

jonathanbossenger (commented):
Added a few suggestions, but nothing that's really blocking a merge. We can always tweak things as we go.

docs/2.getting-started.md
justlevine: Yup, I believe this is the last PR before we cut v0.1.0

gziolo (approved):

PR #49, opened 2025-08-29
Updated Inline Documentation Order

shail-mehta:
Updated Inline Documentation Order. (Added `@since` tag before `@see` tag according to [PHP Documentation Standards
](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/)

**Files:** 
- `includes/abilities-api/class-wp-abilities-registry.php`
- `includes/abilities-api/class-wp-ability.php`


gziolo (approved):
Good catch 👍

shail-mehta:
Thank you for the review.

PR #54, opened 2025-09-02
dev!: handle property registration inside WP_Ability

### LINKED ISSUE #53: Improve the properties validation when using `ability_class` during registration
I left a note [here](https://github.com/WordPress/abilities-api/pull/21#issuecomment-3222854123) about making `execute_callback` optional when `ability_class` is passed as one of the properties:

> `invoke` on `Content_Rewrite_Tool` would have to be static to fit here because `execute_callback` is mandatory at the moment during ability registration (label and description, too). If that is too limiting, we could revisit `$properties` validation. I also think that we should make `$properties` mandatory, as you still must provide these 3 properties, so validation passes before the ability instance gets created.

I have an example from @felixarntz of using `ability_class` from the [AI chatbot demo using the PHP AI Client SDK](https://github.com/felixarntz/wp-ai-sdk-chatbot-demo/tree/main) presented at WC US:

https://github.com/felixarntz/wp-ai-sdk-chatbot-demo/blob/326266fd62fc805ceac0dfd4fe71cd3cc3e7cad3/includes/Abilities/Abilities_Registrar.php#L26-L44

Image

It's where not only the `execute_callback` was mocked, but also the description.

## Proposal

- Make `$properties`  mandatory.
- Make `execute_callback` optional when `ability_class` gets passed.
- Consider making `label` and `description` optional when `ability_class` gets passed, with the assumption that the developer will set it in the custom class extending `WP_Ability`.

justlevine:
## What

Refactors `WP_Ability` to validate its own properties.

## How 

`WP_Ability::validate_properties()` throws an exception, which is caught and translated into a `_doing_it_wrong()` by `WP_Abilities_API::register()` which was previously handled registration and validation.

More specific implementation notes are in the diff.

## Why

- Fixes #53 

This approach improves separation of concerns and SRP between `WP_Ability` and `WP_Ability_Registry` in a way that also simplifies the DX for users needing to extend the `WP_Ability` class. Instead of needing to shim instantiation to comply or circumvent `WP_Ability_Registry`, developers can overload `WP_Ability::validate_properties()` colocated with the other changes that justify extending the class, or even choose to bypass it entirely in their overloaded constructor.

## Beyond the Scope

- Renaming references of ability `properties` to `args` now that there's no semantic implication from the distinction.
- Auditing the validation logic, error messages, or use of `_doing_it_wrong()` over other flavors of WordPress error handling
- Backfilling WPUnit tests for previously committed validation logic..

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -100,20 +100,11 @@ class WP_Ability {
 	 * @param array<string,mixed> $properties An associative array of properties for the ability. This should
 	 *                                        include `label`, `description`, `input_schema`, `output_schema`,
 	 *                                        `execute_callback`, `permission_callback`, and `meta`.
-	 *
-	 * @phpstan-param array{
-	 *   label: string,
-	 *   description: string,
-	 *   input_schema?: array<string,mixed>,
-	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback: callable( array<string,mixed> $input): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( array<string,mixed> $input ): (bool|\WP_Error),
-	 *   meta?: array<string,mixed>,
-	 *   ...<string, mixed>,
-	 * } $properties
 	 */
 	public function __construct( string $name, array $properties ) {
 		$this->name = $name;
+
+		$this->validate_properties( $properties );
 
 		foreach ( $properties as $property_name => $property_value ) {
 			if ( ! property_exists( $this, $property_name ) ) {
@@ -200,6 +191,76 @@ class WP_Ability {
 	 */
 	public function get_meta(): array {
 		return $this->meta;
+	}
+
+	/**
+	 * Validates the properties used to instantiate the ability.
+	 *
+	 * Errors are thrown as exceptions instead of \WP_Errors to allow for simpler handling and overloading. They are then
+	 * caught and converted to a WP_Error when by WP_Abilities_Registry::register().
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @see WP_Abilities_Registry::register()
+	 *
+	 * @param array<string,mixed> $properties An associative array of properties to validate.
+	 *
+	 * @return void
+	 * @throws \InvalidArgumentException if the properties are invalid.
+	 *
+	 * @phpstan-assert array{
+	 *   label: string,
+	 *   description: string,
+	 *   input_schema?: array<string,mixed>,
+	 *   output_schema?: array<string,mixed>,
+	 *   execute_callback: callable( array<string,mixed> $input): (mixed|\WP_Error),
+	 *   permission_callback?: ?callable( array<string,mixed> $input ): (bool|\WP_Error),
+	 *   meta?: array<string,mixed>,
+	 *   ...<string, mixed>,
+	 * } $properties
+	 */
+	protected function validate_properties( array $properties ) {
+		if ( empty( $properties['label'] ) || ! is_string( $properties['label'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties must contain a `label` string.' )
+			);
+		}
+
+		if ( empty( $properties['description'] ) || ! is_string( $properties['description'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties must contain a `description` string.' )
+			);
+		}
+
+		if ( isset( $properties['input_schema'] ) && ! is_array( $properties['input_schema'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties should provide a valid `input_schema` definition.' )
+			);
+		}
+
+		if ( isset( $properties['output_schema'] ) && ! is_array( $properties['output_schema'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties should provide a valid `output_schema` definition.' )
+			);
+		}
+
+		if ( empty( $properties['execute_callback'] ) || ! is_callable( $properties['execute_callback'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties must contain a valid `execute_callback` function.' )
+			);
+		}
+
+		if ( isset( $properties['permission_callback'] ) && ! is_callable( $properties['permission_callback'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties should provide a valid `permission_callback` function.' )
+			);
+		}
+
+		if ( isset( $properties['meta'] ) && ! is_array( $properties['meta'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties should provide a valid `meta` array.' )
+			);
+		}
 	}
 
 	/**
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -85,69 +85,7 @@ final class WP_Abilities_Registry {
 			return null;
 		}
 
-		if ( empty( $properties['label'] ) || ! is_string( $properties['label'] ) ) {
-			_doing_it_wrong(
-				__METHOD__,
-				esc_html__( 'The ability properties must contain a `label` string.' ),
-				'0.1.0'
-			);
-			return null;
-		}
-
-		if ( empty( $properties['description'] ) || ! is_string( $properties['description'] ) ) {
-			_doing_it_wrong(
-				__METHOD__,
-				esc_html__( 'The ability properties must contain a `description` string.' ),
-				'0.1.0'
-			);
-			return null;
-		}
-
-		if ( isset( $properties['input_schema'] ) && ! is_array( $properties['input_schema'] ) ) {
-			_doing_it_wrong(
-				__METHOD__,
-				esc_html__( 'The ability properties should provide a valid `input_schema` definition.' ),
-				'0.1.0'
-			);
-			return null;
-		}
-
-		if ( isset( $properties['output_schema'] ) && ! is_array( $properties['output_schema'] ) ) {
-			_doing_it_wrong(
-				__METHOD__,
-				esc_html__( 'The ability properties should provide a valid `output_schema` definition.' ),
-				'0.1.0'
-			);
-			return null;
-		}
-
-		if ( empty( $properties['execute_callback'] ) || ! is_callable( $properties['execute_callback'] ) ) {
-			_doing_it_wrong(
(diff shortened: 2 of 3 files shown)


includes/abilities-api/class-wp-ability.php
```diff
@@ -100,21 +100,12 @@ class WP_Ability {
...
-	 *   description: string,
-	 *   input_schema?: array<string,mixed>,
-	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback: callable( array<string,mixed> $input): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( array<string,mixed> $input ): (bool|\WP_Error),
-	 *   meta?: array<string,mixed>,
-	 *   ...<string, mixed>,
-	 * } $properties
 	 */
 	public function __construct( string $name, array $properties ) {
```
justlevine, line 104: A small part of me thinks that we should change all the references of `properties` to `args` to bring it inline with other WP core naming, now that we've broken the direct dependency to add a validator.  🤷

I could add it in this PR but i didnt want to obfuscate the discussion on #53

includes/abilities-api/class-wp-ability.php
```diff
@@ -100,21 +100,12 @@ class WP_Ability {
...
 		$this->name = $name;
 
+		$this->validate_properties( $properties );
+
 		foreach ( $properties as $property_name => $property_value ) {
```
justlevine, line 109: An alternative approach is to replace the `validate_*()` pattern with the `prepare_*(): array|WP_Error()` , and then throw in the constructor.


This would improve the DX both mapping and validation in the same step, but I'm not sure how flexible we want things to be at this initial stage vs once we've given the initial API + round of feedback time to percolate, so I went with the approach in the diff.


e.g. (pseudocode)
```php
$properties = $this->prepare_args( $args );

if ( is_wp_error( $properties ) ) {
  throw \InvalidArgumentException( $properties->getMessage() );
}

// This is still outside the function so extenders don't need to reimplement it.
foreach( $properties as $name => $value ) {
  ...
}
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -202,6 +193,76 @@ public function get_meta(): array {
...
+	 *   description: string,
+	 *   input_schema?: array<string,mixed>,
+	 *   output_schema?: array<string,mixed>,
+	 *   execute_callback: callable( array<string,mixed> $input): (mixed|\WP_Error),
+	 *   permission_callback?: ?callable( array<string,mixed> $input ): (bool|\WP_Error),
+	 *   meta?: array<string,mixed>,
+	 *   ...<string, mixed>,
+	 * } $properties
+	 */
+	protected function validate_properties( array $properties ) {
```
justlevine, line 222: The only changes here are that they're now `InvalidArgumentExceptions()` instead of `_doing_it_wrong()`. The conditionals and error message are identical to what was in `WP_Abilities_API::register()`

includes/abilities-api/class-wp-ability.php
```diff
@@ -202,6 +193,76 @@ public function get_meta(): array {
...
+		if ( empty( $properties['label'] ) || ! is_string( $properties['label'] ) ) {
```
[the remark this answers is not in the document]
justlevine: 1. It's not a valid label (semanticly)
2. These are preexisting conditions and beyond the scope of this PR to audit. They've only been relocated with their logic preserved.

includes/abilities-api/class-wp-ability.php
```diff
@@ -202,6 +193,76 @@ public function get_meta(): array {
...
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties must contain a `label` string.' )
+			);
+		}
+
+		if ( empty( $properties['description'] ) || ! is_string( $properties['description'] ) ) {
```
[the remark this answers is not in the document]
justlevine: 1. It's not a valid label (semanticly)
2. These are preexisting conditions and beyond the scope of this PR to audit. They've only been relocated with their logic preserved.

includes/abilities-api/class-wp-ability.php
```diff
@@ -202,6 +193,76 @@ public function get_meta(): array {
...
+			);
+		}
+
+		if ( isset( $properties['output_schema'] ) && ! is_array( $properties['output_schema'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties should provide a valid `output_schema` definition.' )
+			);
+		}
+
+		if ( empty( $properties['execute_callback'] ) || ! is_callable( $properties['execute_callback'] ) ) {
```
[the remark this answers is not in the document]
justlevine: 1. It's not a valid label (semanticly)
2. These are preexisting conditions and beyond the scope of this PR to audit. They've only been relocated with their logic preserved.

includes/abilities-api/class-wp-ability.php
gziolo: We have now `ability_class` which isn't strictly a property of the object, so I don't mind changing to `args` at this point. If you want to take care of it, let's put it into a separate PR to keep the current refactor lean.

includes/abilities-api/class-wp-ability.php
gziolo: I double-checked with the implementation @felixarntz used in the demo plugin and it seems to be compatible even with the strict checks for madatory properties: label, description and execute callback.

https://github.com/felixarntz/wp-ai-sdk-chatbot-demo/blob/326266fd62fc805ceac0dfd4fe71cd3cc3e7cad3/includes/Abilities/Abstract_Ability.php#L32-L48

gziolo (commented):
I have just realized that `$name` is passed as a first argument so it won't be possible to make it optional with the intent that implementer extending `WP_Abilities` defines it similar to label, description, and the execute callback. That's perfectly fine. With the changes included, the code from the demo plugin will be simplified nicely:

```diff
wp_register_ability(
			'wp-ai-sdk-chatbot-demo/get-post',
			array(
-				'label'            => __( 'Get Post', 'wp-ai-sdk-chatbot-demo' ),
-				'description'      => 'Mock description.',
-				'execute_callback' => $mock_execute_callback,
				'ability_class'    => Get_Post_Ability::class,
			)
		);
```

gziolo:
Let's see what feedback @felixarntz has to share before landing these changes.

includes/abilities-api/class-wp-ability.php
justlevine: @gziolo / @felixarntz would love some specific thoughts on this if you have them. The more I think about it the more I feel like `protected function prepare_properties( array<string,mixed> ): array<validated-shape>|WP_Error` is the better approach 🤔

gziolo: Ok, so you prefer to have `prepare_properties` that returns `WP_Error` as soon as something goes wrong, and it gets translated to an exception, or the properties returned gets passed to the loop. That sounds good to me, to avoid having a method that only throws an exception when someting goes wrong. In fact, it might be simpler to ever introduce a filter if folks want to change this properties as an alternative to what I proposed here:
- https://github.com/WordPress/abilities-api/pull/37

The filter used with block types `register_block_type_args` comes to my find as a good reference:

https://github.com/WordPress/wordpress-develop/blob/9b6c234bc6a78ce4a57929ea5413d0b08a3d706e/src/wp-includes/class-wp-block-type.php#L565-L573

justlevine: My thoughts exactly. After the chat, I'll push a change with that approach.

includes/abilities-api/class-wp-ability.php
felixarntz: +1 to changing this to use `$args`.

includes/abilities-api/class-wp-ability.php
felixarntz: I'm not strongly opposed, but I also don't see the benefit of using `prepare_properties` and returning a `WP_Error` from it if something is wrong. Can you clarify why that's better?

To me it seems unnecessarily complex to rely on a method that can return `WP_Error`, only to turn that into an `InvalidArgumentException`, only to catch that and turn it into a `_doing_it_wrong()`.

Why not simply throw the exceptions?

felixarntz (approved):
@justlevine This looks great, thank you! I left a few comments on the open discussion points (I'm slightly favoring simply throwing exceptions and keeping the code as it is now), but I'll preemptively approve, since either way it's not a blocker.

felixarntz:
Once this is merged and available in a release, I'll update the demo plugin :)

includes/abilities-api/class-wp-ability.php
justlevine: Clarifying that my focus is on returning the array of (prepared) args instead of `void`. 😅

@@felixarntz I agree with you from a code quality POV

From an implementer POV my thought was less "exceptions should be exceptional" and more "let's do it the WP™️ way 💪" :

- most the other parts of the exposed API return WP_Errors to be handled instead of throwing.
- In general WordPress developers are more comfortable returning error objects than throwing.
- **most importantly**: it gives me a justification to leave the return type off the method signature (since php7.4 doesn't support union return types) since I didn't want to argue about why a legacy project wanting to migrate to this API but keep their existing DTO object isn't a justifiable use case for polymorphic tech debt at this early stage. 😅

(I drafted this before you both aligned yesterday on #53 and only saw after I pushed, so in my head this was still very much a proposal ).

My takeaway from this conversation is that we're good not caring about that last point (between overloading and the filter if someone really, _realy_ thinks shimming a DTO or VO into this api is a good idea, they have ways), so if y'all don't think it's outweighed by the first two (I don't), I'll use `prepare_*( array<string,mixed> ): array<valid-shape>` and just `throw`

gziolo:
As I start my day, I'll draft the refactoring described in https://github.com/WordPress/abilities-api/pull/54#discussion_r2319819580. I landed this PR so all the essential changes are included in the next package release as soon as possible.

justlevine:
@gziolo feel free to lift https://github.com/justlevine/abilities-api/tree/dev/wp_ability-prepare_args I didnt have time to review it before passing out last night, but might save you a few minutes

gziolo:
Oh, nice. I see you started the process of renaming to `$args`. I will employ some coding agent to finish it 😄

gziolo:
> @gziolo feel free to lift https://github.com/justlevine/abilities-api/tree/dev/wp_ability-prepare_args I didnt have time to review it before passing out last night, but might save you a few minutes

I continue based on the commit from your branch in https://github.com/WordPress/abilities-api/pull/59.

includes/abilities-api/class-wp-ability.php
gziolo: See https://github.com/WordPress/abilities-api/pull/54#issuecomment-3258015189.

PR #55, opened 2025-09-03
Fix REST API initialization for composer package usage

budzanowski:
## Summary
- Fixes REST API routes not being initialized when abilities-api is used as a composer package
- Moves REST API route registration from plugin file to bootstrap.php to ensure initialization in all cases
- Prevents duplicate registration by placing hook inside class existence check

## Changes
- Moved `rest_api_init` hook registration from `abilities-api.php` to `includes/bootstrap.php`
- Consolidated REST API initialization logic in bootstrap file
- Ensures REST API routes work for both plugin and composer package installations

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/includes/bootstrap.php
+++ b/includes/bootstrap.php
@@ -34,4 +34,9 @@ if ( ! function_exists( 'wp_register_ability' ) ) {
 // Load REST API init class for plugin bootstrap.
 if ( ! class_exists( 'WP_REST_Abilities_Init' ) ) {
 	require_once __DIR__ . '/rest-api/class-wp-rest-abilities-init.php';
+	
+	// Initialize REST API routes when WordPress is available.
+	if ( function_exists( 'add_action' ) ) {
+		add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
+	}
 }
--- a/includes/bootstrap.php
+++ b/includes/bootstrap.php
@@ -13,6 +13,10 @@
 
 declare( strict_types = 1 );
 
+if ( ! defined( 'ABSPATH' ) ) {
+	return; // Not in WordPress context
+}
+
 // Version of the plugin.
 if ( ! defined( 'WP_ABILITIES_API_VERSION' ) ) {
 	define( 'WP_ABILITIES_API_VERSION', '0.1.0' );
--- a/includes/bootstrap.php
+++ b/includes/bootstrap.php
@@ -34,7 +34,7 @@ if ( ! function_exists( 'wp_register_ability' ) ) {
 // Load REST API init class for plugin bootstrap.
 if ( ! class_exists( 'WP_REST_Abilities_Init' ) ) {
 	require_once __DIR__ . '/rest-api/class-wp-rest-abilities-init.php';
-	
+
 	// Initialize REST API routes when WordPress is available.
 	if ( function_exists( 'add_action' ) ) {
 		add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -34,7 +34,8 @@ require_once $_test_root . '/includes/functions.php';
 tests_add_filter(
 	'muplugins_loaded',
 	static function (): void {
-		require_once dirname( __DIR__ ) . '/abilities-api.php';
+		// Require ( to bypass require_once ).
+		require dirname( __DIR__ ) . '/includes/bootstrap.php';
 	}
 );
 
--- a/abilities-api.php
+++ b/abilities-api.php
@@ -28,7 +28,3 @@ define( 'WP_ABILITIES_API_DIR', plugin_dir_path( __FILE__ ) );
 
 
 require_once WP_ABILITIES_API_DIR . 'includes/bootstrap.php';
-
-if ( function_exists( 'add_action' ) ) {
-	add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
-}


gziolo:
PHPUnit tests fail. It's very likely that `add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );` will have to be called explicitly in the bootstrap file to account for the fact that vendor autoloading happens before any WP code is loaded.

gziolo (approved):
This looks like a good compromise to me. Thank you for proposing the enhancement.

@justlevine, I would appreciate your sanity check before we land and cut a new release.

includes/bootstrap.php
```diff
@@ -34,4 +38,9 @@
 // Load REST API init class for plugin bootstrap.
 if ( ! class_exists( 'WP_REST_Abilities_Init' ) ) {
 	require_once __DIR__ . '/rest-api/class-wp-rest-abilities-init.php';
+
+	// Initialize REST API routes when WordPress is available.
+	if ( function_exists( 'add_action' ) ) {
```
justlevine, line 43: Do we need the function check if we've already proven `ABSPATH`, or do we need `ABSPATH` if we're anyway checking for `add_action()` before running any WP code?

tests/bootstrap.php
```diff
@@ -34,7 +34,8 @@
 tests_add_filter(
 	'muplugins_loaded',
 	static function (): void {
-		require_once dirname( __DIR__ ) . '/abilities-api.php';
+		// Require ( to bypass require_once ).
+		require dirname( __DIR__ ) . '/includes/bootstrap.php';
```
justlevine, line 38: (I'd feel better about skipping `abilities-api.php` if understood the purpose of all this, but since it's being left out of core, it doesn't actually matter)

justlevine (approved):
Nothing here is breaking or going into core. Worst case in the ~7 weeks before merge someone reports a bug caused by a race condition on action loading, and we tell them to either to instead install it as a plugin or move their `require_once vendor/autoload.php` somewhere else 😅

A couple smaller and equally nonimportant comments on the diff, but otherwise LGTM 🚀

--- 

### That said,

I think _I_ still need a sanity check when it comes to this whole Composer use case (cc @Jameswlepage ). **nonblocking** feedback with that in mind:

1. Is initialization something that is wanted/expected when _autoloading_ a **library**? (I would usually assume no.)
2. If it is (or the real use case is as a plugin), then do we even need an `includes/boostrap.php` or can we just use `abilities-api.php` as the entrypoint?

(Only tangential to this PR: is _autoloading_ something that is wanted/expected or is a downstream `require_once vendor/wordpress/abilities-api/{whatever-entrypoint}.php` a better pattern for this particular repo's goals?)

includes/bootstrap.php
gziolo: Good point.

budzanowski: `ABSPATH` prevents processing of the file outside WP env - this is generally useful and in our context stops processing of the file during PHPUnit start procedure, when WP is not yet ready. We want to control the moment this gets loaded.

`add_action` tests could be considered redundant if `ABSPATH` is present. Though in WP startup procedure `ABSPATH` is defined earlier than `add_action` becomes available. Potentially we can have `ABSPATH` but no `add_action`. This just prevents a very "wrong" usage patter from crashing the site.

tests/bootstrap.php
budzanowski: This is a test bootstrap. The purpose of this is to circumvent the quirkiness of PHPUnit which automatically loads autoloader, at the very beginning, and thus is triggering the loading of the library. Since it is doing this before WordPress environment has been loaded nothing gets actually loaded ( ABSPATH test prevents it ). Though PHP remembers the `require_once` calls. If we would load again `abilities-api.php` nothing would happen, because `require_once` for `bootstrap.php` file has been already called. Thus I have made a decision to call it directly for tests once again ( using `require` to force the load ) - after the WP environment has been loaded. This allows proper loading of all library elements and registering of the filter.

For normal operations from inside WP or a plugin this alterations would not be necessary - this is just for the unit tests and avoids tampering with `require_once` directives inside the production code which I wanted to avoid.

gziolo: This one is tricky as autoloader requires the same file, but it doesn't run any logic because of `ABSPATH` check, so it looks that all is correct.

includes/bootstrap.php
gziolo: It's also fine to be extra careful. Let's land this as is, as it won't be included in WP core in the same fashion.

gziolo:
I enabled auto-merge. It's interesting to see Claude on the list of commit authors 😄

budzanowski:
@justlevine Thank you very much for the review.

> Is initialization something that is wanted/expected when autoloading a library? (I would usually assume no.)\

I agree in general about the initialization. But I think the pattern here is just to make all of the global functions available.


> If it is (or the real use case is as a plugin), then do we even need an includes/boostrap.php or can we just use abilities-api.php as the entrypoint?

This is mostly a semantic separation. `abilities-api.php` is plugin entry point and `boostrap.php` is code logic entry point and it is used by `composer` autoloader to enable the functionality. They could be merged, but I see that we may in future require some logic in the plugin part that should not be part of the package backend.

tests/bootstrap.php
justlevine: (@budzanowski yes I understood the what and the how, it's the premise that we should be "triggering the loading of the library" - or more precisely calling that autoloading is now conflated with instantiating the REST endpoint - but still have a separate bootstrap file that confuses me. The architectural implication is that there would be other functionality in `./abilities-api.php` now or in the future that we'd no longer be able to test... but again since it's not true IRL it doesn't matter)

justlevine:
> They could be merged, but I see that we may in future require some logic in the plugin part that should not be part of the package backend.

I'm good with this separation seperation of concerns, thanks both for the contribution and your time with the explanation!

tests/bootstrap.php
budzanowski: @justlevine I agree this could makes things tricky down the road, but I did not wanted to spend too much time anticipating what may be required in the future without actually knowing how this will be used. I bet we will be able to test `./abilities-api.php` separately if needed so hopefully this is not limiting.

PR #57, opened 2025-09-04
Updated Inline Documentation Order

sandipr942:
Changed Documentation Order in `includes/abilities-api.php` (Added `@since` tag before `@see` tag according to [PHP Documentation Standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/)


gziolo (approved):
Thank you for contribution 👍

PR #59, opened 2025-09-05
Refactor the usage of `$properties` param to `$args` to align with WordPress usage

gziolo:
Follow up for:
- https://github.com/WordPress/abilities-api/pull/54

More details in this https://github.com/WordPress/abilities-api/pull/54#discussion_r2316593195. The most relevant part from @justlevine:

> My takeaway from this conversation is that we're good not caring about that last point (between overloading and the filter if someone really, realy thinks shimming a DTO or VO into this api is a good idea, they have ways), so if y'all don't think it's outweighed by the first two (I don't), I'll use `prepare_*`( `array<string,mixed>` ): `array<valid-shape>` and just `throw`.

This PR refactors the usage of `$properties` to `$args` to reflect the conversation. As part of the refactoring, the definition for ability registration was also updated to remove the optionality of `$args` because to work correctly, there needs to be at least `ability_class` passed, or more frequently, three fields: `label`, `description`, and `execute_callback`.

Entire codebase was slightly refactored to follow the convention agreed upon.

### DIFF
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -98,6 +98,7 @@ final class WP_Abilities_Registry {
 		unset( $args['ability_class'] );
 
 		try {
+			// WP_Ability::prepare_properties() will throw an exception if the properties are invalid.
 			$ability = new $ability_class( $name, $args );
 		} catch ( \InvalidArgumentException $e ) {
 			_doing_it_wrong(
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -43,12 +43,12 @@ final class WP_Abilities_Registry {
 	 *
 	 * @see wp_register_ability()
 	 *
-	 * @param string              $name       The name of the ability. The name must be a string containing a namespace
-	 *                                        prefix, i.e. `my-plugin/my-ability`. It can only contain lowercase
-	 *                                        alphanumeric characters, dashes and the forward slash.
-	 * @param array<string,mixed> $properties An associative array of properties for the ability. This should include
-	 *                                        `label`, `description`, `input_schema`, `output_schema`,
-	 *                                        `execute_callback`, `permission_callback`, `meta`, and ability_class.
+	 * @param string              $name The name of the ability. The name must be a string containing a namespace
+	 *                                  prefix, i.e. `my-plugin/my-ability`. It can only contain lowercase
+	 *                                  alphanumeric characters, dashes and the forward slash.
+	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should include
+	 *                                  `label`, `description`, `input_schema`, `output_schema`,
+	 *                                  `execute_callback`, `permission_callback`, `meta`, and ability_class.
 	 * @return ?\WP_Ability The registered ability instance on success, null on failure.
 	 *
 	 * @phpstan-param array{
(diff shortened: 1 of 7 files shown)


gziolo:
@justlevine, I cherry-picked your branch and continued refactoring. I'm inclined to go with `prepare_properties( array $args )` because the result of that method is used to set class properties. However, it's a minor thing that I wanted to further discuss. Otherwise, I tried to go with `args` in all other places instead of properties, including unit tests.

justlevine:
> I'm inclined to go with `prepare_properties( array $args )` because the result of that method is used to set class properties.

Semantically, I'm good either way (do you "prepare the args" _to be used_ as class properties or do you "prepare properties" _so they can be used_?)

Only reason I went with `prepare_args()` is because I'm assuming we'd want the eventual _filter_ to use args ( e.g. `return return apply_filters( 'ability_args', $args, $this->name )`, or even `ability_prepare_args`, or whatever naming semantics ), to [keep the developer DX and terminology intuitive](https://github.com/WordPress/ai/issues/40).

But filters don't _need_ to match the semantics for direct class overloading, and they also have a much easier core deprecation path if we want to change the name compared to a `protected function *`, so 🤷

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -86,23 +86,19 @@ public function register( string $name, array $properties = array() ): ?WP_Abili
...
 			);
 			return null;
 		}
-		$ability_class = $properties['ability_class'] ?? WP_Ability::class;
-		unset( $properties['ability_class'] );
+		$ability_class = $args['ability_class'] ?? WP_Ability::class;
+		unset( $args['ability_class'] );
 
 		try {
-			// WP_Ability::validate_properties() will throw an exception if the properties are invalid.
```
gziolo, line 101: It might be a code agent that removed the comment 😅

tests/unit/abilities-api/wpRegisterAbility.php
```diff
@@ -315,13 +315,13 @@ public function test_permission_callback_receives_input(): void {
 		do_action( 'abilities_api_init' );
 
 		$received_input                                       = null;
```
gziolo: CI on WP core will complain, need to fix spacing here and most likely in a few other places.

gziolo:
> Only reason I went with prepare_args() is because I'm assuming we'd want the eventual filter to use args ( e.g. return return apply_filters( 'ability_args', $args, $this->name ), or even ability_prepare_args, or whatever naming semantics ), to https://github.com/WordPress/ai/issues/40.

Right, both names are fine here in that context. The filter name doesn't need to follow the method name, so we can be flexible.

gziolo:
I verified the changes in the WP core env (https://github.com/WordPress/wordpress-develop/pull/9410), and applied some small tweaks to ensure CI is green there, too.

I also set the version for the plugin and Composer package to `v0.2.0` to account for the fact that we slightly modified the public API params – `$args` are no longer optional, which better reflects the real usage.

gziolo:
@justlevine or @emdashcodes. It looks like I can no longer merge without approval from someone else. I would appreciate a final check from one of you.

emdashcodes (approved):
Just had a chance to look. 👍

justlevine:
> I also set the version for the plugin and Composer package to v0.2.0

In the future, would we be able to use `@since n.e.x.t.` and save pre-release versioning for a separate PR / release step? 
1. It prevents us from missing other things that need to be versioned (e.g. the constant, a docs ref too iirc)
2. It prevents `trunk` from lying to us or our LLMs. There's no v0.2.0 until we cut it.

gziolo:
> In the future, would we be able to use @since n.e.x.t. and save pre-release versioning for a separate PR / release step?

Is it something that could be automated with a GitHub action triggered when a new release is created through GitHub UI?

justlevine:
Yes it is, though if there isn't already FOSSable prior art somewhere, it might not make sense to waste time on a solution for _this repo alone_ over the next six weeks.

PR #60, opened 2025-09-05
Add client-side package for Abilities API

### LINKED ISSUE #41: Implement client-side integration for abilities registered on the server
We are close to wrapping up the [Abilities API 1.0 Milestone](https://github.com/WordPress/abilities-api/milestone/1), covering all server-side aspects. Now, it's time to start working on implementing a client-side registry for the Abilities API to streamline usage with JavaScript.

We can use for inspiration the existing implementation in the WP Feature API plugin:

https://github.com/Automattic/wp-feature-api/tree/trunk/packages/client

## Planned Tasks

I propose we start with developing a JavaScript package on top of the server-side registry, so it's easy to:
 
- list all available abilities
- find a specific ability by name
- execute a selected ability on the server

Internally, this would use the existing REST API endpoints added in https://github.com/WordPress/abilities-api/pull/6.

See the prototyped version:

https://github.com/Automattic/wp-feature-api/blob/trunk/packages/client/src/api.ts

## Future

1. Implement a way to register abilities that work only on the client.
2. It's also worth considering a solution that allows extending the ability registered on the server with an alternative implementation optimized for client usage. A good example is operations on data entities (such as posts or pages) through the `@wordpress/core-data` package, which provides an abstraction layer on top of REST API operations, offering optimistic updates, data caching, and tight integration with the editor's UI.

emdashcodes:
Closes https://github.com/WordPress/abilities-api/issues/41

This PR introduces a client library that provides access to abilities registered on the server, making it easy to discover and execute WordPress abilities from JavaScript code.

- `getAbilities()` - Fetches all available abilities
- `getAbility(name)` - Retrieves a specific ability by name
- `executeAbility(name, input)` - Executes abilities

In addition to the client API, this also includes automatic script registration when used as a plugin and a helper function for Composer-based installations.  The API is exposed in the `wp.abilities` namespace when enqueued.

It consolidates other aspects of client-side registry work for Abilities API:
- WordPress/abilities-api#69
- WordPress/abilities-api#70 

### Usage Example

```javascript
const {getAbilities, executeAbility } = wp.abilities;

// Getall abilities

// Execute an ability
const result = await executeAbility('my-plugin/create-post', {
    title: 'New Post',
    content: 'Post content'
});
```

### Testing

```bash
cd packages/client

# Install dependencies
npm install

# Build the package
npm run build
# or npm run build:client from root
```

I recommend registering at least 1 tool and 1 resource, here are 2 dummy ones that I will use in the examples below: https://gist.github.com/emdashcodes/a2f2d2fe440d52d9736fac4ab535cef8

Open the developer console and paste the following:

`await wp.abilities.getAbilities()`

You should see both abilities (and any others you registered) listed.

Paste in the following:

```
await wp.abilities.executeAbility('demo/get-pages', {
    limit: 5,
    orderby: 'title',
   order: 'ASC'
});
```

You should see a list of pages returned.

Paste in:

```
await wp.abilities.getAbility('demo/get-pages');
```

See that info for the `demo/get-pages` ability is called.

Finally, paste in the following:

```
await wp.abilities.executeAbility('demo/create-post', {
    title: 'Test Post from Abilities API',
    content: '<p>This post was created using the WordPress Abilities API!</p>',
    excerpt: 'A test post created via the browser console',
    status: 'draft'
});
```

See that a demo draft post has been created.


includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
```diff
@@ -128,13 +128,13 @@ public function get_items( $request ) {
 		if ( $page > 1 ) {
 			$prev_page = $page - 1;
 			$prev_link = add_query_arg( 'page', $prev_page, $base );
-			$response->add_link( 'prev', $prev_link );
+			$response->link_header( 'prev', $prev_link );
```
emdashcodes, line 131: We use link_header() instead of add_link() to keep the response as a pure array.

Before, When using add_link(), it would add "_links" to the response body, converting it from an array / to a JSON object like {"0": {...}, "_links": {...}}.

This would be a breaking fix, but I also don't know anyone if anyone would have been depending on pagination yet since you need a large number. This can be tested by setting it to something lower like `1` and testing the listAbilities method.

gziolo: I see that calling `->link_header()` is a well-established pattern in WP Core for `prev`/`next`.

includes/bootstrap.php
```diff
@@ -44,3 +44,8 @@
 		add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
 	}
 }
+
+// Load client registration helper for Composer users.
+if ( ! function_exists( 'wp_abilities_register_client_assets' ) ) {
+	require_once __DIR__ . '/abilities-api/client-registration.php';
```
gziolo: I see this would be available through a Composer package, but the developer would have to call it explicitly to initialize.

Aside: in WordPress core, this will be handled through the default scripts handler rather than through a standalone. However, similar sitiuation applies to `WP_REST_Abilities_Init` so this should be perfectly fine.

emdashcodes: One thing to note for the assets, I think @budzanowski also noticed that we need to add some kind of build step for the JS for this to work without an extra step on the part of developers shipping this in composer.

And yes, this is definitely meant to be temporary!

gziolo: Interesting with the bundled JS in the Composer package. That would definitely be straightforward for plug-in in other places.

includes/abilities-api/client-registration.php
```diff
@@ -0,0 +1,72 @@
...
+		// Running as a plugin
+		$base_path = WP_ABILITIES_API_DIR;
+		$base_url  = plugins_url( '', dirname( __DIR__, 2 ) . '/abilities-api.php' );
+	} else {
+		// Running as a Composer package
+		$base_path = dirname( __DIR__, 2 );
+
+		// For Composer, we need to determine the URL based on the installation location
+		$plugin_dir = WP_PLUGIN_DIR;
+		if ( strpos( $base_path, $plugin_dir ) === 0 ) {
```
gziolo: For Windows, it might require some path normalization. 

See an example function to compare how paths are handled:

https://github.com/WordPress/wordpress-develop/blob/cc9f9664c370146c23ce8022bf7f72a882e18144/src/wp-includes/blocks.php#L90

packages/client/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
```
justlevine: Should we be using the `compilerOptions` from Gutenberg? Or at least aligning where it makes sense?

E.g. there's no reason to have a lower target/module

https://github.com/WordPress/gutenberg/blob/trunk/tsconfig.base.json

packages/client/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "wpScript": true,
+  "types": "build-types",
+  "files": [
+    "build",
+    "build-module",
+    "build-types",
+    "src",
+    "*.md"
+  ],
+  "scripts": {
```
justlevine: Is there a a reason to have these all `scripts` here instead of as part of our root `package.json`?

packages/client/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@typescript-eslint/eslint-plugin": "^8.42.0",
+    "@typescript-eslint/parser": "^8.42.0",
+    "@wordpress/eslint-plugin": "^22.16.0",
+    "@wordpress/prettier-config": "^4.30.0",
+    "@wordpress/scripts": "^30.23.0",
+    "eslint": "^8.57.0",
+    "eslint-plugin-import": "^2.32.0",
+    "eslint-plugin-jest": "^29.0.1",
+    "prettier": "npm:wp-prettier@3.0.3",
+    "typescript": "^5.9.2"
```
justlevine: Can we move all these into the root `package.json`? I see no reason to juggle versions of our dev-dependencies.

packages/client/.prettierrc.js
justlevine, line 1: No reason to keep this around if we format from root.

packages/client/.eslintrc.cjs
justlevine, line 1: Is there a reason not to manage this root-level?

packages/client/.eslintrc.cjs
```diff
@@ -0,0 +1,57 @@
...
+				project: './tsconfig.json',
+			},
+		},
+	},
+	env: {
+		browser: true,
+		es6: true,
+		node: true,
+	},
+	rules: {
```
justlevine: There's probably other rules we want to `error` for a la Gutenberg: https://github.com/WordPress/gutenberg/blob/60526a2dae790f202571f3a581c5c3e67dee4e3f/.eslintrc.js#L159

packages/client/package.json
justlevine: Also arguably the `"workspaces": ["packages/*"]` should probably be added to the root `package.json` too.

includes/bootstrap.php
justlevine: > However, similar situation applies to WP_REST_Abilities_Init so this should be perfectly fine.

IIRC https://github.com/WordPress/abilities-api/pull/55 changed this, as you can see right above this code block `WP_REST_Abilities_Init` is always hooked even for Composer libraries, no manual additional step required.

> we need to add some kind of build step for the JS

Really straightforward to do in the current release step, and will be even more straightforward if we handle packages as an actual NPM workspace instead of nesting everything inside packages/client.

Either way once a decision is made on that, I'm happy to open a PR against this for the CI + docs change.

includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
justlevine: (If for some reason the rest of this PR doesn't make it into 0.2.0, then at minimum we should backport this so that breaking change makes it in. )

includes/abilities-api/client-registration.php
justlevine, line 1: Don't like that we're muddying up the global namespace unnecessarily, or that this file is nested inside `includes/abilities-api`. Also it's breaks the current repo CS.

IMO, **either** this should be moved up into `includes/*` **or** the plugin-specific needs should be sanitized out of the WordPress-core needs into a `class-assets-init.php` or similar (following the pattern established in `class-wp-rest-abilities-init.php`)

gziolo: I might leave a similar feedback in another PR. This will stay local, similar to `WP_REST_Abilities_Init`. So maybe in both cases, the name shouldn't be prefixed with `wp`, but in this PR it's fine to change the newly proposed name.

packages/client/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+Client library for the WordPress Abilities API, providing a standardized way to discover and execute WordPress capabilities.
+
+## Installation
+
+The client is available in two ways:
+
+### 1. As an npm Package (Coming Soon)
+
+The npm package `@wordpress/abilities` is planned for future publication. Once published, you'll be able to install it via:
```
gziolo: That's something that we might keep local until the code lands in Gutenberg and the usual npm publishing takes over the ownership. In particular, taking into account that the approach with Composer would offer early access to the code, too.

packages/client/src/api.ts
```diff
@@ -0,0 +1,84 @@
...
+ * The method (GET or POST) is determined by the ability's type metadata.
+ *
+ * @param name  The ability name.
+ * @param input Optional input parameters for the ability.
+ * @return Promise resolving to the ability execution result.
+ * @throws Error if the ability is not found.
+ */
+export async function executeAbility(
+	name: string,
+	input: AbilityInput = {}
```
gziolo: With the recent changes, the `input` is no longer an object:

https://github.com/WordPress/abilities-api/blob/59d8ec6869c54ab26ef897efd36f1bb7edd348c5/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php#L208

Not sure how this should be reflected with TypeScript tnow.

packages/client/src/store/reducer.ts
```diff
@@ -0,0 +1,48 @@
...
+const DEFAULT_STATE: Record< string, Ability > = {};
+
+/**
+ * Reducer managing the abilities by ID.
+ *
+ * @param state  Current state.
+ * @param action Dispatched action.
+ * @return New state.
+ */
+function abilitiesById(
```
gziolo: Nit: abilities are stored by name, so that might be better to reflect it in the name of the reducer: 
```suggestion
function abilitiesByName(
```

packages/client/src/store/resolvers.ts
```diff
@@ -0,0 +1,87 @@
...
+		}
+
+		const select = registry.select( coreStore );
+		const totalPages = select.getEntityRecordsTotalPages(
+			ENTITY_KIND,
+			ENTITY_NAME,
+			{ per_page: PER_PAGE }
+		);
+
+		if ( totalPages === 1 ) {
```
gziolo: See my comment https://github.com/WordPress/abilities-api/pull/69#discussion_r2343349207 regarding using `-1` to automate paginated support.

packages/client/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Usage
+
+```javascript
+// In your WordPress plugin or theme JavaScript
+const { listAbilities, getAbility, executeAbility } = wp.abilities;
+// or import { listAbilities, getAbility, executeAbility } from '@wordpress/abilities'; depending on your setup
+
+// List all abilities
+const abilities = await listAbilities();
```
gziolo: In https://github.com/WordPress/abilities-api/pull/70#discussion_r2343608896 I suggested we rename the method to `getAbilities()`. I see that it's also what is used in the data store.

packages/client/README.md
```diff
@@ -0,0 +1,170 @@
...
+// Execute an ability
+const result = await executeAbility( 'my-plugin/my-ability', {
+    param1: 'value1',
+    param2: 'value2'
+} );
+```
+
+### Using with React and WordPress Data
+
+The client includes a data store that integrates with `@wordpress/data` for use in React components:
```
gziolo, line 42: This is important to promote using the selectors from the data store directly in React components, ensuring they re-render when something changes. The store isn't going to change that often in practice, but it still is essential to account for extensibility in case someone wants to register/unregister abilities on the fly.

packages/client/src/api.ts
```diff
@@ -0,0 +1,84 @@
...
+		Object.entries( input ).forEach( ( [ key, value ] ) => {
+			params.append( `input[${ key }]`, String( value ) );
+		} );
+		path = `${ path }?${ params.toString() }`;
+	} else if ( method === 'POST' ) {
+		options.data = { input };
+	}
+
+	try {
+		return await apiFetch< AbilityOutput >( {
```
gziolo, line 199: @mamaduka or @jsnajdr, do we have actions in WordPress data stores that are thunks and alone don't update the state? I'm mostly interested in understanding whether the proposed implementation, expressed also as an action, would bring any benefits or only confusion.

packages/client/src/store/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+dispatch( coreStore ).addEntities( [
+	{
+		name: ENTITY_NAME,
+		kind: ENTITY_KIND,
+		key: 'name',
+		baseURL: '/wp/v2/abilities',
+		baseURLParams: { context: 'edit' },
+		plural: 'abilities',
+		label: 'Abilities',
```
gziolo: This would have to be wrapped with a `@wordpress/i18n` helper.

packages/client/src/api.ts
```diff
@@ -0,0 +1,84 @@
...
+import type { Ability, AbilityInput, AbilityOutput } from './types';
+
+/**
+ * List all available abilities.
+ *
+ * @return Promise resolving to array of abilities.
+ */
+export async function listAbilities(): Promise< Ability[] > {
+	return await resolveSelect( store ).getAbilities();
+}
```
Mamaduka, line 23: Maybe we should use `getAbilities` for the API name as well. It will make it easier to remember.

packages/client/src/store/resolvers.ts
```diff
@@ -0,0 +1,87 @@
...
+ */
+import { ENTITY_KIND, ENTITY_NAME } from './constants';
+import { receiveAbilities } from './actions';
+import type { Ability } from '../types';
+
+/**
+ * Resolver for getAbilities selector.
+ * Fetches all abilities from the server with pagination.
+ */
+export function getAbilities() {
```
Mamaduka, line 16: Will there be a case when we want to invalidate the resolution for this resolver?

packages/client/src/api.ts
Mamaduka: I don't recall any similar actions from more recent packages. I think `trashPost` from `core/editor` doesn't update the store, but that's for different reasons, and we could consider that a legacy action.

P.S. It's much easier to introduce new APIs or actions later than to remove them 😄

packages/client/src/api.ts
```diff
@@ -0,0 +1,84 @@
...
+		method,
+	};
+
+	if ( method === 'GET' && Object.keys( input ).length > 0 ) {
+		// e.g., input[format]=iso&input[timezone]=UTC
+		const params = new URLSearchParams();
+		Object.entries( input ).forEach( ( [ key, value ] ) => {
+			params.append( `input[${ key }]`, String( value ) );
+		} );
+		path = `${ path }?${ params.toString() }`;
```
jsnajdr: The query string formatting can be done with `buildQueryString` or `addQueryArgs` from `@wordpress/url`. No need to implement our own version.

packages/client/src/store/resolvers.ts
```diff
@@ -0,0 +1,87 @@
...
+			}
+
+			const abilities = await resolveSelect.getEntityRecords(
+				ENTITY_KIND,
+				ENTITY_NAME,
+				{ per_page: PER_PAGE, page }
+			);
+
+			const combined = [ ...accumulated, ...( abilities || [] ) ];
+			return fetchRemainingPages( page + 1, combined );
```
jsnajdr: Within `async` function it's unusual to use recursion like this. It can lead to a deep call stack because JS engines don't optimize tail recursion. Can be easily rewritten into a `while` loop.

And as @gziolo says, this entire "fetch all pages" is already implemented in `apiFetch` as a middleware, triggered by `page: -1`. Its probably redundant here.

packages/client/src/api.ts
jsnajdr: I think it's a good idea to implement `executeAbility` as an action, for consistency reasons. Most of the API is done in the data store and there are thin wrappers around it. The API can be accessed both ways, on the store and with regular functions. `executeAbility` would be an unpleasant exception.

But I would question the need for the `abilities` data store. Why is it needed at all? The query is performed in the `core` store anyway and the data are cached there. Why does the `abilities` store cache it for the second time? The wrappers in this `api.ts` module could work directly on the `coreStore`.

Mamaduka: > But I would question the need for the abilities data store. Why is it needed at all?

That's a good point. @youknowriad also listed adding stores to new packages as a mistake we should avoid -  Ref (very last paragraph) - https://make.wordpress.org/core/2024/09/12/gutenberg-development-practices-and-common-pitfalls/.

gziolo: > But I would question the need for the abilities data store. Why is it needed at all? The query is performed in the core store anyway and the data are cached there. 

There is need for the store that keeps registered abilities from the client as proposed in #69. However, it is a really good point regarding all these abilities that live only on the server. I'm not sure what's the best way to merge both types on the client, so I would appreciate some ideas.

It's also worth noting, that I anticipate that we might have hybrid abilities where the abilitie defined on the server, can still have a different execute callback defined on the client to better integrate with the context where it is executed. To give a quick example, imagine there is an ability that edits a post (or any other post type). On the server, it could be handled with `edit_post()` function, while on the client the same operation would be better suited to be handled with the entity `editEntityRecord( 'postType', 'post', postId, { ... } )` to benefit from the deep integrations with the editor.

includes/bootstrap.php
emdashcodes: I have updated this in https://github.com/WordPress/abilities-api/pull/60/commits/dce8f8cda07c09ba3b4fe64a5d1fdc4bd3f43a2c to follow #55. This should be always hooked now as well.

includes/abilities-api/client-registration.php
emdashcodes: In https://github.com/WordPress/abilities-api/pull/60/commits/dce8f8cda07c09ba3b4fe64a5d1fdc4bd3f43a2c, I have moved these from the global namespace and created a `WP_Abilities_Assets_Init` class.

includes/abilities-api/client-registration.php
emdashcodes: Added some normalization: https://github.com/WordPress/abilities-api/pull/60/commits/7423d9d4a1ce9d67a97133a652d6e3a7dff1611f

packages/client/tsconfig.json
emdashcodes: I've aligned/pulled in the ones that make sense here. Unfortunately, I don't think there is a way we can pull in the exact config until this lands as a part of Gutenberg some point in the future. https://github.com/WordPress/abilities-api/pull/60/commits/6bb68dd96e4b9fcb605a30a6542d9cd2baf2d8cc

packages/client/package.json
emdashcodes: Moved the main scripts to the root, added the workspace to the root. https://github.com/WordPress/abilities-api/pull/60/commits/6bb68dd96e4b9fcb605a30a6542d9cd2baf2d8cc

packages/client/package.json
emdashcodes: Done in https://github.com/WordPress/abilities-api/pull/60/commits/6bb68dd96e4b9fcb605a30a6542d9cd2baf2d8cc

packages/client/.prettierrc.js
emdashcodes: Cleaned up in https://github.com/WordPress/abilities-api/pull/60/commits/6bb68dd96e4b9fcb605a30a6542d9cd2baf2d8cc

packages/client/.eslintrc.cjs
emdashcodes: Cleaned up in https://github.com/WordPress/abilities-api/pull/60/commits/6bb68dd96e4b9fcb605a30a6542d9cd2baf2d8cc

packages/client/.eslintrc.cjs
emdashcodes: I have pulled over applicable rules in https://github.com/WordPress/abilities-api/pull/60/commits/6bb68dd96e4b9fcb605a30a6542d9cd2baf2d8cc.

packages/client/README.md
emdashcodes: I have removed mention of the npm package for now. Doc updates in https://github.com/WordPress/abilities-api/pull/60/commits/ebbabcbb9f935be7023fccb90aa9551a15540fdf

packages/client/README.md
emdashcodes: Renamed to `getAbilities` in https://github.com/WordPress/abilities-api/pull/60/commits/43ed397712f1b9028c3a29f67e51d4b507801b46.

packages/client/src/api.ts
emdashcodes: Agreed, renamed to getAbilities in https://github.com/WordPress/abilities-api/pull/60/commits/43ed397712f1b9028c3a29f67e51d4b507801b46

packages/client/src/store/reducer.ts
emdashcodes: Renamed in https://github.com/WordPress/abilities-api/pull/60/commits/43ed397712f1b9028c3a29f67e51d4b507801b46

packages/client/src/api.ts
emdashcodes: Updated in https://github.com/WordPress/abilities-api/pull/60/commits/fabe5af2fbaeccb8ea933764720b5c127064ba97 so that it can accept other types now: primitive, array, object, or null.

packages/client/src/store/resolvers.ts
emdashcodes: Fixed in https://github.com/WordPress/abilities-api/pull/60/commits/4bce923e1671164220d2e860c21c83aac106b4e4.

packages/client/src/store/resolvers.ts
emdashcodes: Fixed in https://github.com/WordPress/abilities-api/pull/60/commits/4bce923e1671164220d2e860c21c83aac106b4e4.

packages/client/src/api.ts
emdashcodes: Good call out. Updated to use `addQueryArgs` in https://github.com/WordPress/abilities-api/pull/60/commits/02a26139c18aa45bc6e3edc61d7503b9e1c822c3

packages/client/src/store/resolvers.ts
emdashcodes: I'm not sure. The registered server abilities should not change super often.

I guess when plugins enable and disable, might be the time where these might need updated?

packages/client/src/store/index.ts
emdashcodes: Wrapped in https://github.com/WordPress/abilities-api/pull/60/commits/5aacbf6e4e8e38caf27132f981108ccfa6215657

packages/client/src/api.ts
emdashcodes: As @gziolo noted, this store is also used for client-only abilities and in the future, potentially hybrid abilities. 

I think a store for this makes sense (keeping in mind client-only abilities from #69) -- but I'm also open to other suggestions.

emdashcodes:
Thank you all for all the greatPR feedback here! I believe I have addressed everything and I've updated the PR instructions as well.

I'll continue addressing feedback left on PR WordPress/abilities-api#69 and WordPress/abilities-api#70.

packages/client/package.json
```diff
@@ -0,0 +1,54 @@
+{
```
gziolo, line 1: I belive that with npm workspaces configured, the lock file is no longer in use and it can be safely deleted.

includes/bootstrap.php
```diff
@@ -44,3 +44,14 @@
 		add_action( 'rest_api_init', array( 'WP_REST_Abilities_Init', 'register_routes' ) );
 	}
 }
+
+// Load assets init class for plugin bootstrap.
+if ( ! class_exists( 'WP_Abilities_Assets_Init' ) ) {
```
gziolo, line 49: Do we have in place the build step for JS assets triggered during a new GitHub release? That would be essential for the Composer package so folks could consume it as is. 

In fact, that would also be necessary for the generated zip file with the plugin that gets triggered. @justlevine might have more insights on how it works in detail.

The good news is that there a check in place which would prevent errors if the file is not found.

gziolo (commented):
It looks like all my feedback was addressed. I left some minor additional feedback. I will wait for responses from other folks that left their review comments.

includes/bootstrap.php
justlevine: @gziolo Havn't had time to review the changes here and in #70 in depth but at a quick glance (so I don't become a blocker)

1. Add the build step to `./github/release.yml` between `npm ci` and `npm run plugin-zip`
2. Add the build _dir_ to the `package.json:files` property

That's all that's needed to make it available for the distribution _plugin_.

---

For the Composer "library", it might be enough to make the built assets also available for `prefer-dist` , but I'd need to confirm.

However, for sure it would **not** make the built assets available if somebody `composer require`d a specific tag/branch.  This is symptom of us calling this a "Composer library" even though [it isnt](https://github.com/WordPress/abilities-api/pull/66#issuecomment-3302670078)  🤷

Workarounds I'm familiar with are:

1. commit the `built` assets dir to the repo (simplest, but trades user friction for contributor merge conflicts)
2. use a 3party composer dependency that will call `nvm install && nvm use && npm install && npm run build` (I don't think we have WordPress/* prior art but there's several popular libs)
3. installation docs telling Composer users to add `"post-install-cwd": nvm install && nvm use && npm install && npm run build"` to their `composer.json` (which is technically a smaller change than what's required while #66 is `wontfix`, but now it means _all_ composer users need make manual config changes to their project json's).
4. releasing on w.org, and then `composer require wpackagist-plugin/abilities-api` (because if you want _integration with wordpress_ you want a plugin, not a library.)
    - If we go this route, I'd recommend moving the hooks/initialization out of the bootstrap and into the root plugin folder. This would then allow people who actually want to use this as a Composer library to do so via `composer require wordpress/abilities-api`, and the people who want to call it a "library" but actually want a Composer-based plugin dependency can use `composer require wpackagist-plugin/abilities-api`.
    
Don't know what to specifically recommend here since our decisions have been inconsistent on the topic, but that last one should let everyone have their cake and eat it too (afaik from what's been communicated publicly)

cc: @felixarntz and @Jameswlepage , and I'll reraise the topic on slack.

packages/client/package.json
gziolo: I see it was done here https://github.com/WordPress/abilities-api/commit/2828f26a371d0b13e3a95bcc70b697ca6227db0b as part of #70.

includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
emdashcodes: Created https://github.com/WordPress/abilities-api/pull/81 just to make sure we get this in and don't cause issues for ourselves.

gziolo:
I merged WordPress/abilities-api#70 (tests and CI integration) into WordPress/abilities-api#69 (store for abilities registered on the front end), and then WordPress/abilities-api#69 into this branch. This way, we can test and review everything integrated.

packages/client/src/store/actions.ts
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @param  ability The ability to register.
+ * @return Action object or function.
+ * @throws {Error} If validation fails.
+ */
+export function registerAbility(ability: Ability) {
+	// @ts-expect-error - registry types are not yet available
+	return ({ select, dispatch }) => {
+		if (!ability.name) {
+			throw new Error(__('Ability name is required'));
```
t-hamano: When an exception is thrown in JS, neither core nor Gutenberg localizes the error message. It may be better to remove the localization function from all exception throw messages.

packages/client/package.json
```diff
@@ -0,0 +1,65 @@
...
+		"build",
+		"build-module",
+		"build-types",
+		"src",
+		"*.md"
+	],
+	"scripts": {
+		"build": "wp-scripts build && tsc --emitDeclarationOnly --outDir build-types",
+		"dev": "wp-scripts build --watch",
+		"clean": "rm -rf build build-module build-types tsconfig.tsbuildinfo",
```
t-hamano: The `rm` command is a Unix-like command and will not work on Windows. We might want to use `rimraf` instead.

Example: https://github.com/WordPress/gutenberg/blob/23fadb3b0a1079c9de61ca93559174ddde97ed81/package.json#L190

.github/workflows/test.yml
```diff
@@ -247,3 +247,93 @@ jobs:
...
+  jslint:
+    name: Run JavaScript code quality checks
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
justlevine: Lets use the v5 pin from the other jobs.

.github/workflows/test.yml
```diff
@@ -247,3 +247,93 @@ jobs:
...
+        with:
+          show-progress: ${{ runner.debug == '1' && 'true' || 'false' }}
+          persist-credentials: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
```
justlevine: Pin at v5 (per the other jobs)

.github/workflows/test.yml
```diff
@@ -247,3 +247,93 @@ jobs:
...
+  jstest:
+    name: Run JavaScript tests
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
justlevine: 5.0 pin

.github/workflows/test.yml
```diff
@@ -247,3 +247,93 @@ jobs:
...
+        with:
+          show-progress: ${{ runner.debug == '1' && 'true' || 'false' }}
+          persist-credentials: false
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
```
justlevine: this too

.github/workflows/test.yml
```diff
@@ -247,3 +247,93 @@ jobs:
...
+          node-version-file: '.nvmrc'
+
+      - name: Install NPM dependencies
+        run: npm ci
+
+      - name: Run JavaScript tests with coverage
+        run: npm run test:client:coverage
+
+      - name: Upload JavaScript coverage to Codecov
+        uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574 # v5.4.0
```
justlevine: We're using 5.5.1 above https://github.com/WordPress/abilities-api/blob/89ef72415874e6286f3d30a7353a86f4db3e4109/.github/workflows/test.yml#L236

packages/client/README.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+### Store Selectors
+
+When using with `@wordpress/data`:
+
+- `getAbilities()` - Returns all abilities from the store
+- `getAbility(name)` - Returns a specific ability from the store
+
+## Development
```
justlevine: I'm a big fan of single sources of truth. Any chance we can just replace this entire section with a `For development and contributing guidelines, see [CONTRIBUTING.md](../CONTRIBUTING.md)` and have people run commands from the root of the repo?

(Similarly would remove the actual commands from the child package.json itself and align with WordPress/gutenberg https://github.com/WordPress/gutenberg/tree/trunk/packages )

packages/client/package.json
```diff
@@ -0,0 +1,65 @@
...
+	"wpScript": true,
+	"types": "build-types",
+	"files": [
+		"build",
+		"build-module",
+		"build-types",
+		"src",
+		"*.md"
+	],
+	"scripts": {
```
justlevine: lets move these to the root package.json also if possible, no reason to juggle them IMO + it aligns with gutenberg (either for prior art or eventual merge): https://github.com/WordPress/gutenberg/tree/trunk/packages

packages/client/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+	"compilerOptions": {
```
justlevine, line 3: Is there is reason why our lib/target/module/etc are deviating from Gutenberg: https://github.com/WordPress/gutenberg/blob/trunk/tsconfig.base.json

justlevine (commented):
Didn't have time to dive into the package itself, but here's some feedback on the boilerplate now that the different PRs have been merged together.

packages/client/src/store/actions.ts
emdashcodes: Thanks, I wasn't sure about these when I was translating the entity definition which was marked in a previous review. I've pulled the localization from the thrown errors. https://github.com/WordPress/abilities-api/pull/60/commits/cfb3731e4ece28ff1706ff86591305c8f40810f6

packages/client/package.json
emdashcodes: Thanks! Switched to rimraf in https://github.com/WordPress/abilities-api/pull/60/commits/b926c2b7cf9ad62415d818c11b778ffca022c5e1.

.github/workflows/test.yml
emdashcodes: Updated this and the other locations in https://github.com/WordPress/abilities-api/pull/60/commits/37e52b7be54e3ce0552694979055a138a77f4bd8.

packages/client/package.json
emdashcodes: All scripts have been moved to the root package. https://github.com/WordPress/abilities-api/pull/60/commits/8c0db8a2ad4cb1530354c5ec3c7e4bcb7d09d604

packages/client/README.md
emdashcodes: Makes sense. Cleaned up the documentation in https://github.com/WordPress/abilities-api/pull/60/commits/be1bfdd3676ec51f09b7bd9ee3e6ee4e815319f4 and moved the scripts in https://github.com/WordPress/abilities-api/commit/8c0db8a2ad4cb1530354c5ec3c7e4bcb7d09d604.

packages/client/tsconfig.json
emdashcodes: I swear I aligned these 😓  (as it was already called out in one of the other PRs), but thanks for catching! The tsconfig matches Gutenberg now https://github.com/WordPress/abilities-api/pull/60/commits/4c3085ccba739c43230c40016e5c2b2f7bc4098f

includes/bootstrap.php
emdashcodes: I have updated the .zip artifact building to properly include the files. https://github.com/WordPress/abilities-api/pull/60/commits/819f99d52811af503d906a1d82d26ede8deab765

After doing a bit of thinking for the composer workarounds (thanks for the details above @justlevine), I am leaning towards documenting the "post-install-cwd" step. I think this would be the most straight forward and not a huge ask of developers. It is essentially an opt-in to the client functionality. What do you think @gziolo @Jameswlepage?

(This should be considered temporary anyway if the client work is going to land with 6.9).

emdashcodes:
Thank you @t-hamano, @gziolo, and @justlevine. I've worked through the latest PR feedback above.

includes/bootstrap.php
gziolo: An alternative that we can also consider would be to open a dedicated branch every time we would like to release changes, and include the `build` folder with JavaScript files after manually running the `npm run build`. This way, the new version of the Composer package would contain all the necessary files. A little effort on our side, but a huge simplification for everyone else. In effect, `packages/client/build` folder would be only present in `release/x.y` branches, but not `trunk`.

gziolo:
I took this branch for a spin and identified a few issues that I tried to fix. First, I rebased the branch with the latest changes from `trunk`.

Next, I covered in https://github.com/WordPress/abilities-api/pull/60/commits/17dfa6227a74f2f7fef2e682266acd3fc27d507f:
- Updated most of the depenndecies to the latest version and removed these that are referenced in the code. I still would like to understand why `ajv-draft-04` is not referenced in the code, but instead, there is `ajv` in use, which has a larger size.
- Fixed security issues reported by `npm audit fix`. There are still 47 remaining warnings, but they can be only fixed upstream in Gutenberg.
- Executed `npm dedupe` to ensure npm packages are only installed in the top-level `node_modules` folder.

Next, I noticed some issues with the Prettier config when running `npm run lint:js`, which I resolved with https://github.com/WordPress/abilities-api/pull/60/commits/a42c8587dcbacc03caa25a07fee76de348c9d12c. This also required `npm run format`, which updated several files to adhere to WordPress coding style conventions.

package.json
```diff
@@ -14,17 +14,12 @@
...
-		"@wordpress/prettier-config": "^4.28.0",
-		"@testing-library/jest-dom": "^6.1.5",
-		"@testing-library/react": "^14.1.2",
-		"@types/jest": "^30.0.0",
-		"@wordpress/eslint-plugin": "^22.16.0",
-		"@wordpress/jest-console": "^8.30.0",
-		"eslint": "^8.57.0",
-		"eslint-plugin-import": "^2.32.0",
-		"eslint-plugin-jest": "^27.9.0",
+		"@types/jest": "^29.5.14",
```
gziolo: `@wordpress/scripts` depends on Jest v29 so I assumed we should align here. I'm still unsure whether we really need this dependency, as we really don't reference types from Jest in any place I could locate.

package.json
```diff
@@ -14,17 +14,12 @@
...
+		"@wordpress/env": "^10.31.0",
+		"@wordpress/eslint-plugin": "^22.17.0",
+		"@wordpress/jest-console": "^8.31.0",
```
gziolo: We most likely don't have to include `@wordpress/jest-console` as it isn't referenced in the code.

gziolo:
I also tested the following commands and they worked correctly with my latest changes applied:
- `npm run plugin-zip`
- `npm run build`
- `npm run dev`
- `npm run clean`
- `npm run format`
- `npm run lint:js`
- `npm run lint:js:fix`
- `npm run lint:all`
- `npm run typecheck`
- `npm run test:client`
- `npm run test:client:watch`
- `npm run test:client:coverage`
- `npm run test:all`
- `npm run wp-env`

The only task left for me that I have to perform later is ensuring that everything works correctly in the browser. In the long run, we might want to introduce some form of e2e testing with `wp-script test-playwright`. However, that can be added progressively as we work on adding abilities as discussed in https://github.com/WordPress/ai/issues/40.

includes/bootstrap.php
emdashcodes: I think that's a really good idea. We can probably automate that too and reduce some of the friction anyway.

emdashcodes:
> Updated most of the depenndecies to the latest version and removed these that are referenced in the code. I still would like to understand why ajv-draft-04 is not referenced in the code, but instead, there is ajv in use, which has a larger size.

Thanks, I believe I messed up the import somewhere along the way, but `ajv-draft-04` is what we actually want to be importing from the validation code.

via https://github.com/ajv-validator/ajv-draft-04

```
You need to install both ajv and this package (to allow upgrading ajv without upgrading this package):

npm i ajv ajv-draft-04
import Ajv from "ajv-draft-04"
const ajv = new Ajv()
```

I also cleaned up the two jest dependencies you noted.

gziolo:
I followed the testing instructions included in the description of this PR. This is the result:

Screenshot 2025-09-25 at 12 40 20
Screenshot 2025-09-25 at 12 40 47
Screenshot 2025-09-25 at 12 43 44
Screenshot 2025-09-25 at 12 42 11

I also performed an additional test to validate what happens when an invalid ability name is requested: 

Screenshot 2025-09-25 at 12 41 20
Screenshot 2025-09-25 at 12 47 54

---

Everything works as expected. My only feedback is that the REST API structure gets propagated to the list of abilities registered on the server. It's reflected in the presence of `_links` for individual abilities. It's rather something we should filter out from the ability objects.

gziolo:
I followed the instructions from https://github.com/WordPress/abilities-api/pull/69 to see how things change when there are also abilities registered on the client. Everything works correctly.

Screenshot 2025-09-25 at 12 57 20
Screenshot 2025-09-25 at 13 00 43

The only nitpick I have is that I still see `location: 'client'` in the ability object.

I can confirm that it isn't possible to register an ability using the same name twice. It is possible to unregister an ability (even when registered on the server), and register it again on the client.

---

The only remaining low-priority task would be aligning the shapes of abilities between the server-side representation and client-side representation. Namely, obsolete `_links` in the abilities coming from the server, and obsolete `location` for the abilities coming from the client.

A separate story is to coninue discussing whether some follow-up work would be required after landing this PR to optimize the abilities data store's design as discussed in https://github.com/WordPress/abilities-api/pull/60#discussion_r2352549262.

gziolo (approved):

packages/client/src/api.ts
gziolo: We should take a closer look at potential optimizations after we land the initial version of this PR. I shared some testing results and thoughts in https://github.com/WordPress/abilities-api/pull/60#issuecomment-3333440124. We could consider some enhancements where in the reducer for abilities we only list the names of the abilities registered on the server and retrieve the details through the entities registry. One aspect to keep in mind is that we need to track record of unregistered abilities for the cases where developers explicitly request to remove them from the usage on the client, or they opt to replace them with client-side first implementation.

emdashcodes:
> The only nitpick I have is that I still see location: 'client' in the ability object.

This was coming from my example abilities and not the code itself, but I agree invalid fields should be stripped alongside things like _links from the server response. 

I added some code for this and tests in 
https://github.com/WordPress/abilities-api/pull/60/commits/dea7b4063fce29bf6cfb3d0873140c7149bcac4f

PR #61, opened 2025-09-05
Fix `$input` type hint in `WP_Ability`

### LINKED ISSUE #58: Fix `$input` type hint in `WP_Ability`
The `WP_Ability` class uses `array $input` in several places.

While this is likely working for 90% of use-cases, the shape of the input depends on the input schema that is used. That input schema can signify any type that can be expressed in JSON schema - but the PHP `array` type hint only works for the JSON schema types `object` and `array`.

While based on my experience using a parameter scheme of `object` is extremely common, any other types are possible and supported by function calling capabilities of most popular LLMs.

As such, I think the `array` type hint on `$input` is a bug and should be fixed. It needs to be without a type hint, as it can be `mixed`.

gziolo:
Closes https://github.com/WordPress/abilities-api/issues/58.

> The `WP_Ability` class uses `array $input` in several places.
>
> While this is likely working for 90% of use-cases, the shape of the input depends on the input schema that is used. That input schema can signify any type that can be expressed in JSON schema - but the PHP `array` type hint only works for the JSON schema types `object` and `array`.

Given that `$input` is used with two public methods `execute`, `has_permission`, and other methods internally, the code in `WP_Ability` class was adjusted accordingly to use a unified approach. This changes the type hint from array to mixed to support all JSON schema types, including `null`, `boolean`, `number`, `integer`, `string`, `array`, and `object`.

- Updated all method signatures and documentation to use mixed instead of array for input parameters.
- Made the `$input` argument optional in `execute_callback` and `permission_callback` when the `input_schema` is not defined and the value is `null`.
- Updated REST API controller to properly handle mixed input types without assuming array format.
- Add new tests covering all possible input types.
- Modified test callbacks to match the new signatures where appropriate.

## Testing

All unit tests should pass:

```bash
npm run test:php
```

### DIFF
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -35,8 +35,8 @@ declare( strict_types = 1 );
  *   description?: string,
  *   input_schema?: array<string,mixed>,
  *   output_schema?: array<string,mixed>,
- *   execute_callback?: callable( mixed $input ): (mixed|\WP_Error),
- *   permission_callback?: callable( mixed $input ): (bool|\WP_Error),
+ *   execute_callback?: callable( mixed $input= ): (mixed|\WP_Error),
+ *   permission_callback?: callable( mixed $input= ): (bool|\WP_Error),
  *   meta?: array<string,mixed>,
  *   ability_class?: class-string<\WP_Ability>,
  *   ...<string, mixed>
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -35,8 +35,8 @@ declare( strict_types = 1 );
  *   description?: string,
  *   input_schema?: array<string,mixed>,
  *   output_schema?: array<string,mixed>,
- *   execute_callback?: callable( array<string,mixed> $input): (mixed|\WP_Error),
- *   permission_callback?: callable( array<string,mixed> $input ): (bool|\WP_Error),
+ *   execute_callback?: callable( mixed $input ): (mixed|\WP_Error),
+ *   permission_callback?: callable( mixed $input ): (bool|\WP_Error),
  *   meta?: array<string,mixed>,
  *   ability_class?: class-string<\WP_Ability>,
  *   ...<string, mixed>
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -56,8 +56,8 @@ final class WP_Abilities_Registry {
 	 *   description?: string,
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback?: callable( array<string,mixed> $input): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( array<string,mixed> $input ): (bool|\WP_Error),
+	 *   execute_callback?: callable( mixed $input ): (mixed|\WP_Error),
+	 *   permission_callback?: ?callable( mixed $input ): (bool|\WP_Error),
 	 *   meta?: array<string,mixed>,
 	 *   ability_class?: class-string<\WP_Ability>,
 	 *   ...<string, mixed>
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -56,8 +56,8 @@ final class WP_Abilities_Registry {
 	 *   description?: string,
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback?: callable( mixed $input ): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( mixed $input ): (bool|\WP_Error),
+	 *   execute_callback?: callable( mixed $input= ): (mixed|\WP_Error),
+	 *   permission_callback?: ?callable( mixed $input= ): (bool|\WP_Error),
 	 *   meta?: array<string,mixed>,
 	 *   ability_class?: class-string<\WP_Ability>,
 	 *   ...<string, mixed>
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -269,13 +269,23 @@ class WP_Ability {
 	 *
 	 * @since 0.1.0
 	 *
-	 * @param mixed $input The input data to validate.
+	 * @param mixed $input Optional. The input data to validate. Default `null`.
 	 * @return true|\WP_Error Returns true if valid or the WP_Error object if validation fails.
 	 */
 	protected function validate_input( $input = null ) {
 		$input_schema = $this->get_input_schema();
 		if ( empty( $input_schema ) ) {
-			return true;
+			if ( null === $input ) {
+				return true;
+			}
+			return new \WP_Error(
+				'ability_missing_input_schema',
+				sprintf(
+					/* translators: %s ability name. */
+					__( 'Ability "%s" does not define an input schema required to validate the provided input.' ),
+					$this->name
+				)
+			);
 		}
 
 		$valid_input = rest_validate_value_from_schema( $input, $input_schema, 'input' );
@@ -326,7 +336,7 @@ class WP_Ability {
 	 *
 	 * @since 0.1.0
 	 *
-	 * @param mixed $input The input data for the ability.
+	 * @param mixed $input Optional. The input data for the ability. Default `null`.
 	 * @return mixed|\WP_Error The result of the ability execution, or WP_Error on failure.
 	 */
 	protected function do_execute( $input = null ) {
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -329,7 +329,7 @@ class WP_Ability {
 	 * @param mixed $input The input data for the ability.
 	 * @return mixed|\WP_Error The result of the ability execution, or WP_Error on failure.
 	 */
-	protected function do_execute( $input ) {
+	protected function do_execute( $input = null ) {
 		if ( ! is_callable( $this->execute_callback ) ) {
 			return new \WP_Error(
 				'ability_invalid_execute_callback',
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -65,7 +65,7 @@ class WP_Ability {
 	 * The ability execute callback.
 	 *
 	 * @since 0.1.0
-	 * @var callable( mixed $input ): (mixed|\WP_Error)
+	 * @var callable( mixed $input= ): (mixed|\WP_Error)
 	 */
 	protected $execute_callback;
 
@@ -73,7 +73,7 @@ class WP_Ability {
 	 * The optional ability permission callback.
 	 *
 	 * @since 0.1.0
-	 * @var ?callable( mixed $input ): (bool|\WP_Error)
+	 * @var ?callable( mixed $input= ): (bool|\WP_Error)
 	 */
 	protected $permission_callback = null;
 
@@ -145,8 +145,8 @@ class WP_Ability {
 	 *   description: string,
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback: callable( array<string,mixed> $input): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( array<string,mixed> $input ): (bool|\WP_Error),
+	 *   execute_callback: callable( mixed $input= ): (mixed|\WP_Error),
+	 *   permission_callback?: ?callable( mixed $input= ): (bool|\WP_Error),
 	 *   meta?: array<string,mixed>,
 	 *   ...<string, mixed>,
 	 * } $args
@@ -314,6 +314,10 @@ class WP_Ability {
 			return true;
 		}
 
+		if ( empty( $this->get_input_schema() ) && null === $input ) {
+			return call_user_func( $this->permission_callback );
+		}
+
 		return call_user_func( $this->permission_callback, $input );
 	}
 
@@ -332,6 +336,10 @@ class WP_Ability {
 				/* translators: %s ability name. */
 				sprintf( __( 'Ability "%s" does not have a valid execute callback.' ), $this->name )
 			);
+		}
+
+		if ( empty( $this->get_input_schema() ) && null === $input ) {
+			return call_user_func( $this->execute_callback );
 		}
 
 		return call_user_func( $this->execute_callback, $input );
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -314,7 +314,7 @@ class WP_Ability {
 			return true;
 		}
 
-		if ( empty( $this->get_input_schema() ) && null === $input ) {
+		if ( empty( $this->get_input_schema() ) ) {
 			return call_user_func( $this->permission_callback );
 		}
 
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -65,7 +65,7 @@ class WP_Ability {
 	 * The ability execute callback.
 	 *
 	 * @since 0.1.0
-	 * @var callable( array<string,mixed> $input): (mixed|\WP_Error)
+	 * @var callable( mixed $input ): (mixed|\WP_Error)
 	 */
 	protected $execute_callback;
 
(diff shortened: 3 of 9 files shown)


gziolo:
This still needs some additional test coverage to account for the different scenarios for input values passed to `execute`. It uses internally `has_permission`, so the only case to ensure it's covered would be when the input is not set at all, and the new default `null` is automatically set.

gziolo:
@galatanovidiu, I would appreciate a sanity check in relation to the MCP Adapter. Would the changes proposed introduce the need for some additional handling there, as the input type might become a simple type now? The same question applies to the input schema, but that part I'm less afraid of, unless the input in MCP has some specific requirements.

includes/abilities-api/class-wp-ability.php
```diff
@@ -321,10 +321,10 @@ public function has_permission( array $input = array() ) {
 	 *
 	 * @since 0.1.0
 	 *
-	 * @param array<string,mixed> $input The input data for the ability.
+	 * @param mixed $input The input data for the ability.
 	 * @return mixed|\WP_Error The result of the ability execution, or WP_Error on failure.
 	 */
-	protected function do_execute( array $input ) {
+	protected function do_execute( $input ) {
```
galatanovidiu: **Ambiguous handling of `null` input in ability execution:**

Currently, when `$input` is `null` in the `do_execute()` method, it gets passed to the callback as `call_user_func($callback, null)`, meaning the callback receives null as its first argument.
This creates an ambiguity:

Case 1: Developer wants to call a callback with no arguments (e.g., `wp_get_current_user()`)
Case 2: Developer wants to pass `null` as a meaningful argument to the callback explicitly

**Problem scenarios:**

Callbacks that expect no parameters (like `wp_get_current_user()`) will receive an unexpected null parameter
No way to distinguish between **"no arguments"** vs **"null argument"**

**Proposed solutions:**

Use Reflection API to inspect the callback's parameter signature and make intelligent decisions:

If callback expects 0 parameters and input is null: call without arguments: call_user_func($callback)
Otherwise pass input as argument: call_user_func($callback, $input)

gziolo: The same problem existed before when this was an array and the default value was `[]`. I agree it would be nice to remove the ambiguity of the `null` value by not using it as the default value.

It would be a nice enhancement to support some built-in functions, as a possible target for `execute_callback`. However, it's a more nuanced topic, because `do_execute` would have to follow the same type `mixed ...$args` like `call_user_func` does, while it's currently a single param. It would be convenient for PHP implementers, but this becomes harder to describe with JSON schema and handle through the REST API. What is the approach you had in mind here?

gziolo:
In [e117cac](https://github.com/WordPress/abilities-api/pull/61/commits/e117cac36c6a2e7a20bca713812c2ed35e14f573) added tests that cover several types of input:
- boolean
- string
- number
- array

Still to cover:
- null
- object
- integer (the current test for number should probably use integer)

gziolo:
@felixarntz, this is in a shape where it could use some feedback, so we can align if this is going in the right direction.

includes/abilities-api/class-wp-ability.php
felixarntz: Why not make the decision for whether to pass `$input` based on the schema? I think using reflection for this is unnecessary and less efficient.

Case in point: `validate_input()` already returns `true` without checking anything if the input schema is empty.

Similarly, we could implement this as follows: If the input schema is empty, we don't pass `$input` to any callbacks (`execute_callback`, `permission_callback`).

* Makes sense because if there's no input schema, you don't expect any input.
* If you at the same time pass a callback that expects an input, you're doing it wrong.

includes/abilities-api.php
```diff
@@ -35,8 +35,8 @@
  *   description?: string,
  *   input_schema?: array<string,mixed>,
  *   output_schema?: array<string,mixed>,
- *   execute_callback?: callable( array<string,mixed> $input): (mixed|\WP_Error),
- *   permission_callback?: callable( array<string,mixed> $input ): (bool|\WP_Error),
+ *   execute_callback?: callable( mixed $input ): (mixed|\WP_Error),
+ *   permission_callback?: callable( mixed $input ): (bool|\WP_Error),
```
felixarntz: We need to change these to allow the callbacks to not expect any `$input` (relevant when `input_schema` is empty).

felixarntz (commented):
This looks good regarding getting rid of the array requirement. I left some thoughts regarding the handling of expecting no input / no parameters.

tests/unit/rest-api/wpRestAbilitiesRunController.php
```diff
@@ -608,7 +608,7 @@ public function test_output_validation_failure_returns_error(): void {
 					),
 					'required'   => array( 'status' ),
 				),
-				'execute_callback'    => static function ( $input ) {
+				'execute_callback'    => static function () {
```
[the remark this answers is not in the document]
gziolo: Good catch. At the same time, the plan is to make that optional as proposed in https://github.com/WordPress/abilities-api/pull/61#discussion_r2330648111.

includes/abilities-api.php
gziolo: Good point. It should be optional to define the `$input` based on the conversations. I'll update these type hints accordingly 👍🏻

gziolo: Addressed in https://github.com/WordPress/abilities-api/pull/61/commits/75d89cfb985d4908f5a36665c1d234bd30f886c3.

includes/abilities-api/class-wp-ability.php
gziolo: The idea shared by @felixarntz that addresses the concerns should be fully implemented in https://github.com/WordPress/abilities-api/pull/61/commits/19357994511dc341786e27d0f746d903ca81c138. I still need to increase test coverage.

gziolo:
It's ready for the final review. I addressed the feedback, increased test coverage, and updated existing documentation.

includes/abilities-api/class-wp-ability.php
```diff
@@ -269,10 +269,10 @@ public function get_meta(): array {
 	 *
 	 * @since 0.1.0
 	 *
-	 * @param array<string,mixed> $input Optional. The input data to validate.
+	 * @param mixed $input The input data to validate.
 	 * @return true|\WP_Error Returns true if valid or the WP_Error object if validation fails.
 	 */
-	protected function validate_input( array $input = array() ) {
+	protected function validate_input( $input ) {
```
felixarntz: Nit-pick: Not strictly necessary, but for clarity I think we should make `$input` optional even on the internal methods.
```suggestion
	protected function validate_input( $input = null ) {
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -322,10 +326,10 @@ public function has_permission( array $input = array() ) {
 	 *
 	 * @since 0.1.0
 	 *
-	 * @param array<string,mixed> $input The input data for the ability.
+	 * @param mixed $input The input data for the ability.
 	 * @return mixed|\WP_Error The result of the ability execution, or WP_Error on failure.
 	 */
-	protected function do_execute( array $input ) {
+	protected function do_execute( $input ) {
```
felixarntz: See above:
```suggestion
	protected function do_execute( $input = null ) {
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -269,10 +269,10 @@ public function get_meta(): array {
...
 	 *
-	 * @param array<string,mixed> $input Optional. The input data to validate.
+	 * @param mixed $input The input data to validate.
 	 * @return true|\WP_Error Returns true if valid or the WP_Error object if validation fails.
 	 */
-	protected function validate_input( array $input = array() ) {
+	protected function validate_input( $input ) {
 		$input_schema = $this->get_input_schema();
 		if ( empty( $input_schema ) ) {
 			return true;
```
felixarntz: Giving this another thought, should we really do it like that? I think it's problematic from a security standpoint:
* The `input_schema` is critical in validating that the input is expected, including from a security perspective.
* It is already required, which makes sense.
* The only reasonable explanation for providing an empty `$input_schema` is that the ability does not expect any input. (It makes sense that it's required regardless, to make that "opt out" explicit.) But technically, it's still possible to pass `$input` to the ability.
* **Problem:** This code interprets the lack of `$input_schema` as "let's not validate" - I think this is a security concern. Instead, it should interpret the lack of `$input_schema` as: If there is any input (i.e. `$input !== null`), this is invalid and it should return a `WP_Error`.

includes/abilities-api/class-wp-ability.php
```diff
@@ -314,6 +314,10 @@ public function has_permission( array $input = array() ) {
...
 		}
 
+		if ( empty( $this->get_input_schema() ) && null === $input ) {
+			return call_user_func( $this->permission_callback );
+		}
```
felixarntz: Related to my previous comment, we should not allow any unvalidated input to be sent through. In other words, this clause should only be dependent on whether there's an input schema.
```suggestion
		if ( empty( $this->get_input_schema() ) ) {
			return call_user_func( $this->permission_callback );
		}
```

felixarntz (changes requested):
@gziolo The type hint fix here LGTM, but I am realizing that there's IMO a security flaw in how the validation of the input is handled. It's not introduced by this PR, but it would be critical to fix, unless I'm missing something.

cc @justlevine

includes/abilities-api/class-wp-ability.php
gziolo: I wasn't sure about that, but I'm happy to propagate the default as `null` in all necessary places 👍🏻

includes/abilities-api/class-wp-ability.php
gziolo: I'm open to hardening this one. I wasn't sure what's the best way to go about it, so all it worth taking into consideration. It means it will become very strict in the sense that the only case when we allow not passing the input schema is when there is no input at all. Otherwise, even when passing `null` as value, we would have to validatr with input schema provided.

felixarntz: To me passing `null` as `$input` is the same as not passing any `$input` at all. If the developer provided an input schema that was only allowing `null`, it would be pointless to do so anyway because it doesn't have any meaning. And if the value `null` is being passed in a situation where passing another value is possible, that would only work in combination with providing a (oneOf) schema anyway.

So to clarify what I meant: If there's no input schema, it's fine as long as `$input === null`. Whether it was passed as `null` or whether it's just the `null` default shouldn't matter.

gziolo: > So to clarify what I meant: If there's no input schema, it's fine as long as `$input === null`. Whether it was passed as `null` or whether it's just the `null` default shouldn't matter.

I agree with the rationale. I committed your proposed changes, and I will update the rest of the codebase accordingly to make it work this way.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -56,8 +56,8 @@ final class WP_Abilities_Registry {
 	 *   description?: string,
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback?: callable( array<string,mixed> $input): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( array<string,mixed> $input ): (bool|\WP_Error),
+	 *   execute_callback?: callable( mixed $input= ): (mixed|\WP_Error),
+	 *   permission_callback?: ?callable( mixed $input= ): (bool|\WP_Error),
```
justlevine, line 60: And elsewhere I think

```suggestion
	 *   execute_callback?: callable( mixed $input ): (mixed|\WP_Error),
	 *   permission_callback?: ?callable( mixed $input ): (bool|\WP_Error),
```

justlevine (commented):
> This changes the type hint from array to mixed to support all JSON schema types, including null, boolean, number, integer, string, array, and object.

Seeing the diff reinforces my concern about how PHP's `mixed` is broader than "all JSON schema types". 

( I'm not sure why we want to allow for non-named arguments here, and still think that we should use `WP_REST_Controller` as the prior art here where only _named_ inputs (`'args': array<string,mixed>|null` ) are supported. I cannot think of a justifiable DX we're we (humans, tools, agents, implementations) would prefer to be fully-dependent on also having the `WP_Ability::$input_schema` to understand the shape, instead of inferring/juggling an array key. ( e.g. get_post( $input ) is always going to less semantic than get_post( $input['user_id'] )` )

If we are going to allow more than named argments , then we can at least be accurate about the PHP type, and use `null|bool|int|float|string|mixed[]|object`, instead of littering the codebase with an avoidable mixed.

Last time I bring all that up I promise 🤐)

---

Regarding the input schema validation, agree that semantically it should be `if ( null === $input_schema ) { return true; }` while `if ( empty( $input_schema ) ) { return new \WP_Error()` 

~A few search-replaces ( `$input= ` => `$input` ) and~ ([nevermind](https://github.com/WordPress/abilities-api/pull/61#discussion_r2334684345) ) LGTM

includes/abilities-api/class-wp-abilities-registry.php
gziolo: It's intentional to account for the case where we don't pass the first argument if there is no input schema and the value is `null`.

includes/abilities-api/class-wp-ability.php
gziolo: [c377af9](https://github.com/WordPress/abilities-api/pull/61/commits/c377af9dd8d0ad62f0ba302854cc465cf500dff5) should do the trick.

I will look into failing unit tests tomorrow as REST API controllers somehow depend on the old behavior.

includes/abilities-api/class-wp-abilities-registry.php
gziolo: It may not be what I wanted to express. The idea is that folks can define the callback without any params, and linters would be happy. It isn't the case when there is `mixed $input`.

justlevine: Never seen that syntax before. Neat!

justlevine: image

gziolo: I found an example confirming it's a thing [here](https://phpstan.org/writing-php-code/phpdoc-types#callables), in particular this:

> `callable(int, int=): string` (second parameter is optional)

I guess `?` is reserved for optional keys on arrays and properties on objects.

gziolo:
@justlevine, so you suggest we replace `mixed` with the explicit list of the allowed types that the REST API can handle? The list of primitive types is [here](https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#primitive-types), which should align with the unit tests I added. The type would be defined on the REST API schema as:

`(integer|number|boolean|string|array|object|null)` 

or more precisely

`'type' => [ 'integer', 'number', 'boolean', 'string', 'array', 'object', 'null' ]`

The order might be wrong as it matters here because of [Type Juggling](https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#type-juggling). Based on your feedback, we would also need to make the object type more structured. Although I see that it needs to implement `JsonSerializable`, this implies that it must contain keys that are strings.

gziolo:
I added specific types in the REST API schema in [a21105f](https://github.com/WordPress/abilities-api/pull/61/commits/a21105f8e43e6e3bf82c7e9c220d54b73e222c6a). @justlevine, what else would you like to have included? I may not fully understand your expectations. This looks good enough for me at this point.

includes/abilities-api/class-wp-ability.php
```diff
@@ -334,6 +348,10 @@ protected function do_execute( array $input ) {
 			);
 		}
 
+		if ( empty( $this->get_input_schema() ) && null === $input ) {
```
felixarntz: From a defensive coding perspective, we should not let any input go through that hasn't been validated. If there's no input schema, that's always the case.

```suggestion
		if ( empty( $this->get_input_schema() ) ) {
```

In practice, this change does not make any difference today. But it helps prevent accidental problems down the line, e.g. if someone later changes `validate_input()` to behave differently.

felixarntz (approved):
@gziolo This looks great to me, thank you for the updates!

Just one minor note that I think would be worth updating before merge.

includes/abilities-api/class-wp-ability.php
gziolo: Good catch. It’s an artifact from my initial implementation.

gziolo:
> what else would you like to have included? I may not fully understand your expectations. This looks good enough for me at this point.

@justlevine, I'm still happy to discuss and iterate when you have a chance to respond. Now, I will land this PR as it addresses a larger issue.

Issue #68, opened 2025-09-11
Make the `permission_callback` argument required


johnbillion:
Reference: https://core.trac.wordpress.org/ticket/50075 plus its comments. Observe that this design deficiency directly led to security vulnerabilities in at least three very popular plugins.

The tl;dr is that it was a mistake for the `permission_callback` argument to not be required when registering a REST API endpoint, for a few reasons:

* It makes it easy for a developer to overlook permissions as they aren't required to explicitly implement a permissions check
* It makes it easy for a typo in the property name to go unnoticed (arguably more of a schema validation problem, but still)

The same concerns apply to the `permission_callback` argument when registering an ability. This argument should be made mandatory and registration of an ability should fail hard when not present.

gziolo:
Good callout. It will still be relatively simple to pass `__return_true` if folks want to make the ability generally available. At the same time, as pointed out, this is going to be an explicit developer's decision 👍🏻

justlevine:
Handled in https://github.com/WordPress/abilities-api/pull/73 and awaiting review.

Issue #62, opened 2025-09-08
Add property for marking Ability hints like destructive, read-only, idempotent


JasonTheAdams:
Oftentimes we want to give the AI different instructions for when it's about to perform a destructive action, such as framing what's about to happen and requesting confirmation to the user. It would be helpful, therefore, to have a property of an Ability that marks it as a destructive ability.

gziolo:
@marcochiesi raised the same use case in https://github.com/WordPress/ai/issues/40. The related points he described are: 

> - **Abilities as "One More Way"**:  I envision the abilities API as "one more way" for users to achieve results, as an alternative to the WordPress admin UI. Of course, we must ensure users are fully aware of what's happening and confirm potentially destructive actions. For this, the Model Context Protocol (MCP) standard includes a feature called "elicitation," which allows the MCP server to request additional information from the user (such as confirmation) before executing a task. More information [here](https://modelcontextprotocol.io/docs/concepts/elicitation).
> - **Handling Destructive Operations**: For the reason above, I would not limit plugin management or other destructive operations, as long as we can guarantee that the user is forced to confirm before proceeding. We should simply ensure that LLMs do not perform such actions without explicit confirmation. I believe over-engineering with the possibility of "undo" actions is out of scope. Currently, when using the standard WordPress UI, the user cannot easily undo these destructive operations.

swissspidy:
Let's go beyond just "destructive" and think about other hints as well. For example, [MCP tools annotation](https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations)s include: read-only, destructive, idempotent, requires confirmation, whether it's "open world"

justlevine:
Recommend that for the initial implementation, we store this in `$meta` as it gives us more flexibility to iterate without worrying too much about coupling our API to transient concerns from an evolving spec, or one specific implementation (MCP) over another (e.g. Command Palette).

galatanovidiu:
Should we use user capabilities/roles on the abilities somehow, so we can easily filter abilities based on user capabilities before even presenting them to AI agents (in any form)?

I know we have the `has_permission` callback for granular permission checks at execution time. Still, a simple capability-based pre-filter could be beneficial for performance, especially in the MCP contexts - avoiding permission callback execution for abilities the user can't access anyway.

Something like:
```php
wp_register_ability( 'delete-post', [
    'label' => __('Delete Post'),
    'hints' => [
        'destructive' => true,
        'requires_confirmation' => true,
[...]

PR #69, opened 2025-09-11
Add client-only ability registration

emdashcodes:
This PR follows up on https://github.com/WordPress/abilities-api/pull/60 and introduces client-side only abilities. This allows for new integrations that deal with things like UI elements, navigation, and client-side data (such as the block editor). In addition to our AI efforts, it's also meant as a way that the command palette can execute client-side abilities in the future.

It adds two new methods to manage the client-only abilities `registerAbility` and `unregisterAbility`.

`registerAbility` takes the same shape as `wp_register_ability` in addition to a client side callback function. A callback can take the provided `input_schema`, do an action, and return the provided `output_schema`.

Here is a simple example:

```typescript
registerAbility( {
	name: 'demo/navigate-admin',
	label: 'Navigate to Admin Page',
	description: 'Navigates to a specific WordPress admin page',
	location: 'client',
	meta: {
		type: 'tool',
	},
	input_schema: {
		type: 'object',
		properties: {
			page: {
				type: 'string',
				description: 'The admin page to navigate to (e.g., "users.php", "options-general.php")',
			},
			params: {
				type: 'object',
				description: 'Optional query parameters',
				additionalProperties: true,
			},
		},
		required: [ 'page' ],
	},
	output_schema: {
		type: 'object',
		properties: {
			success: { type: 'boolean' },
			url: { type: 'string' },
		},
		required: [ 'success', 'url' ],
	},
	callback: async function( input ) {
		const adminUrl = clientAbilitiesDemo.adminUrl || '/wp-admin/';
		let url = adminUrl + input.page;

		// Add query parameters if provided
		if ( input.params ) {
			const params = new URLSearchParams( input.params );
			url += '?' + params.toString();
		}

		// Navigate to the URL
		window.location.href = url;

		return {
			success: true,
			url: url,
		};
	},
} );
```

`unregisterAbility` is a convenience method for removing client-side only abilities.

Client-side abilities are saved in the store alongside server-side ones. They are marked specifically with a location property which is used to either call the callback or call the REST endpoint. 

The input and output schemas are validated using [Ajv](https://www.npmjs.com/package/ajv) and the `ajv-formats` plugin.

I have configured Ajv to support the intersection of rules that JSON Schema draft-04, WordPress (a subset of JSON Schema draft-04), and providers like OpenAI and Anthropic support.

Ajv is [already in use by Gutenberg](https://github.com/search?q=repo%3AWordPress%2Fgutenberg+ajv&type=code) so it is already part of the ecosystem.

⚠️ Note: Tests and CI setup have been added in https://github.com/WordPress/abilities-api/pull/70

### Testing

```bash
cd packages/client

# Install dependencies
npm install

# Build the package
npm run build
```

I have created a dummy plugin  that registers a few different abilities as an example: [client-abilities-demo.zip](https://github.com/user-attachments/files/22280357/client-abilities-demo.zip)

Open the developer console and paste the following:

`await wp.abilities.listAbilities()`

You should see the client side abilities (and any other server side abilities you registered) listed.

Paste in the following:

`await wp.abilities.executeAbility('demo/get-current-user')`

See that you get a simple user response object back. 

Paste in the following:

```typescript
await wp.abilities.executeAbility('demo/navigate-admin', { 
	page: 'users.php' 
});
```

See that you are redirected to the users.php screen.

Paste in the following:

```typescript
wp.abilities.executeAbility('demo/show-notification', { 
	message: 'Hello from client ability!',
	type: 'success'
});
```

See that a UI notice briefly shows on the user page.

Now we can test registering and unregistering our own ability:

```typescript
wp.abilities.registerAbility({
    name: 'demo/console-log',
    label: 'Console Logger',
    description: 'Logs messages to the browser console with different levels',
    location: 'client',
    input_schema: {
        type: 'object',
        properties: {
            message: {
                type: 'string',
                description: 'Message to log',
            },
            level: {
                type: 'string',
                enum: ['log', 'info', 'warn', 'error', 'debug'],
                description: 'Console log level',
                default: 'log'
            },
            data: {
                description: 'Additional data to log',
            }
        },
        required: ['message']
    },
    output_schema: {
        type: 'object',
        properties: {
            logged: { type: 'boolean' },
            timestamp: { type: 'string' },
            message: { type: 'string' }
        }
    },
    callback: async ({ message, level = 'log', data }) => {
        const timestamp = new Date().toISOString();
        const prefix = `[WP Ability ${timestamp}]`;

        switch (level) {
            case 'info':
                console.info(prefix, message, data || '');
                break;
            case 'warn':
                console.warn(prefix, message, data || '');
                break;
            case 'error':
                console.error(prefix, message, data || '');
                break;
            case 'debug':
                console.debug(prefix, message, data || '');
                break;
            default:
                console.log(prefix, message, data || '');
        }

        return {
            logged: true,
            timestamp: timestamp,
            message: message
        };
    }
});
```

Now test the ability:

```typescript
// Simple log
await wp.abilities.executeAbility('demo/console-log', {
    message: 'Hello from WordPress abilities!'
});

// Warning with data
await wp.abilities.executeAbility('demo/console-log', {
    message: 'Low memory warning',
    level: 'warn',
    data: { available: '100MB', required: '500MB' }
});

// Error log
await wp.abilities.executeAbility('demo/console-log', {
    message: 'Failed to save settings',
    level: 'error',
    data: { code: 'NETWORK_ERROR', retry: true }
});

```

See that the various responses and different console types are executed. 

To test the validation, you can change the return type and see what happens if you try to return a different type or something invalid against the schema. 

Finally, you can unregister the client-side ability: `wp.abilities.unregisterAbility('demo/console-log');`. Trying to execute it again will throw an error.

### Next Steps (not addressed in this PR)

Tests & CI setup for the whole client package have been added in https://github.com/WordPress/abilities-api/pull/70


packages/client/src/api.ts
```diff
@@ -30,29 +31,142 @@ export async function getAbility( name: string ): Promise< Ability | null > {
...
+ * @param ability The ability definition including callback.
+ * @throws Error if ability is invalid or missing required fields.
+ *
+ * @example
+ * ```js
+ * registerAbility({
+ *   name: 'my-plugin/navigate',
+ *   label: 'Navigate to URL',
+ *   description: 'Navigates to a URL within WordPress admin',
+ *   location: 'client',
```
emdashcodes: Maybe `location` should be a part of `meta`? I made it a top level  though like the Feature API had. This is also only used in the client package. The server will only know about server side abilities.

packages/client/src/api.ts
```diff
@@ -30,29 +31,142 @@ export async function getAbility( name: string ): Promise< Ability | null > {
...
+	}
+	if ( ! ability.label ) {
+		throw new Error( 'Ability label is required' );
+	}
+	if ( ! ability.description ) {
+		throw new Error( 'Ability description is required' );
+	}
+	if ( ! ability.callback || typeof ability.callback !== 'function' ) {
+		throw new Error(
+			'Client abilities must include a callback function'
```
emdashcodes: We add a number of strings here for errors. These are for developers, not end users, but I'm guessing we might still want to wrap this in gettext calls.

swissspidy:
Do we need permission callbacks on the client as well?

> The input and output schemas are validated using [Ajv](https://www.npmjs.com/package/ajv) and the `ajv-formats` plugin.

Probably the best choice.

Curious, how big does this make the bundle?

> Ajv is [already in use by Gutenberg](https://github.com/search?q=repo%3AWordPress%2Fgutenberg+ajv&type=code) so it is already part of the ecosystem.

What do mean with "already part of the ecosystem"? It's a mere dev dependency in Gutenberg for use in some tests. It's not in any of the published packages. That's a big difference and not really an argument for picking this dependency.

emdashcodes:
> What do mean with "already part of the ecosystem"? It's a mere dev dependency in Gutenberg for use in some tests. It's not in any of the published packages. That's a big difference and not really an argument for picking this dependency.

It's also used in WordPress Playground and a few other utils too. I included it to say we have opted to use it in other places, which to me is worth considering over another package. I haven't looked at the bundle size (yet), but I personally think it is worth it instead of needing to hit a  `/validate` endpoint or something instead for client side abilities.

emdashcodes:
> Do we need permission callbacks on the client as well?

I was thinking about this as well and it's worth discussing. I think it would be nice to avoid needing to hit the REST API for running these, but maybe just providing a callback is enough? I'm open to ideas here!

gziolo:
https://bundlephobia.com/package/ajv-draft-04@1.0.0

Screenshot 2025-09-12 at 09 14 48

It's a very reasonable size for the functionality it provides. In the long run we could reuse the logic in Gutenberg to validate block attributes schema in the development mode.

---

> Do we need permission callbacks on the client as well?

It looks like on the server, we will enforce permission callbacks to increase the security as explained by @johnbillion in:
- https://github.com/WordPress/abilities-api/issues/68

It still might be valuable to have a way to define permission callbacks that checks whether a user can do something in the UI that's only for priviliged users so they have more streamlined experience. Otherwise we risk AI shows too often an message about failed attempe to do something it is not allowed to do. Similarily, @galatanovidiu in this comment https://github.com/WordPress/abilities-api/issues/62#issuecomment-3270506908 proposed a new property that does more high-level checks whether certain abilities should be filtered out if it's known upfront (without providing specific input) they aren't allowed to use them.

packages/client/src/api.ts
gziolo: This is a broader question on how to best group abilities or describe where they can be executed and who can do it. Similarily to whether the permission callback is needed as noted in https://github.com/WordPress/abilities-api/pull/69#issuecomment-3282768167.

In WP Feature API, was there some assumption that client features are sent to the server? I don't fully understand why we would have to explicitly annotate that as this could be done through the registration process. In fact, we might end up in the situation where we have three types of abilities:
- server-side only with `execute_callback` in PHP
- client-side only with `executeCallback` in JavaScript
- hybrid with `execute_callback` in PHP, and an alternative implementation on the client through `executeCallback` in JavaScript

The last one `hybrid` might be a good fit for entities from `@wordpress/core-data`. On the client, you could take advantage of optimistic updates, caching, etc, but on the server you can directly perform the same operations through WP functions.

packages/client/src/api.ts
gziolo: In PHP, even developer errors go through translation pipeline. In JavaScript things differ substantially. I'm not sure what the origin is, but looking at Gutenberg codebase, it should be fine to leave as is if it isn't user facing.

packages/client/src/api.ts
```diff
@@ -25,60 +31,200 @@ export async function listAbilities(): Promise< Ability[] > {
...
+	if (!ability) {
+		throw new Error(`Ability not found: ${name}`);
+	}
+
+	// Route to appropriate execution method
+	if (ability.location === 'client') {
+		return executeClientAbility(ability, input);
+	}
+
+	return executeServerAbility(ability, input);
```
gziolo: Taking further my thoughts from https://github.com/WordPress/abilities-api/pull/69/files#r2343288632, would it be enough to use `ability.callback` instead of the `location`? For server-only ability it should neven be defined, so it'll need to do the `apiFetch` call regardless.

packages/client/src/store/reducer.ts
```diff
@@ -24,25 +30,54 @@ const DEFAULT_STATE: Record< string, Ability > = {};
...
+			// Only allow unregistering client abilities
+			if (state[action.name].location !== 'client') {
+				// eslint-disable-next-line no-console
+				console.warn(
+					`Cannot unregister server-side ability: ${action.name}`
+				);
+				return state;
+			}
+			const newState = { ...state };
+			delete newState[action.name];
```
gziolo: Interesting case. What if the intention of the developer is to prevent exposure and execution of some server-side abilities and they opt to do it through JavaScript? I know they could do it also with PHP, but it's a choice.

I'm curious to hear the rationale behind this additional check. Is there something that could break? Would it re-register when someone asks for the same ability?

packages/client/src/store/resolvers.ts
```diff
@@ -27,28 +27,28 @@ export function getAbilities() {
 			{ per_page: PER_PAGE, page: 1 }
 		);
 
-		if ( ! firstPage || firstPage.length === 0 ) {
-			dispatch( receiveAbilities( [] ) );
+		if (!firstPage || firstPage.length === 0) {
```
gziolo: It's probably more feedback for the parent branch (on this diff it is most likely due to Prettier formatting), but this can be simplified by passing `page: -1` which will automatically fetch all pages through a special fetch all middleware:

https://github.com/WordPress/gutenberg/blob/3cf0226ee98012dc24b0dc5f2d59715d13d8e1e7/packages/api-fetch/src/middlewares/fetch-all-middleware.ts#L66-L72

packages/client/src/types.ts
```diff
@@ -31,13 +38,27 @@ export interface Ability {
...
-	output_schema?: Record< string, any >;
+	output_schema?: Record<string, any>;
+
+	/**
+	 * Where the ability is executed.
+	 * 'server' means it's executed via REST API on the server.
+	 * 'client' means it's executed locally in the browser.
+	 * Defaults to 'server'.
+	 */
+	location?: 'server' | 'client';
```
gziolo: The more I see the code, I'm incrasingly confident it's redundant as it can be inferred from the existance of the optional `callback` property.

Aside, technically speaking a client-side ability could still use `callback` to call purely server-side ability. I'm mostly pointing out this distinction is mostly about where the callback's execution starts.

packages/client/src/validation.ts
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * Validates a value against a JSON Schema.
+ *
+ * @param value The value to validate.
+ * @param args  The JSON Schema to validate against.
+ * @param param Optional parameter name for error messages.
+ * @return True if valid, error message string if invalid.
+ */
+export function validateValueFromSchema(
```
gziolo, line 144: This is nice! We might even find it a better home in a standalone WordPress package. I see extensive test coverage in https://github.com/WordPress/abilities-api/pull/70 👍🏻

packages/client/src/api.ts
```diff
@@ -25,60 +31,200 @@ export async function listAbilities(): Promise< Ability[] > {
...
+ *   },
+ *   callback: async ({ url }) => {
+ *     window.location.href = url;
+ *     return { success: true };
+ *   }
+ * });
+ * ```
+ */
+export function registerAbility(ability: ClientAbility): void {
+	if (!ability.name) {
```
gziolo: We should consider additional check to ensure we don't override accidentally an existing ability. This would mirror how it works on the server, and how WordPress usually handles registries.

gziolo (commented):
This is shaping up nicely. I left my feedback for consideration and to better understand some decisions made.

packages/client/src/api.ts
emdashcodes: These ended up translated in #60.

packages/client/src/store/resolvers.ts
emdashcodes: Yup, this was fixed in #60!

packages/client/src/api.ts
emdashcodes: In the original proposal for the WP Feature API, the registry was going to handle the location differently and maybe expose things differently. But I agree that we don't really need that right now unless a need arises. 

Looking through the other pieces of feedback you left, I agree we can just depend on a callback and hybrid ones can do what they need to. 

This has been updated in https://github.com/WordPress/abilities-api/pull/69/commits/5255a1e385e3dddfd1169ca8d5c394fb89e0f7ba.

packages/client/src/api.ts
emdashcodes: In https://github.com/WordPress/abilities-api/pull/69/commits/5255a1e385e3dddfd1169ca8d5c394fb89e0f7ba, we just check for `ability.callback` now.

packages/client/src/store/reducer.ts
emdashcodes: My intention here was to limit registerAbility and unregisterAbility to client abilities only. But considering we're removing that distinction, and abilities will just have a callback or not, I went ahead and removed this, and we can unregister anything now. 

Done in https://github.com/WordPress/abilities-api/pull/69/commits/5255a1e385e3dddfd1169ca8d5c394fb89e0f7ba

packages/client/src/types.ts
emdashcodes: Agreed, this was removed in https://github.com/WordPress/abilities-api/pull/69/commits/5255a1e385e3dddfd1169ca8d5c394fb89e0f7ba.

packages/client/src/validation.ts
emdashcodes: I did think some kind of lightweight validation package might make sense. I could see a validator package with AJV defaults being a thing in the future. Perhaps as part of the move to Gutenberg?

emdashcodes:
Regarding permissions, I have added a permissionCallback in https://github.com/WordPress/abilities-api/pull/69/commits/f7a3dd2b4fd24261959a4226313a3d71bd74ff90. It will run before an ability is executed. The implementation is intentionally left up to the user. This could be a simple JS based check, or they could hit the API or use nonces. The server permission callback will also still run when the REST API endpoint is called.

packages/client/src/api.ts
emdashcodes: Good call. I added a check here: https://github.com/WordPress/abilities-api/pull/69/commits/4c6ba7dc9bd3dac63e17717c6c661f8c44f7bc1c.

emdashcodes:
All PR feedback has been handled correctly now. I'll continue fixing the tests and any other things pointed out in #70.

gziolo (approved):
All the feedback I shared was addressed. It looks good to land from my side.

gziolo:
I merged #70 into this PR. If all CI checks still pass, I will merge everything together into #60 to review and test all the related changes integrated.

PR #70, opened 2025-09-11
Add client library tests and code quality infrastructure 

emdashcodes:
This PR adds test coverage and code quality to the client package. It follows-up on the following two PRs: https://github.com/WordPress/abilities-api/pull/60 and https://github.com/WordPress/abilities-api/pull/69.

It adds:
- Jest test setup based on `wp-scripts`
- 95% code coverage for client code
- ESLint & Prettier setup based on `wp-scripts`
- Contributor documentation
- CI infrastructure

Things are broken down into a few separate commits to make things easier to review. Apologizes again for `package-lock.json` inflation.

- [Add testing & linting infrastructure](https://github.com/WordPress/abilities-api/pull/70/commits/8f6352dd7a4190c11ce6fbee4bb0bcd0ee1adbb5) https://github.com/WordPress/abilities-api/pull/70/commits/8f6352dd7a4190c11ce6fbee4bb0bcd0ee1adbb5
- [Minor formatting fixes](https://github.com/WordPress/abilities-api/pull/70/commits/50722d51f755949e648b1200a53f44fa2c357814) https://github.com/WordPress/abilities-api/pull/70/commits/50722d51f755949e648b1200a53f44fa2c357814
- [Client package tests](https://github.com/WordPress/abilities-api/pull/70/commits/1ffe05df47cac5a6c07de6633c6690813e82a705) https://github.com/WordPress/abilities-api/pull/70/commits/1ffe05df47cac5a6c07de6633c6690813e82a705

## Testing 

- See that all CI checks pass below

- Locally

```bash
# Lint all JavaScript/TypeScript files
npm run lint:js

# Run TypeScript typecheck
npm run typecheck

# Run all JavaScript tests
npm run test:client
```


packages/client/package.json
```diff
@@ -43,19 +43,28 @@
...
 		"@wordpress/core-data": "^7.30.0",
 		"@wordpress/data": "^10.30.0",
-		"ajv-draft-04": "^1.0.0"
+		"ajv-draft-04": "^1.0.0",
+		"ajv-formats": "^3.0.1"
 	},
 	"devDependencies": {
+		"@testing-library/jest-dom": "^6.1.5",
+		"@testing-library/react": "^14.1.2",
+		"@types/jest": "^30.0.0",
```
gziolo: I noticed that `@types/wordpress__api-fetch` should not be needed as the package ships types.

Double-checked on npm and GitHub:

https://github.com/WordPress/gutenberg/blob/3cf0226ee98012dc24b0dc5f2d59715d13d8e1e7/packages/api-fetch/package.json#L30

packages/client/src/store/__tests__/actions.test.ts
```diff
@@ -0,0 +1,131 @@
...
+			const ability: Ability = {
+				name: 'test/server-ability',
+				label: 'Server Ability',
+				description: 'Server-side ability',
+				location: 'server',
+				input_schema: { type: 'object' },
+				output_schema: { type: 'object' },
+			};
+
+			const action = registerAbility(ability);
```
gziolo, line 109: I noticed that the action doesn't have opinions about registering a server-side ability, but the more public-facing API throws an Error. It's worth noting that actions are also publicly available, so there should be consistency. There are cases in Gutenberg where actions perform validation, like block type registration:

https://github.com/WordPress/gutenberg/blob/3cf0226ee98012dc24b0dc5f2d59715d13d8e1e7/packages/blocks/src/store/process-block-type.js#L75C14-L75C30

packages/client/src/store/__tests__/resolvers.test.ts
```diff
@@ -0,0 +1,304 @@
...
+				ENTITY_KIND,
+				ENTITY_NAME,
+				'non-existent'
+			);
+			expect(mockDispatch).not.toHaveBeenCalled();
+		});
+
+		it('should handle namespaced ability names', async () => {
+			const mockAbility: Ability = {
+				name: 'my-plugin/feature/action',
```
gziolo: This would be inconsistent with the server implementation where the names has a specific shape and has to be unique:

https://github.com/WordPress/abilities-api/blob/2796a60272785c3482520b7ef1e30154f6b83a20/includes/abilities-api/class-wp-abilities-registry.php#L67-L86

packages/client/src/__tests__/api.test.ts
```diff
@@ -0,0 +1,520 @@
...
+ * WordPress dependencies
+ */
+import { dispatch, resolveSelect } from '@wordpress/data';
+import apiFetch from '@wordpress/api-fetch';
+
+/**
+ * Internal dependencies
+ */
+import {
+	listAbilities,
```
gziolo: Could it be renamed to align better with `wp_get_abilities`. How about `getAbilities`?


For other methods, it aligns nicely with the PHP counterpart:
- `wp_get_ability` -> `getAbility`
- `wp_register_ability` -> `registerAbility`
- `wp_unregister_ability` -> `unregisterAbility`

packages/client/src/__tests__/api.test.ts
```diff
@@ -0,0 +1,520 @@
...
+	getAbility,
+	registerAbility,
+	unregisterAbility,
+	executeAbility,
```
gziolo, line 19: `executeAbility` would be the most considerable difference with the server. This is more high-level feedback, which doesn't have to be immediately addressed, more like a discussion starter.

Would it make sense to mirror that on the server and offer `wp_execute_ability`?

In the same fashion, could `getAbility` always return an object that would allow calling:

```js
const ability = await getAbility( 'my/ability' );
const result = await ability.execute( input );
```

Aside, all tests deal with sync callbacks, but I noticed that some examples documented could be async. We also have server abilities that don't have the callback, but maybe we could provide a general handler there, too.

.github/workflows/js-lint.yml
```diff
@@ -0,0 +1,147 @@
...
+          cache: 'npm'
+          node-version-file: '.nvmrc'
+
+      - name: Install NPM dependencies
+        run: npm ci
+
+      - name: Check Prettier formatting
+        run: |
+          cd packages/client
+          npx prettier --check "**/*.{js,jsx,ts,tsx,json,md}" \
```
gziolo: Why not reuse `npx wp-scripts format` or `npm run format`?

.github/workflows/test.yml
```diff
@@ -247,3 +247,62 @@ jobs:
...
+
+      - name: Run JavaScript linting
+        run: |
+          cd packages/client
+          npm run lint:js
+
+      - name: Run TypeScript type checking
+        run: |
+          cd packages/client
+          npm run typecheck
```
gziolo: What's the difference with the other workflow that runs the same commands? In a monorepo, it's perfectly fine to run everything from the root of the project.

packages/client/package.json
```diff
@@ -43,19 +43,28 @@
...
 	"dependencies": {
 		"@wordpress/api-fetch": "^7.30.0",
 		"@wordpress/core-data": "^7.30.0",
 		"@wordpress/data": "^10.30.0",
-		"ajv-draft-04": "^1.0.0"
+		"ajv-draft-04": "^1.0.0",
+		"ajv-formats": "^3.0.1"
 	},
 	"devDependencies": {
+		"@testing-library/jest-dom": "^6.1.5",
```
gziolo: Dev dependencies could also get moved to the root of the project. With namespaces, it's also possible to use only one lock file that will control everything. In effect, only production `dependencies` would be listed here, and most of the npm specific dance would happen in `package.json` and `package-lock.json` in the root of the repository.

packages/client/src/__tests__/validation.test.ts
```diff
@@ -0,0 +1,429 @@
+/**
+ * Tests for schema validation utilities.
+ */
+
+import { validateValueFromSchema } from '../validation';
+
+describe('validateValueFromSchema', () => {
```
gziolo, line 7: Nice coverage here 💯

packages/client/src/__tests__/validation.test.ts
```diff
@@ -0,0 +1,429 @@
...
+			expect(validateValueFromSchema([1, 'two', 3], schema)).toBe(
+				'[1] is not of type number.'
+			);
+		});
+	});
+
+	describe('edge cases', () => {
+		it('should pass validation when no schema provided', () => {
+			expect(validateValueFromSchema('anything', {})).toBe(true);
+			expect(validateValueFromSchema('anything')).toBe(true);
```
gziolo: I checked how `rest_validate_value_from_schema()` is implemented on the server. It looks like it expects an array, so the second parameter is mandatory. In addition, if the `type` property is not defined or invalid, it will call `_doing_it_wrong()` but continue execution.

I don't know how much we would like to align here, but since a similar validation happens on the server for both schemas defined by the ability, staying as close as possible would be favorable.

gziolo (commented):
I reviewed this PR and left my feedback to consider.

justlevine (commented):
I made the mistake of reviewing #60 before looking at this. Tl;Dr

1. I think we'll all have a much easier time if we treat this like a monorepo, use NPM workspaces and run all the dev/build commands top-level, and more generally follow the patterns used in the Gutenberg repo (whether that's the eventual home of a merged package or not, it's still peer-reviewed prior art that contributors are already familiar with)

2. Either way, not sure I follow the need for two GH workflows. If this tiny public repo really needs some `paths` filtering, then we can do it with https://github.com/dorny/paths-filter or similar.  (Otherwise at least filter `packages/*` _out_ of `test.yml` and do all the JS related stuff, including testing in ~js-lint.yml~ `js-test.yml`

jonathanbossenger (approved):
Commenting only on the documentation updates (which is where my focus has been for this project), it all looks good to me.

packages/client/src/__tests__/api.test.ts
emdashcodes: Agreed, this was renamed in #60.

packages/client/package.json
emdashcodes: Thanks for pointing that out. Updated in https://github.com/WordPress/abilities-api/pull/70/commits/bcc4a1e5748219148e3dda5ac4628f5c2435cf49.

packages/client/src/store/__tests__/actions.test.ts
emdashcodes: I have updated validation to be handled in the store now. See https://github.com/WordPress/abilities-api/pull/70/commits/54896816d01fe1ac3227c3fbf7e879084654828c.

packages/client/src/store/__tests__/resolvers.test.ts
emdashcodes: I have updated the validation to check for this format as well, fixed these tests, and added a new test for the validation. See https://github.com/WordPress/abilities-api/pull/70/commits/7c5ab6b1873f63f77d39c7e0e77bce9820988e40

.github/workflows/test.yml
emdashcodes: I had actually meant to clean up the extra entries _here_, but I agree that considering the size of the repo, we can just run these jobs from the main test.yml and there is no need for another workflow. Cleaned things up in https://github.com/WordPress/abilities-api/pull/70/commits/bfcea31bc8aa45c6292bd482671560e127ab3018.

.github/workflows/js-lint.yml
emdashcodes: Fixed in https://github.com/WordPress/abilities-api/pull/70/commits/bfcea31bc8aa45c6292bd482671560e127ab3018

packages/client/package.json
emdashcodes: Cleaned up some remaining ones here and removed the client package-lock.json. See https://github.com/WordPress/abilities-api/pull/70/commits/2828f26a371d0b13e3a95bcc70b697ca6227db0b

packages/client/src/__tests__/validation.test.ts
emdashcodes: I have not been sure how to handle these `_doing_it_wrong` calls on the client. Should they just throw `console.warning`?

packages/client/src/__tests__/api.test.ts
emdashcodes: `wp_execute_ability` does seem like it could be useful.  I need to think about this discussion more but wanted to note that I saw it.

`ability.execute` feels a little odd to me at initial glance 🤔

packages/client/src/__tests__/validation.test.ts
emdashcodes: I have updated the second parameter (schema) to be mandatory, and am better matching the server behavior here. See https://github.com/WordPress/abilities-api/pull/70/commits/5d50358a99022c7ab02f5700ca812a2013934438

packages/client/src/__tests__/api.test.ts
gziolo: @copilot, what are your thoughts on this thread?

Let's see if it can respond here 😄

gziolo: > ability.execute feels a little odd to me at initial glance 🤔

Right, it doesn't translate that well into JavaScript code.

packages/client/src/__tests__/validation.test.ts
gziolo: Yes, `console.warning` is used in many places. I personally prefer:

```js
import warning from '@wordpress/warning';

warning( 'Something to report' );
```

It has built-in integration with the build tooling so it only gets printed in the development mode, but completely eleminted from the source for production usage. It tries to replicate `SCRIPT_DEBUG` or `WP_DEBUG` on the server.

gziolo (approved):
I don't have more feedback for this PR other than a nitpick regarding usage `console.warning` over `warnining` from `@wordpress/warning`. It looks like all my important concerns raised where addressed.

It still needs some testing, which I plan to do tomorrow.

@emdashcodes, I'm not entirely sure how you would like to go about merging the code, but since it isn't targeting `trunk` I figured out, I approve it in case you would like to combine #69 and #70 together. It still will require testing, so this way I would review changes applied to #69 and test both PR at once.

PR #72, opened 2025-09-13
chore: update dev-deps to latest

justlevine:
## What

This PR bumps the Composer / NPM / GitHub Action _dev_ dependencies to their latest (compatible) versions. There are no changes to production code 

In addition, `php-coveralls/php-coveralls` has been removed from the Composer dev deps as we use codecov


gziolo (approved):
LGTM.

PR #73, opened 2025-09-13
feat!: make `permission_callback` arg required

### LINKED ISSUE #68: Make the `permission_callback` argument required

justlevine:
## What

This PR makes `permission_callback` a required argument.

> [!WARNING]
> This is a **breaking change**. 
> As a result of this change, Abilities that do not include a `permission_callback` will now return a `WP_Error`.

## Why

- Fixes #68

## How

Specific implementation notes on diff.

### DIFF
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -33,10 +33,10 @@ declare( strict_types = 1 );
  * @phpstan-param array{
  *   label?: string,
  *   description?: string,
- *   input_schema?: array<string,mixed>,
- *   output_schema?: array<string,mixed>,
  *   execute_callback?: callable( mixed $input= ): (mixed|\WP_Error),
  *   permission_callback?: callable( mixed $input= ): (bool|\WP_Error),
+ *   input_schema?: array<string,mixed>,
+ *   output_schema?: array<string,mixed>,
  *   meta?: array<string,mixed>,
  *   ability_class?: class-string<\WP_Ability>,
  *   ...<string, mixed>
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -46,18 +46,17 @@ final class WP_Abilities_Registry {
 	 * @param string              $name The name of the ability. The name must be a string containing a namespace
 	 *                                  prefix, i.e. `my-plugin/my-ability`. It can only contain lowercase
 	 *                                  alphanumeric characters, dashes and the forward slash.
-	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should include
-	 *                                  `label`, `description`, `input_schema`, `output_schema`,
-	 *                                  `execute_callback`, `permission_callback`, `meta`, and ability_class.
+	 * @param array<string,mixed> $args An associative array of arguments for the ability. See wp_register_ability() for
+	 *                                  details.
 	 * @return ?\WP_Ability The registered ability instance on success, null on failure.
 	 *
 	 * @phpstan-param array{
 	 *   label?: string,
 	 *   description?: string,
+	 *   execute_callback?: callable( mixed $input= ): (mixed|\WP_Error),
+	 *   permission_callback?: callable( mixed $input= ): (bool|\WP_Error),
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback?: callable( mixed $input= ): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( mixed $input= ): (bool|\WP_Error),
 	 *   meta?: array<string,mixed>,
 	 *   ability_class?: class-string<\WP_Ability>,
 	 *   ...<string, mixed>
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -73,9 +73,9 @@ class WP_Ability {
 	 * The optional ability permission callback.
 	 *
 	 * @since 0.1.0
-	 * @var ?callable( mixed $input= ): (bool|\WP_Error)
+	 * @var callable( mixed $input= ): (bool|\WP_Error)
 	 */
-	protected $permission_callback = null;
+	protected $permission_callback;
 
 	/**
 	 * The optional ability metadata.
@@ -143,15 +143,16 @@ class WP_Ability {
 	 * @phpstan-return array{
 	 *   label: string,
 	 *   description: string,
+	 *   execute_callback: callable( mixed $input= ): (mixed|\WP_Error),
+	 *   permission_callback: callable( mixed $input= ): (bool|\WP_Error),
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   execute_callback: callable( mixed $input= ): (mixed|\WP_Error),
-	 *   permission_callback?: ?callable( mixed $input= ): (bool|\WP_Error),
 	 *   meta?: array<string,mixed>,
 	 *   ...<string, mixed>,
 	 * } $args
 	 */
 	protected function prepare_properties( array $args ): array {
+		// Required args must be present and of the correct type.
 		if ( empty( $args['label'] ) || ! is_string( $args['label'] ) ) {
 			throw new \InvalidArgumentException(
 				esc_html__( 'The ability properties must contain a `label` string.' )
@@ -164,6 +165,19 @@ class WP_Ability {
 			);
 		}
 
+		if ( empty( $args['execute_callback'] ) || ! is_callable( $args['execute_callback'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties must contain a valid `execute_callback` function.' )
+			);
+		}
+
+		if ( empty( $args['permission_callback'] ) || ! is_callable( $args['permission_callback'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties must provide a valid `permission_callback` function.' )
+			);
+		}
+
+		// Optional args only need to be of the correct type if they are present.
 		if ( isset( $args['input_schema'] ) && ! is_array( $args['input_schema'] ) ) {
 			throw new \InvalidArgumentException(
 				esc_html__( 'The ability properties should provide a valid `input_schema` definition.' )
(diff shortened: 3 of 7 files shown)


includes/abilities-api.php
```diff
@@ -33,10 +33,10 @@
  * @phpstan-param array{
  *   label?: string,
  *   description?: string,
- *   input_schema?: array<string,mixed>,
- *   output_schema?: array<string,mixed>,
```
justlevine, line 35: Required before optional.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -46,18 +46,17 @@ final class WP_Abilities_Registry {
 	 * @param string              $name The name of the ability. The name must be a string containing a namespace
 	 *                                  prefix, i.e. `my-plugin/my-ability`. It can only contain lowercase
 	 *                                  alphanumeric characters, dashes and the forward slash.
-	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should include
```
justlevine, line 49: Seems silly to have so many sources of truth, happy to have one less now.

includes/abilities-api/class-wp-ability.php
```diff
@@ -320,14 +323,6 @@ public function has_permission( $input = null ) {
 			return $is_valid;
 		}
 
-		if ( ! is_callable( $this->permission_callback ) ) {
-			return true;
-		}
-
-		if ( empty( $this->get_input_schema() ) ) {
-			return call_user_func( $this->permission_callback );
-		}
```
justlevine, line 329: This is handled by `::validate_input()`

tests/unit/rest-api/wpRestAbilitiesRunController.php
```diff
@@ -703,38 +703,6 @@ public function test_ability_without_type_defaults_to_tool(): void {
 		$this->assertEquals( 200, $post_response->get_status() );
 	}
 
-	/**
-	 * Test permission check with null permission callback.
-	 */
-	public function test_permission_check_passes_when_callback_not_set(): void {
```
justlevine, line 709: Nothing to test anymore on REST, since if there's no callback we cant register the ability itself.

tests/unit/abilities-api/wpAbilitiesRegistry.php
justlevine, line 1: This is here because reordering the `prepare_properties()` by required/not-required (et al) seemed like the least intrusive way to do this, but it's more of an Integration test pattern and not pure PHPUnit.

If wp-core needs vanilla unit tests, we can relocate and refactor these tests to use `new WP_Ability()`

includes/abilities-api/class-wp-abilities-registry.php
gziolo: It's expanded with PHPStan annotation so it's perfectly fine to reference a canonical place 👍🏻

includes/abilities-api/class-wp-ability.php
gziolo: The point is that, if there is no schema, then we don't pass any input. What we agreed upon was that if someone wants to explicitly pass `null`, they need to provide the schema. This builds predictability, and that's why the condition is in place also here.

tests/unit/rest-api/wpRestAbilitiesRunController.php
gziolo: True, it's covered elsewhere 👍🏻

gziolo (commented):
I had only one minor feedback where I explained the reasoning why to keep extra logic. The rest of the changes look exactly as expected 💯

includes/abilities-api/class-wp-ability.php
gziolo: Previously, I considered adding a method for calling these callbacks, but it felt like more code than necessary. However, having a method could help document this expected behavior better.

justlevine: > What we agreed upon was that if someone wants to explicitly pass null, they need to provide the schema.

Can you clarify? If I understood you correctly that behavior is what this change enforces.

Previously diff, if there was no input schema, then any `$input` would get stripped and the execution callback would be be handled as if nothing was passed. As a result, if the schema allows an "explicit `null`, then the ability would still execute even if the supplied `$input` doesn't meet the correct shape.

With this diff, both no `$input` must be passed _and_ the schema must support a `null` value for the ability to execute successfully.

(If you're referring to _our internal_ ability to differentiate between `null` and **`unset`**, that was  just one one of the concerns about going to `mixed` that I raised on #58 . Until core's minimum PHP supports named arguments, nothing can compete with either the dx or future-compat of a partially-sealed array, nullable or otherwise. 🤷‍♂️)

gziolo: I'm not sure I follow, so let me explain it in a different way:

- if there is no input schema declared, it means there is no input to pass, so the callback should be executed with no arguments
- If there is an input schema provided, it means it needs to be validated. This also allows passing an explicit `null` as the first argument, but also any other type of data: `string`, `boolean`, `array`, etc.

justlevine: > if there is no input schema declared, it means there is no input to pass, so the callback should be executed with no arguments

If there is no input schema declared, but a user/mcp/traditional-tool nevertheless passes an `$input` what should the expected behavior be? E.g.

```php
$post_getter => wp_get_ability( 'my-custom/get-posts ' );

$posts               = $post_getter->execute(); 
$posts_explicit_null = $post_getter->execute( null );
$just_products       = $post_getter->execute( [ 'post_type' => 'product' ] );
```

- `$posts` will return an error if an `input_schema` is defined, but will proceed if an input schema is _not_ defined. (✔️ both ways)
- `$posts_explicit_null`
   - old: will pass if no `input_schema` is defined, will fail if an `input_schema` is defined but does not allow for an explicit null.
   - this pr: will _fail_ both if the `input_schema` is defined but doesn't allow for an explicit null **and** if no `input_schema` is defined. (because you're not passing "nothing", you're explicitly passing a `null`) 
   
- `$just_products` : Here's the IRL dx footgun:
  - old: if no `input_schema` is defined (IRL: filtering isn't supporting for by this specific ability), it will **return the same results as both `$posts` and `$post_explicit_null`. Only if there _is_ an `input_schema` defined will you get a WP_Error saying that e.g. the expected argument is called `type` not `post_type` .
     (or worse, that you should have called `$get_all_posts->execute( int $amount_of_posts )` and that ability will never be able to support post type filtering without breaking schema changes. This was my argument against #58 and why `mixed` is arguably the worst thing we could do to this or any public API 🤷)
  - this: will _fail_ both  if the `input_schema` is defined (as it should), but _also_ if there is no `input_schema`, correctly alerting the user that this is likely not the `WP_Ability` they're looking for.

I believe the behavior here is still ideal, even if the change to `mixed` negates a small part of the DX win.

Either way once clarified, I'll add backfill the necessary unit test to guarantee the behavior here doesn't change unintentionally

gziolo: If there is no input schema, and there is any input provided, then `validate_input()` kicks in. Reference:

https://github.com/WordPress/abilities-api/blob/59d8ec6869c54ab26ef897efd36f1bb7edd348c5/includes/abilities-api/class-wp-ability.php#L275-L289

So there are no concerns that someone passes an input that can't be validated due to a missing input schema. That's why we can safely skip the argument, making a better experience for devs as they don't have to always include `( $input = null )` for the callbacks that don't consume anything.

It's all a minor thing about ensuring we don't pass the default `null` when it is not needed.

justlevine: Okay now I'm sure we're not following each other, because that's literally the initial comment I left here explaining why I removed this conditional here, except you're using it to (seemingly) justify the opposite of what I am 😭

I'll add the unit tests for the above edge cases, if that doesn't clear things up (for either of us), i'll revert here and open a separate bug report that follows up #61 .

gziolo: Works for me.

felixarntz: The key here is that it needs to be impossible to pass unvalidated input through to `permission_callback` (and similarly `execute_callback`).

* @justlevine You're right that technically we can drop the `empty( $this->get_input_schema() )` condition because if this is the case and the input isn't `null`, the `validate_input` method would return a `WP_Error` and this code wouldn't be reached.
* That being said, I do think this is good to keep as a safe guard for good measure. We have the same in `do_execute`. Having this extra check in there ensures directly before the relevant call that we don't pass any input (i.e. enforce `null`) when there's no input schema.

Is it necessary? No. Is it a good measure for defensive coding and clarifying the importance of this to someone new working on this code in the future? Yes, I think so.

IMO we should just keep it. It doesn't hurt.

gziolo: Can we bring it back here as discussed?

```php
if ( empty( $this->get_input_schema() ) ) {
	return call_user_func( $this->permission_callback );
}
```

That's the only remaining item to approve and land this PR.

includes/abilities-api/class-wp-ability.php
```diff
@@ -320,14 +323,6 @@ public function has_permission( $input = null ) {
...
-
 		return call_user_func( $this->permission_callback, $input );
```
gziolo, line 330:
```suggestion
		if ( empty( $this->get_input_schema() ) ) {
			return call_user_func( $this->permission_callback );
		}

		return call_user_func( $this->permission_callback, $input );
```

includes/abilities-api/class-wp-ability.php
gziolo: I addressed it with https://github.com/WordPress/abilities-api/pull/73/commits/00d5099d482455172ce989ae2d21d2015a52d556.

gziolo (approved):

PR #56, opened 2025-09-03
Add execute actions to the ability object

gziolo:
Illustration for the ideas to make abilities more extensible:
- https://github.com/WordPress/abilities-api/issues/39

This PR adds extensibility hooks to the ability execution flow by introducing two action hooks that fire before and after ability execution. This allows developers to hook into the ability lifecycle for monitoring, logging, or extending functionality.

- Added `before_execute_ability` action hook that fires before ability execution
- Added `after_execute_ability` action hook that fires after successful ability execution
- Refactored the return logic to properly handle validation errors and execute the post-execution hook only on success. It is up for debate what to do on error.

## Example Usage

**Basic Logging**

```php
// Log all ability executions.
add_action( 'before_execute_ability', function( $ability_name, $input ) {
    error_log( 'Executing ability: ' . $ability_name );
    if ( $input !== null ) {
        error_log( 'Input: ' . wp_json_encode( $input ) );
    }
}, 10, 2 );

add_action( 'after_execute_ability', function( $ability_name, $input, $result ) {
    error_log( 'Completed ability: ' . $ability_name );
    error_log( 'Result: ' . wp_json_encode( $result ) );
}, 10, 3 );
```

PR #74, opened 2025-09-13
dev: add `register_ability_args` filter [Proposal]

justlevine:
## What

This PR exposes the  `register_ability_args` filter as a way to centrally extend and change the behavior of registered abilities.

## Why
- Part of #39
- This is a counterproposal to #37 :

   - Placing the filter in `WP_Abilities_Registry::register()` makes behavior robust against `WP_Ability` class overloads and keeps the lifecycle + validation intact.  In other words: filtering abilities remains idempotent to their non-filtered versions.
   - Smaller API footprint (one filter vs several) means smaller chance for future tech debt.
   - Arrays continue to be unbeaten when it comes to future backcompat.

## How
- ~`wp_ability_args` is used before the merge to core, but it should be changed to `abilities_api` as part of 6.9 IMO.~ changed to `register_ability_args`


docs/5.hooks.md
```diff
@@ -0,0 +1,75 @@
...
+  $args['has_permissions' ] = static function ( $input = null ) use ( $args ) {
+    $previous_check = is_callable( $args['has_permissions'] ) ? $args['has_permissions']( $input ) : true;
+
+    // If we already failed, no need for stricter checks.
+    if ( ! $previous_check || is_wp_error( $previous_check ) ) {
+      return $previous_check;
+    }
+
+    return current_user_can( 'my_custom_ability_cap', $args['name'] )
+  }
```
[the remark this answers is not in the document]
justlevine: I dont think this is true 🤔 we're only _calling_ inside the function, not reassigning.

docs/5.hooks.md
```diff
@@ -0,0 +1,75 @@
...
+  // Even if they're callbacks.
+  $args['has_permissions' ] = static function ( $input = null ) use ( $args ) {
+    $previous_check = is_callable( $args['has_permissions'] ) ? $args['has_permissions']( $input ) : true;
+
+    // If we already failed, no need for stricter checks.
+    if ( ! $previous_check || is_wp_error( $previous_check ) ) {
+      return $previous_check;
+    }
+
+    return current_user_can( 'my_custom_ability_cap', $args['name'] )
```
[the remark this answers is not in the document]
justlevine: Resolved in https://github.com/WordPress/abilities-api/pull/74/commits/dc10954dbf8571f508bde043d5a98caca67f6e42 when fixing the `$ability_name` reference.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -64,6 +64,16 @@ final class WP_Abilities_Registry {
...
 	public function register( string $name, array $args ): ?WP_Ability {
+		/**
+		 * Filters the ability arguments before they are validated.
+		 *
+		 * @since n.e.x.t
+		 *
+		 * @param array<string,mixed> $args The arguments used to instantiate the ability.
+		 * @param string              $name The name of the ability, with its namespace.
+		 */
+		$args = apply_filters( 'wp_ability_args', $args, $name );
```
gziolo: Minor note, this could get moved after all checks against `$name` to ensure it doesn't get processed on abilities that have an incorrect name.

docs/5.hooks.md
```diff
@@ -0,0 +1,75 @@
...
+
+There are currently **no Action hooks** provided by this version of the Abilities API.
+
+## Filters
+
+### `wp_ability_args`
+
+> [!IMPORTANT]
+> This filter is prefixed with `wp_` to avoid potential naming collisions.
+> Once merged into WordPress core, the prefix will likely be removed to preserve backward-compatibility.
```
gziolo: I would go with the final name from the start to avoid the hassle. `register_ability_args` should be distinct enough.

docs/5.hooks.md
```diff
@@ -0,0 +1,75 @@
...
+  // You can use the old args to build new ones.
+  $args['description'] = sprintf(
+    /* translators: 1: Ability name 2: Previous description */
+    __('This is a custom description for the ability %s. Previously the description was %s', 'text-domain'),
+    $ability_name,
+    $args['description'] ?? 'N/A'
+  );
+
+  // Even if they're callbacks.
+  $args['has_permissions' ] = static function ( $input = null ) use ( $args, $ability_name ) {
```
gziolo: This should be `permission_callback` rather than `has_permissions`.

gziolo (commented):
I left some minor feedback. Overall, I'm very much in favor of the proposed extensibility option.

docs/5.hooks.md
justlevine: `register_ability_args` is better than `ability_args` I had in mind too as it aligns with `register_{post_type|taxonomy}` args.

includes/abilities-api/class-wp-abilities-registry.php
justlevine: My intent here is intentionally the opposite, since it allows changes/recoverability of the `$name` while still ensuring it passes the necessary validation for downstream checks.

IMO (pseudocode):
```php
add_filter( 'register_ability_args',
  // Or whatever prefixes/missing-namespaces are enforced by validation.
  static fn ( &$args, $original_name ) => $args['name'] = sprintf( '!$#@#_%s', str_replace( '/', '_', $original_name ),
10, 2 );
```

should not be allowed since it means adapters/downstream cant reliably depend on the namespace/shape for their needs.

Meanwhile, I'm believe that e.g.

```
add_filter( 'register_ability_args',
  static function( $args,) {
    if ( str_starts_with( 'my_legacy_namespace/', $args['name' ) ) {
      $args['name'] = str_replace( 'my_legacy_namespace/', 'my_new_namespace/', $args['name'] );
    }
  }
);
```

should still be flexible enough for any downstream/back-compat considerations without making the expected shape of the Ability API unreliable, but I'm happy to evaluate other use cases (the only one I can think of is if a user intentionally registers a namespace _wrong_ initially, but the immediate error they face should prevent that from ever shipping and thus preempting the need for a downstream filter)

(PS: moving this to after the checks after a 6.9 merge is a _nonbreaking change_. moving it up to enforce name validation

docs/5.hooks.md
gziolo: I'm closely familiar with `register_block_type_args`, which was likely inspired by these you listed.

includes/abilities-api/class-wp-abilities-registry.php
gziolo: `$args` is not used to read the name. It's a separate variable `$name` which can't be modified by the hook.

justlevine: Indeed 🤦 too much task switching that even with the function in front of me I got our `$name` handling confused with `$ability_class`. Sorry for wasting your time 😬

I'll move it down to after the `->is_registered()` check. When that and the other comments are handled I'll ping you for re-review 🙇

gziolo: No worries, it's why additional eyes are helpful to the implementation from a different perspective 👍🏻

gziolo:
We will also need some basic unit test coverage to illustrate the expected behavior better and prevent future regressions. 

I plan to add some tests today to:
- https://github.com/WordPress/abilities-api/pull/56

We can coordinate how to ensure we avoid more complex merge conflicts.

gziolo:
> I plan to add some tests today to:
>
> https://github.com/WordPress/abilities-api/pull/56

For your information. Tests for hooks live in https://github.com/WordPress/abilities-api/blob/trunk/tests/unit/abilities-api/wpAbility.php. Docs for hooks live in https://github.com/WordPress/abilities-api/blob/trunk/docs/6.hooks.md.

gziolo:
@galatanovidiu, this proposed filter `register_ability_args` that allows changing `$args` during registration should be essential for extenders that want to annotate existing abilities for usage with the MCP adapter when such information is missing.

docs/5.hooks.md
justlevine: https://github.com/WordPress/abilities-api/pull/74/commits/93439d8ea07af01752def1295dde856587632e03

includes/abilities-api/class-wp-abilities-registry.php
justlevine: https://github.com/WordPress/abilities-api/pull/74/commits/93439d8ea07af01752def1295dde856587632e03

justlevine:
@gziolo feedback addressed, tests added, and merged the action/filter docs (structure too, not just pasting the two together). PTAL and let me know what you think 🙏

docs/6.hooks.md
```diff
@@ -1,68 +1,134 @@
...
+        $args['description'] ?? 'N/A'
+    );
+
+    // Even if they're callbacks.
+    $args['permission_callback' ] = static function ( $input = null ) use ( $args, $ability_name ) {
+        $previous_check = is_callable( $args['permission_callback'] ) ? $args['permission_callback']( $input ) : true;
+
+        // If we already failed, no need for stricter checks.
+        if ( ! $previous_check || is_wp_error( $previous_check ) ) {
+        return $previous_check;
```
gziolo: Nitpick, spacing:


```suggestion
            return $previous_check;
```

tests/unit/abilities-api/wpAbilitiesRegistry.php
```diff
@@ -427,4 +427,115 @@ public function test_wp_ability_invalid_properties_throws_exception() {
...
+				'a' => 1,
+				'b' => 2,
+			)
+		);
+
+		$this->assertTrue( $was_filter_callback_fired, 'The execute callback defined in the filter was not fired.' );
+		$this->assertSame( 3, $result, 'The original execute callback did not return the expected result.' );
+
+		// Clean up by removing the filter.
+		remove_filter( 'register_ability_args', $filter_callback, 10 );
```
gziolo: Nitpick, here in other tests, it'd be safer to follow WP core and move test cleanup before assertion to ensure they always execute even when the assertion fails.

tests/unit/abilities-api/wpAbilitiesRegistry.php
```diff
@@ -427,4 +427,115 @@ public function test_wp_ability_invalid_properties_throws_exception() {
...
+		$filter_callback = static function ( $args, $name ) {
+			if ( self::$test_ability_name !== $name ) {
+				// Do not modify args for other abilities.
+				return $args;
+			}
+
+			$args['label'] = 'Modified label for specific ability';
+			return $args;
+		};
+		add_filter( 'register_ability_args', $filter_callback, 10, 2 );
```
gziolo: The filter must be removed so it doesn't inadvertently affect other tests.

gziolo (approved):
Nice addition to the documentation and test coverage. Thank you for addressing my feedback.

I left some small notes to address before merging changes 🚀

tests/unit/abilities-api/wpAbilitiesRegistry.php
justlevine: Thanks for catching this, will handle via https://github.com/WordPress/abilities-api/pull/74#discussion_r2387168887

tests/unit/abilities-api/wpAbilitiesRegistry.php
justlevine: https://github.com/WordPress/abilities-api/pull/74/commits/36c0e992a259af669eb8bd717985a552de3cb04f

Issue #75, opened 2025-09-13
REST endpoints should themselves be Abilities


justlevine:
Seems counterintuitive that we're creating traditional REST endpoints, when REST endpoints themselves are just shaped abilities.

Given that discovery is itself a conceptual Ability - and a useful one e.g. in the context of MCP Adapter and limiting tool numbers - I propose we dogfood our own APIs, and reimplement these endpoints as Abilities themselves that are just "adapted to" (i.e. wrapped in) `\WP_REST_Controller`.

Happy to create a POC if nobody beats me to it; IRL this should probably wait until after WordPress/ai#40 or be part of that spec.

gziolo:
The repeating feedback I noticed so far is the opposite – how to also register an existing custom REST API endpoint as an ability or set of abilities. I'm noting that mostly to acknowledge the current reasoning of developers. I'm definitely in favor of exploring how we would build guidelines around when to avoid creating more custom REST API endpoints in favor of abilities. More importantly, for existing projects, there will be a need for a set of recommendations on how to turn them into abilities. I'm curious to see your POC to understand your current thinking on the topic better.

justlevine:
> The repeating feedback I noticed so far is the opposite – how to also register an existing custom REST API endpoint as an ability or set of abilities.

To clarify - I don't think these needs are exclusive. Users with existing REST endpoints who want the quickest migratory path to have them consumable as abilities shouldn't conflict with the long-term desire for the more robust "register once, consume anywhere" pattern of using the Ability as the source of truth for all last-mile adaptations.

Sent you a DM in the meantime but hopefully the POC will illustrate this better 🙇

PR #76, opened 2025-09-14
Fix has_permission() return type inconsistency (Issue #67)

### LINKED ISSUE #67: Rename usage of `has_permission()` to account for a `WP_Error`
The following pages document how to use return value of the `has_permission()` method:

* https://github.com/WordPress/abilities-api/blob/v0.1.1/docs/2.getting-started.md
* https://github.com/WordPress/abilities-api/blob/v0.1.1/docs/4.using-abilities.md

This method can return a `WP_Error` or a boolean, so a check like `if ( $ability->has_permission( $input ) )` does not do what is expected. A `WP_Error` instance will evaluate to true and cause this logical check to unexpectedly pass.

The `WP_Ability::execute()` method correctly handles a `WP_Error` being returned by the permissions check.

There are a few options.

* Adjust `has_permission()` to only return a boolean, so it behaves more like a traditional capability check function such as `current_user_can()`.
  * Can be combined with retaining the existing method under a different name.
  * Can be combined with throwing an exception. Not very WordPress-y.
* Retain everything as-is but update all the developer documentation to include checks for `is_wp_error()`. I think this is risky, it's too easy for a developer to implement a simple boolean check and therefore incorrectly handle a `WP_Error`.

Ref34t:
## Summary
Simplifies permission checking API and updates documentation to show proper `is_wp_error()` usage patterns for preventing WP_Error truthiness issues.

## Problem Solved
The original `has_permission()` method could return either `bool` or `WP_Error`, creating a dangerous bug where WP_Error objects (which are truthy) could incorrectly pass permission checks in if statements.

## Implementation
- `check_permissions($input, $skip_validation = false)`: New recommended method with clear `bool|WP_Error` return type
- `has_permission($input)`: Deprecated wrapper for backward compatibility
- Updated documentation: Shows proper `is_wp_error()` usage patterns throughout
- Streamlined API: Uses `$skip_validation` parameter for REST API permission checks

## Testing
Existing comprehensive tests verify both methods work correctly:
- Permission validation with proper error handling
- Input validation integration with permission checks
- Backward compatibility for deprecated `has_permission()` method

Fixes #67

### DIFF
--- a/docs/2.getting-started.md
+++ b/docs/2.getting-started.md
@@ -133,10 +133,27 @@ add_action( 'admin_init', 'my_plugin_use_ability' );
 function my_plugin_use_ability() {
     $ability = wp_get_ability( 'my-plugin/get-site-title' );
 
-    if ( $ability && $ability->has_permission() ) {
-        $site_title = $ability->execute();
-        // $site_title now holds the result of get_bloginfo('name')
-        // error_log( 'Site Title: ' . $site_title );
+    if ( $ability ) {
+        // Check permissions first - always use is_wp_error() to handle errors properly
+        $permission = $ability->check_permission();
+        if ( is_wp_error( $permission ) ) {
+            // Handle permission error
+            error_log( 'Permission error: ' . $permission->get_error_message() );
+            return;
+        } elseif ( $permission ) {
+            // Permission granted - safe to execute
+            $site_title = $ability->execute();
+            if ( is_wp_error( $site_title ) ) {
+                // Handle execution error
+                error_log( 'Execution error: ' . $site_title->get_error_message() );
+            } else {
+                // $site_title now holds the result of get_bloginfo('name')
+                // error_log( 'Site Title: ' . $site_title );
+            }
+        } else {
+            // Permission denied
+            error_log( 'Permission denied for ability execution' );
+        }
     }
 }
 ```
--- a/docs/4.using-abilities.md
+++ b/docs/4.using-abilities.md
@@ -124,9 +124,14 @@ if ( $ability ) {
 }
 ```
 
-**Checking Permissions (`$ability->has_permission()`)**
+**Checking Permissions (`$ability->check_permission()`)**
 
-Before executing an ability, you can check if the current user has permission:
+Before executing an ability, you can check if the current user has permission. The `check_permission()` method returns either `true`, `false`, or a `WP_Error` object, so you must use `is_wp_error()` to handle errors properly:
+
+```php
+// Method signature:
+// check_permission( $input = null )
+```
 
 ```php
 $ability = wp_get_ability( 'my-plugin/update-option' );
@@ -136,12 +141,17 @@ if ( $ability ) {
         'option_value' => 'New Site Name',
     );
 
-    // Check permission before execution
-    if ( $ability->has_permission( $input ) ) {
+    // Check permission before execution - always use is_wp_error() first
+    $permission = $ability->check_permission( $input );
+    if ( is_wp_error( $permission ) ) {
+        // Handle permission check error (validation, callback error, etc.)
+        echo 'Permission check failed: ' . $permission->get_error_message();
+    } elseif ( $permission ) {
+        // Permission granted - safe to execute
         $result = $ability->execute( $input );
         if ( is_wp_error( $result ) ) {
-	        // Handle WP_Error
-	        echo 'Error: ' . $result->get_error_message();
+	        // Handle execution error
+	        echo 'Execution error: ' . $result->get_error_message();
         } else {
         	// Use $result
 			if ( $result['success'] ) {
@@ -150,6 +160,7 @@ if ( $ability ) {
 			}
         }
     } else {
+        // Permission denied
         echo 'You do not have permission to execute this ability.';
     }
 }
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -412,7 +412,7 @@ class WP_Ability {
 	 * @return mixed|\WP_Error The result of the ability execution, or WP_Error on failure.
 	 */
 	public function execute( $input = null ) {
-		$has_permissions = $this->has_permission( $input );
+		$has_permissions = $this->check_permission( $input );
 		if ( true !== $has_permissions ) {
 			if ( is_wp_error( $has_permissions ) ) {
 				if ( 'ability_invalid_input' === $has_permissions->get_error_code() ) {
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -312,12 +312,12 @@ class WP_Ability {
 	 *
 	 * The input is validated against the input schema before it is passed to to permission callback.
 	 *
-	 * @since 0.1.0
+	 * @since N.E.X.T
 	 *
 	 * @param mixed $input Optional. The input data for permission checking. Default `null`.
 	 * @return bool|\WP_Error Whether the ability has the necessary permission.
 	 */
-	public function has_permission( $input = null ) {
+	public function check_permission( $input = null ) {
 		$is_valid = $this->validate_input( $input );
 		if ( is_wp_error( $is_valid ) ) {
 			return $is_valid;
@@ -328,6 +328,24 @@ class WP_Ability {
 		}
 
 		return call_user_func( $this->permission_callback, $input );
+	}
+
+	/**
+	 * Checks whether the ability has the necessary permissions (deprecated).
+	 *
+	 * The input is validated against the input schema before it is passed to to permission callback.
+	 *
+	 * @deprecated N.E.X.T Use check_permission() instead.
+	 * @see WP_Ability::check_permission()
+	 *
+	 * @since 0.1.0
+	 *
+	 * @param mixed $input Optional. The input data for permission checking. Default `null`.
+	 * @return bool|\WP_Error Whether the ability has the necessary permission.
+	 */
+	public function has_permission( $input = null ) {
+		_deprecated_function( __METHOD__, 'N.E.X.T', 'WP_Ability::check_permission()' );
+		return $this->check_permission( $input );
 	}
 
 	/**
--- a/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
+++ b/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
@@ -163,7 +163,7 @@ class WP_REST_Abilities_Run_Controller extends WP_REST_Controller {
 		}
 
 		$input = $this->get_input_from_request( $request );
-		if ( ! $ability->has_permission( $input ) ) {
+		if ( ! $ability->check_permission( $input ) ) {
 			return new \WP_Error(
 				'rest_ability_cannot_execute',
 				__( 'Sorry, you are not allowed to execute this ability.' ),
--- a/tests/unit/abilities-api/wpRegisterAbility.php
+++ b/tests/unit/abilities-api/wpRegisterAbility.php
@@ -134,7 +134,7 @@ class Test_Abilities_API_WpRegisterAbility extends WP_UnitTestCase {
 		$this->assertSame( self::$test_ability_args['output_schema'], $result->get_output_schema() );
 		$this->assertSame( self::$test_ability_args['meta'], $result->get_meta() );
 		$this->assertTrue(
-			$result->has_permission(
+			$result->check_permission(
 				array(
 					'a' => 2,
 					'b' => 3,
@@ -289,7 +289,7 @@ class Test_Abilities_API_WpRegisterAbility extends WP_UnitTestCase {
 
 		$result = wp_register_ability( self::$test_ability_name, self::$test_ability_args );
 
-		$actual = $result->has_permission(
+		$actual = $result->check_permission(
 			array(
 				'a'       => 2,
 				'b'       => 3,
@@ -305,6 +305,27 @@ class Test_Abilities_API_WpRegisterAbility extends WP_UnitTestCase {
 		$this->assertSame(
 			'Ability "test/add-numbers" has invalid input. Reason: unknown is not a valid property of Object.',
 			$actual->get_error_message()
+		);
+	}
+
+	/**
+	 * Tests that the deprecated has_permission() method still works and shows deprecation.
+	 *
+	 * @expectedDeprecated WP_Ability::has_permission
+	 */
+	public function test_has_permission_deprecated_method(): void {
+		do_action( 'abilities_api_init' );
+
+		$result = wp_register_ability( self::$test_ability_name, self::$test_ability_args );
+
+		// Test that deprecated method still works
+		$this->assertTrue(
+			$result->has_permission(
+				array(
+					'a' => 2,
+					'b' => 3,
+				)
+			)
 		);
 	}
 
@@ -325,7 +346,7 @@ class Test_Abilities_API_WpRegisterAbility extends WP_UnitTestCase {
 
 		// Test with a > b (should be allowed)
 		$this->assertTrue(
-			$result->has_permission(
+			$result->check_permission(
 				array(
 					'a' => 5,
 					'b' => 3,
@@ -342,7 +363,7 @@ class Test_Abilities_API_WpRegisterAbility extends WP_UnitTestCase {
 
 		// Test with a < b (should be denied)
 		$this->assertFalse(
-			$result->has_permission(
+			$result->check_permission(
 				array(
 					'a' => 2,
 					'b' => 8,
--- a/tests/unit/abilities-api/wpRegisterAbility.php
+++ b/tests/unit/abilities-api/wpRegisterAbility.php
@@ -309,27 +309,6 @@ class Test_Abilities_API_WpRegisterAbility extends WP_UnitTestCase {
 	}
 
 	/**
-	 * Tests that the deprecated has_permission() method still works and shows deprecation.
-	 *
-	 * @expectedDeprecated WP_Ability::has_permission
-	 */
-	public function test_has_permission_deprecated_method(): void {
-		do_action( 'abilities_api_init' );
-
-		$result = wp_register_ability( self::$test_ability_name, self::$test_ability_args );
-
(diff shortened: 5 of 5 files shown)


justlevine:
(@Ref34t keep #68 in mind (and #73) before you waste too much time because of failing tests - I'll give a real review one you tag ppl, just saw the rest errors and guessing there might be an unnecessary headache you dont need to care about)

Ref34t:
@justlevine, thanks for keeping me aware of what is happening. You can review the fix itself now

includes/abilities-api/class-wp-ability.php
```diff
@@ -331,6 +331,21 @@ public function has_permission( $input = null ) {
...
+	 * Checks whether the ability has the necessary permissions.
+	 * If the permission callback is not set, the default behavior is to allow access
+	 * when the input provided passes validation.
+	 *
+	 * @since N.E.X.T
+	 *
+	 * @param mixed $input Optional. The input data for permission checking. Default `null`.
+	 * @return bool Whether the ability has the necessary permission.
+	 */
+	public function has_permission( $input = null ): bool {
```
justlevine: Since the other method is public, having _another_ public method that's sole goal is coercing things to a boolean seems superfluous, and unnecessary tech-debt for an initial WP6.9 release. 

Resurfaced this back on the issue at https://github.com/WordPress/abilities-api/issues/67#issuecomment-3289608930

includes/abilities-api/class-wp-ability.php
```diff
@@ -395,7 +410,7 @@ protected function validate_output( $output ) {
 	 * @return mixed|\WP_Error The result of the ability execution, or WP_Error on failure.
 	 */
 	public function execute( $input = null ) {
-		$has_permissions = $this->has_permission( $input );
+		$has_permissions = $this->get_permission_status( $input );
```
justlevine: Case in point to https://github.com/WordPress/abilities-api/pull/76#discussion_r2347378304 - we're only using this internally while our only use of `::has_permission()` is in a REST API endpoint (which long term [should probably be an ability anyway](https://github.com/WordPress/abilities-api/issues/75)

includes/abilities-api/class-wp-ability.php
```diff
@@ -305,16 +305,16 @@ protected function validate_input( $input = null ) {
...
 	 *
-	 * @since 0.1.0
+	 * @since N.E.X.T
 	 *
 	 * @param mixed $input Optional. The input data for permission checking. Default `null`.
-	 * @return bool|\WP_Error Whether the ability has the necessary permission.
+	 * @return bool|\WP_Error Whether the ability has the necessary permission, or WP_Error with details.
 	 */
-	public function has_permission( $input = null ) {
+	public function get_permission_status( $input = null ) {
```
justlevine: Permission "status" isn't a concept that currently exists in the WordPress core codebase, so I would avoid creating it. Some alternatives:

- `::check_permissions()` - completely unused, but it:
   - follows good semantic naming for an action
	 - `::check_*` is _mostly_ used to return `{some_type}|\WP_Error`, but there are some instances where its `: void` or `: bool`
	 - Specifically, `::check_{someitem}_permissions()` is common on REST and returns `bool`.
- `::permissions_check()`,  `::get_permissions_check()` or the more verbose `::get_ability_permissions_check()`
   - Mirrors what's used by `WP_Rest_Controller` - because it has multiple perms of checks in a class (CRUD) and `WP_oEmbed_Controller::get_proxy_item_permissions_check()` assumedly because it's a rest endpoint even though it doesn't inherit the class.
   - `::permissions_check()` is used in several `WP_Rest_Controller` children instead of `get_*()` , and this is the one I recommend here. (There are a couple uses of `::do_permissions_check()` but afaik they're all only `protected` and leaked in on much larger PRs where the specific rest endpoint wasnt necessarily code reviewed in detail since there wasn't an expectation of it being extended).
   
**PS: note the use of permission*s* (plural). I couldn't find anywhere in core that uses the singular `permission` in a method name.

justlevine (commented):
Aside from [whether or not](https://github.com/WordPress/abilities-api/issues/67#issuecomment-3289608930) we need _both_ a public checker function and a (also public) wrapper whose sole purpose is to coerce it to a `bool`, I think we need a better name for the function itself.

image

includes/abilities-api/class-wp-ability.php
justlevine: >  while our only use of ::has_permission() is in a REST API endpoint

Follow up: 👆(and not #73) is the reason why the tests are failing. `WP_REST_Abilities_Run_Controller::run_ability_permissions_check()` explicitly only checks against `false` and specifically allows the WP_Error through so invalid method errors are only shown to a user with the correct permissions when the callback is actually run:

https://github.com/ref34t/abilities-api/blob/8fcfcac9136dee73abb798bbaded0a21c17b5cd8/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php#L93-L111

@Ref34t In the context of _this_ PR, change `::run_ability_permissions_check()` to `if ( false === $ability->get_permission_status()` (or whatever it's renamed to).

---
More broadly, I'm like 95% sure this leaked in from the A8C prior art and should be **removed and punted to MCP Adapter**, as "Resource" and "Tool" are not concepts of the Abilities API and don't belong in core. @emdashcodes / @gziolo please confirm, but either way it can be handled as part of #75.

Ref34t:
@justlevine Thanks for the feedback! Based on your comments and WP Core research, I’ve updated the PR to:
	•	Rename has_permission() → check_permissions() to avoid the misleading has_* boolean expectation.
	•	Standardize return types to true | false | WP_Error, matching Core patterns (like permissions_check()) and preventing WP_Error from being treated as true.
	•	Update execute() to always enforce permissions safely, centralizing logic and ensuring abilities can’t be bypassed.
	•	Align with Core standards while keeping developer usage simple and predictable.

These changes address #67, improve security, and follow Core conventions for REST and capability checks.

gziolo:
You are exploring more advanced refactorings rather than a simple documentation update to account for possible incorrect usage. I left some other ideas to consider in https://github.com/WordPress/abilities-api/issues/67#issuecomment-3290617975 and https://github.com/WordPress/abilities-api/issues/67#issuecomment-3291135690.

includes/abilities-api/class-wp-ability.php
```diff
@@ -305,19 +306,22 @@ protected function validate_input( $input = null ) {
...
 	 * @param mixed $input Optional. The input data for permission checking. Default `null`.
-	 * @return bool|\WP_Error Whether the ability has the necessary permission.
+	 * @param bool  $skip_validation Optional. Whether to skip input validation. Default `false`.
+	 * @return bool|\WP_Error Whether the ability has the necessary permission, or WP_Error with details.
 	 */
-	public function has_permission( $input = null ) {
-		$is_valid = $this->validate_input( $input );
-		if ( is_wp_error( $is_valid ) ) {
-			return $is_valid;
+	public function check_permissions( $input = null, $skip_validation = false ) {
```
gziolo: Could you elaborate on the reasoning for this change? It doesn't look related to the issue reported.

> Streamlined API: Uses $skip_validation parameter for REST API permission checks

This comment seems related, but it doesn't explain why to sacrifice the simplicity of the method to simplify implementation for a specific usage in the REST controller.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
justlevine, line 1: Only reviewing this on my phone, but want to make sure we're limiting the refactoring in this PR to specifically address in #67. We can do other stuff in other PRs that isn't blocking v0.2.0.

Will pull for a real review hopefully this evening.

gziolo (changes requested):
I wanted to echo the feedback shared by @justlevine in https://github.com/WordPress/abilities-api/pull/76#discussion_r2355503610. Can we limit this PR to renaming `has_permission()` to `check_permission()` with the included deprecation logic and documentation changes?

That part is ready to go.

includes/abilities-api/class-wp-ability.php
gziolo: > More broadly, I'm like 95% sure this leaked in from the A8C prior art and should be removed and punted to MCP Adapter, as "Resource" and "Tool" are not concepts of the Abilities API and don't belong in core. @emdashcodes / @gziolo please confirm, but either way it can be handled as part of https://github.com/WordPress/abilities-api/issues/75.

I noticed this when reviewing the PR, and I found it useful. I’m not 100% sure we should use this specific set of criteria to distinguish between GET and POST, but I didn’t have the bandwidth to start a larger conversation about the optimal approach as this seems more like an implementation detail. I think we could tap into [Add property for marking Ability hints like destructive, read-only, idempotent](https://github.com/WordPress/abilities-api/issues/62) to iron out the detection mechanism.

Overall, I’m tempted to borrow some of the solutions from MCP without using nomenclature and structure too tied this particular protocol.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
Ref34t: resolved

gziolo:
@Ref34t, I see you are actively working on addressing feedback. I triggered review from GitHub Copilot to see if it catches anything. I'm still not sure whether `check_permission` or `check_permissions` would be a better choice. The part that triggered my thinking was documentation which uses plural everywhere.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
gziolo: nice!

Ref34t:
@gziolo I had it before plural and now due to your last feedback I changed it again 😔 what do you think the final call here? 🤔

gziolo:
The REST API uses `permission_callback`, so let's stick with the singular form, as it is also used for the ability. It looks like no further changes are necessary because in the docs I see the plural form when explaining the concept for REST API:

https://developer.wordpress.org/rest-api/extending-the-rest-api/adding-custom-endpoints/#permissions-callback

tests/unit/abilities-api/wpRegisterAbility.php
```diff
@@ -308,6 +308,32 @@ public function test_permission_callback_no_input_schema_match(): void {
...
 
+	/**
+	 * Tests that deprecated has_permission() method still works
+	 */
+	public function test_has_permission_deprecated_coverage(): void {
+		do_action( 'abilities_api_init' );
+
+		$result = wp_register_ability( self::$test_ability_name, self::$test_ability_args );
+
+		// Suppress deprecation notice for coverage testing
```
gziolo: I'm curious whether the standard WP test helper is avialable here. Something like that would work in core:
```php
$this->expectedDeprecated( 'WP_Ability::has_permission' );
```

gziolo: Maybe it's done with `@expectedDeprecated` and PHPDoc. It might be fine as is given the limited testing context.

Ref34t: yes, after checking that's right. Resolved here https://github.com/WordPress/abilities-api/pull/76/commits/d55a0bb4d2cde3fc3e965193472ae0843a015c07

gziolo (approved):
All feedback addressed. It looks good to go from my angle. @justlevine how about you?

@jonathanbossenger, could you perform a quick sanity check on the docs changes?

justlevine:
> The REST API uses permission_callback, so let's stick with the singular form, as it is also used for the ability.

I'm sorry to be pedantic, but as [I previously noted](https://github.com/WordPress/abilities-api/pull/76#discussion_r2347392156) the REST API uses `check_{*_}_permissions()` _plural_, even though the registry arg is `permission_callback`. It's also more grammatically accurate when paired with the verb `check_*`.

I'm happy to do this myself in a follow-up PR if we feel bad about making @Ref34t keep swapping back-and-forth, but I do feel strongly about uniformity with core APIs.

includes/abilities-api/class-wp-ability.php
justlevine: > Overall, I’m tempted to borrow some of the solutions from MCP without using nomenclature and structure too tied this particular protocol.

Yeah it's that latter part that's critical, and considering that [MCP's value as a whole is doubtful](https://wordpress.slack.com/archives/C08TJ8BPULS/p1758990133058409), I would want a discussion of what a "resource" or "tool" is meant to be in our context, and whether it is inherently useful - if/when this gets revisited.

Either way, no longer part of this PR, so can be resolved.

docs/2.getting-started.md
```diff
@@ -133,10 +133,27 @@ add_action( 'admin_init', 'my_plugin_use_ability' );
 function my_plugin_use_ability() {
     $ability = wp_get_ability( 'my-plugin/get-site-title' );
 
-    if ( $ability && $ability->has_permission() ) {
-        $site_title = $ability->execute();
-        // $site_title now holds the result of get_bloginfo('name')
-        // error_log( 'Site Title: ' . $site_title );
+    if ( $ability ) {
+        // Check permissions first - always use is_wp_error() to handle errors properly
+        $permission = $ability->check_permission();
```
justlevine, line 138: I think this (and the previous example) is incorrect. By calling `$ability->execute()`, `$ability->get_permission()` gets called, there's rarely a reason to explicitly check permissions as a separate step.

```php
if ( $ability  ) {
        $site_title = $ability->execute();
        // $site_title now holds the result of get_bloginfo('name')
        // error_log( 'Site Title: ' . $site_title );
    }
}

File: docs/4.using-abilities.md [hunk 7]
```diff
@@ -124,9 +124,14 @@ if ( $ability ) {
...
-**Checking Permissions (`$ability->has_permission()`)**
+**Checking Permissions (`$ability->check_permission()`)**
 
-Before executing an ability, you can check if the current user has permission:
+Before executing an ability, you can check if the current user has permission. The `check_permission()` method returns either `true`, `false`, or a `WP_Error` object, so you must use `is_wp_error()` to handle errors properly:
+
+```php
+// Method signature:
+// check_permission( $input = null )
+```
```
justlevine, line 134: Is this intentional or did it leak?

docs/4.using-abilities.md
```diff
@@ -136,12 +141,17 @@ if ( $ability ) {
         'option_value' => 'New Site Name',
     );
 
-    // Check permission before execution
-    if ( $ability->has_permission( $input ) ) {
+    // Check permission before execution - always use is_wp_error() first
```
justlevine, line 144: I know the doc in general uses a lot of unnecessary nesting, but I think this also make's it clearer than sandwiching the happy-path between two error states.

```php
    // Use a strict check to catch both false and WP_Error
    $permission = $ability->check_permission( $input );
    if ( true === $permission ) {
        // Permission granted - safe to execute.
        $result = $ability->execute( $input );
        if ( is_wp_error( $result ) ) {
            // Handle execution error
            echo 'Execution error: ' . $result->get_error_message();
        } else {
            // Use $result
            if ( $result['success'] ) {
                echo 'Option updated successfully!';
                echo 'Previous value: ' . $result['previous_value'];
            }
        }
    } else {
        // Don't leak permission errors to unauthenticated users.
        if ( is_wp_error( $permission ) ) {
            error_log( 'Permission check failed: ' . $permission->get_error_message() );
        }

        echo 'You do not have permission to execute this ability.';
    }
}
```

Though imo we shouldn't be actively promoting `check_permission()` and instead have users fully handle `execute()`

```php
$ability = wp_get_ability( 'my/ability' );

$result = $ability->execute( $input ) ;
// If we want to only handle perm errors and not also invalid inputs etc.
if ( is_wp_error() && 'ability_invalid_permissions' === $result->get_error_code() ) {
 // handle the error
}

File: includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php [hunk 9]
```diff
@@ -163,7 +163,7 @@ public function run_ability_permissions_check( $request ) {
 		}
 
 		$input = $this->get_input_from_request( $request );
-		if ( ! $ability->has_permission( $input ) ) {
+		if ( ! $ability->check_permission( $input ) ) {
```
justlevine, line 166: Are we intentionally only handling false and saving the WP_Errors for `->execute()`? cc @emdashcodes 

```suggestion
		if ( true !== $ability->check_permission( $input ) ) {
```

justlevine (commented):
Looks good! Left some comments that have nothing to do with [naming things](https://github.com/WordPress/abilities-api/pull/76#issuecomment-3342052640).

gziolo:
> I'm sorry to be pedantic, but as https://github.com/WordPress/abilities-api/pull/76#discussion_r2347392156 the REST API uses check_{*_}_permissions() plural, even though the registry arg is permission_callback. It's also more grammatically accurate when paired with the verb check_*.

I'm not a native speaker, so I'm happy to follow your guidance here. I will land this PR and try address all the feedback for docs in another PR.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
gziolo: Let's fix it. I don't think it was intentional.

gziolo: In my testing, the proposed code changes fail the result of 4 tests in the run controller. We would have to better handle the reason the request errored so it doesn't default to 403. It probably means, we would have to move some logic to `run_ability_permissions_check`.

gziolo:
A follow-up https://github.com/WordPress/abilities-api/pull/94 is ready for review. It addresses all feedback from @justlevine, but https://github.com/WordPress/abilities-api/pull/76#discussion_r2384361124, which I think needs more thought.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
gziolo: An alternative would be to remove `run_ability_permissions_check()` (skip `permission_callback`) in the run controller, taking into account that `run_ability` (part of the `callback`) validates permissions anyway. It still would have to handle 403 when `check_permissions()` from the ability returns `false`.

Technically speaking, we are good. However, we can improve the implementation/documentation to avoid future misunderstandings.

PR #77, opened 2025-09-16
Docs: REST API Reference

### LINKED ISSUE #28: Document REST API usage
Document the available REST API endpoints and how to use them for registered abilities.

gziolo:
Closes #28.

Documents the available REST API endpoints and how to use them for registered abilities.

Covers three routes available:

- fetching all abilities
- fetching a single ability
- executing a single ability

This is a first pass generated and iterated with Claude Code.

### DIFF
--- a/docs/5.rest-api.md
+++ b/docs/5.rest-api.md
@@ -14,17 +14,19 @@ Abilities are represented in JSON with the following structure:
 {
   "name": "my-plugin/get-site-info",
   "label": "Get Site Information",
-  "description": "Retrieves basic information about the WordPress site",
-  "input_schema": {
-    "type": "object",
-    "properties": {},
-    "additionalProperties": false
-  },
+  "description": "Retrieves basic information about the WordPress site.",
   "output_schema": {
     "type": "object",
     "properties": {
-      "name": {"type": "string"},
-      "url": {"type": "string", "format": "uri"}
+      "name": {
+        "type": "string",
+        "description": "Site name"
+      },
+      "url": {
+        "type": "string",
+        "format": "uri",
+        "description": "Site URL"
+      }
     }
   },
   "meta": {}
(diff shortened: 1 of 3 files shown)


jonathanbossenger:
@gziolo quick question, from my local testing it appears that all Abilities API REST Endpoints require authentication, is this correct? Want to make sure of my facts before adding review comments.

gziolo:
> quick question, from my local testing it appears that all Abilities API REST Endpoints require authentication, is this correct? Want to make sure of my facts before adding review comments.

I had to double-check, but the answer is:
- to get all abilities or a single ability `current_user_can( 'read' )` check is performed, so the user needs to be authorized
- to run an ability, it all depends on the ability itself and what permission it defined

@johnbillion, any thoughts about the current approach to permissions when calling the REST API endpoints?

docs/5.rest-api.md
```diff
@@ -0,0 +1,196 @@
+# 5. REST API Reference
+
+The WordPress Abilities API provides REST endpoints that allow external systems to discover and execute abilities via HTTP requests.
+
+## Schema
+
+The Abilities API endpoints are available under the `/wp/v2/abilities` namespace.
+
```
jonathanbossenger, line 12: Let's add a short section here about the fact that all Abilities endpoints require an authenticated user (with a link to the Authentication section), and that access to execute specific abilities is tied to the Ability's capabilities_check.

jonathanbossenger (changes requested):
One minor suggestion, but otherwise it all looks good, and checks out.

docs/5.rest-api.md
jonathanbossenger: Suggested text

## User access

Access to all Abilities REST API endpoints requires an authenticated user (see the Authentication section). Access to execute individual Abilities is restricted based on the `permission_callback()` of the Ability.

gziolo: Added in [5276094](https://github.com/WordPress/abilities-api/pull/77/commits/5276094c7105ecd72e2b6a805da1c12a6155aff5).

jonathanbossenger: LGTM

jonathanbossenger (approved):

PR #79, opened 2025-09-17
fix: Wrong callback function for the Rest API endpoints.

### LINKED ISSUE #78: REST Abilities List Controller calling wrong callback for the schema.
There is a broken REST schema callback in the WP_REST_Abilities_List_Controller, as it is calling `get_public_item_schema` instead of `get_item_schema`,

levinbaria:
Fix the broken REST schema callback in the WP_REST_Abilities_List_Controller, as it is calling get_public_item_schema instead of get_item_schema.

Fixes https://github.com/WordPress/abilities-api/issues/78

### DIFF
--- a/includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
+++ b/includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php
@@ -60,7 +60,7 @@ class WP_REST_Abilities_List_Controller extends WP_REST_Controller {
 					'permission_callback' => array( $this, 'get_permissions_check' ),
 					'args'                => $this->get_collection_params(),
 				),
-				'schema' => array( $this, 'get_public_item_schema' ),
+				'schema' => array( $this, 'get_item_schema' ),
 			)
 		);
 
@@ -80,7 +80,7 @@ class WP_REST_Abilities_List_Controller extends WP_REST_Controller {
 					'callback'            => array( $this, 'get_item' ),
 					'permission_callback' => array( $this, 'get_permissions_check' ),
 				),
-				'schema' => array( $this, 'get_public_item_schema' ),
+				'schema' => array( $this, 'get_item_schema' ),
 			)
 		);
 	}


gziolo:
@emdashcodes, can you confirm it is a bug? `get_public_item_schema` is method on `WP_REST_Controller` and it calls internally `get_item_schema` so everything seems coorect. There is a unit test in place that passes with both implementations.

justlevine:
> get_public_item_schema is method on WP_REST_Controller and it calls internally get_item_schema so everything seems correct.

That (and the passing tests) is only because we're not _currently_ using `arg_options` anywhere.

If there isn't a specific reason to deviate, I'd recommend we follow core as proposed here(implement `get_item_schema`, _call_ `get_public_item_schema`) instead of turning this into an inconsistent implementation detail contributors/maintainers need to watch out for if/when the schema warrants changes.

(Code LGTM and I don't think this needs tests either, not "approving" so it doesn't get accidently merged before @emdashcodes can weigh in)

gziolo:
> If there isn't a specific reason to deviate, I'd recommend we follow core as proposed here(implement get_item_schema, call get_public_item_schema) instead of turning this into an inconsistent implementation detail contributors/maintainers need to watch out for if/when the schema warrants changes.

Let's keep things as they are 👍🏻 

@levinbaria, thank you for bringing this potential issue to our attention.

emdashcodes:
> @emdashcodes, can you confirm it is a bug? get_public_item_schema is method on WP_REST_Controller and it calls internally get_item_schema so everything seems coorect. 

I'm sorry for missing this one last week while I was AFK a few days.

From my understanding, this is how we want to do things. I based this on core which only passes `get_public_item_schema` to `schema`. https://github.com/search?q=repo%3AWordPress%2FWordPress%20get_public_item_schema&type=code. 

`get_item_schema` is only defined and called internally: https://github.com/search?q=repo%3AWordPress%2FWordPress+get_item_schema&type=code&p=3.

> If there isn't a specific reason to deviate, I'd recommend we follow core as proposed here(implement get_item_schema, call get_public_item_schema) instead of turning this into an inconsistent implementation detail contributors/maintainers need to watch out for if/when the schema warrants changes.

Yup, this is exactly how things are implemented right now. 

`get_item_schema` - https://github.com/WordPress/abilities-api/blob/2f3cda7cb61bc41ae898b1a91339b44eb0f8c3a5/includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php#L231

`get_public_item_schema` - https://github.com/WordPress/abilities-api/blob/2f3cda7cb61bc41ae898b1a91339b44eb0f8c3a5/includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php#L63

> (Code LGTM and I don't think this needs tests either, not "approving" so it doesn't get accidently merged before @emdashcodes can weigh in)

Just to clarify -- we **don't** want this change. `trunk` is already doing this the correct way.

PR #81, opened 2025-09-18
fix: use link_header instead of add_link for pagination responses

emdashcodes:
This PR uses `link_header` instead of `add_link` to add next/prev pagination links to keep the response as a pure array.

When using `add_link`, `_links` is added to the response body, converting it from an array to a JSON object like `{"0": {...}, "_links": {...}}`.

Note that this is a breaking fix from 0.1, though you would have only noticed if you set per_page to something low or had 50+ abilities. 

See also https://github.com/WordPress/abilities-api/pull/60#discussion_r2325014788.

## Testing Instructions

* Make sure that all unit tests pass
* Make a test GET request to `/wp/v2/abilities?per_page=1` and see that both `X-WP-Total` and `X-WP-TotalPages` are set
* Make a test GET request to `/wp/v2/abilities?per_page=1&page=2` and see that the response is an array of items


gziolo (approved):
Thank you very much for cherry-picking changes from #60 👍🏻

PR #82, opened 2025-09-22
dev: Bump the PHPStan testing to PHP 8.4

johnbillion:
PHP unit tests are run on PHP 8.4 so PHPStan should check in the context of 8.4 too.


JasonTheAdams (approved):
Good catch!

PR #87, opened 2025-09-26
Create reference docs for client-side API

### LINKED ISSUE #86: Docs: client-side reference for Abilities API
Now that we have a client-side package, we need reference docs for interacting with Abilities in JavaScript.

It should cover the basic client API.

- getAbilities() - Fetches all available abilities
- getAbility(name) - Retrieves a specific ability by name
- executeAbility(name, input) - Executes abilities

jonathanbossenger:
Creates Abilities client-side API reference docs
- Includes link to package readme for installing and usage instructions
- Lists all API functions with example code

Closes #86


jonathanbossenger:
Finally got round to testing example code
Added link back to package readme

docs/7.javascript-client.md
```diff
@@ -0,0 +1,186 @@
...
+
+**Returns:** `void`
+
+**Example:**
+
+```javascript
+// showNotification function
+const showNotification = (message) => {
+  new Notification(message);
+  return { success: true, displayed: message };
```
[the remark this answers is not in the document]
jonathanbossenger: I'm closing this out, as the ability callback that calls showNotification already performs these checks.

docs/7.javascript-client.md
```diff
@@ -0,0 +1,186 @@
...
+Returns an array of all registered abilities (both server-side and client-side).
+
+**Parameters:** None
+
+**Returns:** `Promise<Array>` - Array of ability objects
+
+**Example:**
+
+```javascript
+const abilities = await getAbilities();
```
gziolo, line 30: We will need some way to show where to import `getAbilities` from. I assume through imports, but this has implications of explaining the build step. However, once we move the code to Gutenberg, this all should work nicely.

```suggestion
import { getAbilities } from `@wordpress/abilities`;

const abilities = await getAbilities();
```

README.md
```diff
@@ -24,6 +24,7 @@
 - [Using Abilities](docs/4.using-abilities.md)
 - [REST API Reference](docs/5.rest-api.md)
 - [Hooks](docs/6.hooks.md)
+- [JavaScript/TypeScript Client](docs/7.javascript-client.md)
```
gziolo, line 27: I would consider moving this document before Hooks, which is more of an extender's area. This way, documents would build nicely on top of each other, creating nicely separated layers:
- server-side
- REST API
- client-side

docs/7.javascript-client.md
```diff
@@ -0,0 +1,186 @@
...
+}
+```
+
+### executeAbility(name, input)
+
+Executes an ability with the provided input data.
+
+**Parameters:**
+- `name` (string) - The ability name
+- `input` (object, optional) - Input data for the ability
```
gziolo:
```suggestion
- `input` (any, optional) - Input data for the ability
```

I need to revisit the client-side navigation because, through the REST API, we now also allow simple types. See:

https://github.com/WordPress/abilities-api/blob/4dc57b3b64d6ec00caccb4ecff33a91f4554be9b/includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php#L206-L210

Otherwise, let's keep `object` and update everything seperately.

docs/7.javascript-client.md
```diff
@@ -0,0 +1,186 @@
...
+```
+
+### unregisterAbility(name)
+
+Removes a previously registered client-side ability.
+
+**Parameters:**
+- `name` (string) - The ability name to unregister
+
+**Returns:** `boolean` - True if the ability was found and removed
```
gziolo: I don't think it's correct:

https://github.com/WordPress/abilities-api/blob/4dc57b3b64d6ec00caccb4ecff33a91f4554be9b/packages/client/src/api.ts#L77

Aside, on the server, both registration and deregistration return the ability object or null when something fails, so maybe in another PR we should mirror the behavior.

gziolo (commented):
I had some minor feedback. Thank you very much for putting it all together.

README.md
gziolo: This came up again here when reviewing another PR in https://github.com/WordPress/abilities-api/pull/102#discussion_r2425864864:

> we really need to remove these numbers, as I would put this one next to registering categories as it fits better there. It's another instance where this ordering causes trouble 😅

gziolo (approved):

README.md
jonathanbossenger: I've made a note of this, and I'll rework the ordering when I remove the numbering

docs/7.javascript-client.md
jonathanbossenger: I wasn't aware we allow simple types, so happy to change this now.

README.md
gziolo: Sounds good, thank you 👍🏻

PR #88, opened 2025-09-26
Introduce `invoke_callback` helper method in `WP_Ability` class

gziolo:
This PR introduces a new `invoke_callback` helper method in the `WP_Ability` class to standardize callback invocation logic. The method handles whether to pass input parameters based on the presence of an input schema, replacing duplicated logic in the permission and execution callback methods. Additional test coverage was added to ensure that functionality continues to work the same way.

You can run tests locally with:

```bash
npm run test:php
```

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -400,6 +400,24 @@ class WP_Ability {
 	}
 
 	/**
+	 * Invokes a callable, ensuring the input is passed through only if the input schema is defined.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param callable $callback The callable to invoke.
+	 * @param mixed    $input    Optional. The input data for the ability. Default `null`.
+	 * @return mixed The result of the callable execution.
+	 */
+	protected function invoke_callback( $callback, $input = null ) {
+		$args = array();
+		if ( ! empty( $this->get_input_schema() ) ) {
+			$args[] = $input;
+		}
+
+		return $callback( ...$args );
+	}
+
+	/**
 	 * Checks whether the ability has the necessary permissions.
 	 *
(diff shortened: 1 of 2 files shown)


includes/abilities-api/class-wp-ability.php
```diff
@@ -399,6 +399,24 @@ protected function validate_input( $input = null ) {
...
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param callable $callback The callable to invoke.
+	 * @param mixed    $input    Optional. The input data for the ability. Default `null`.
+	 * @return mixed The result of the callable execution.
+	 */
+	protected function invoke_callback( callable $callback, $input = null ) {
+		$args = array();
+		if ( ! empty( $this->get_input_schema() ) ) {
```
JasonTheAdams, line 395: While we're cleaning things up, what do you think about adding a `$this->has_input_schema()` method to make this sort of check more declarative?

JasonTheAdams (approved):
Nice little refactor! I made a minor suggestion, but it's small enough where I'll approve regardless of your decision.

includes/abilities-api/class-wp-ability.php
gziolo: Sounds like a good idea 👍

felixarntz (approved):
@gziolo Good idea, LGTM!

includes/abilities-api/class-wp-ability.php
gziolo: I will play with this, and the other method idea https://github.com/WordPress/abilities-api/pull/104#discussion_r2420627883 in a follow-up PR 😄

gziolo: Would it make sense to have a similar method for the output schema? Given all the other tasks necessary for WP 6.9, this one is nice to have we can introduce later.

PR #94, opened 2025-09-30
Rename `check_permission()` to `check_permissions()` on `WP_Ability`

### LINKED ISSUE #67: Rename usage of `has_permission()` to account for a `WP_Error`

gziolo:
Follow up for #76.
Closes https://github.com/WordPress/abilities-api/issues/67.

Addresses feedback from @justlevine:
- https://github.com/WordPress/abilities-api/pull/76/files#r2384356477
- https://github.com/WordPress/abilities-api/pull/76/files#r2384359503
- https://github.com/WordPress/abilities-api/pull/76/files#r2384360545
- https://github.com/WordPress/abilities-api/pull/76#issuecomment-3342052640

> I'm sorry to be pedantic, but as https://github.com/WordPress/abilities-api/pull/76#discussion_r2347392156 the REST API uses check_{*_}_permissions() plural, even though the registry arg is permission_callback. It's also more grammatically accurate when paired with the verb check_*.

### DIFF
--- a/docs/2.getting-started.md
+++ b/docs/2.getting-started.md
@@ -22,9 +22,9 @@ wp plugin install https://github.com/WordPress/abilities-api/releases/latest/dow
   "$schema": "https://schemas.wp.org/trunk/wp-env.json",
   // ... other config ...
   "plugins": [
-    "WordPress/abilities-api",
+    "WordPress/abilities-api"
     // ... other plugins ...
-  ],
+  ]
   // ... more config ...
 }
 ```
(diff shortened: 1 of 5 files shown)


justlevine (approved):
🤩

PR #95, opened 2025-09-30
docs: Improve permission checking documentation and examples

Ref34t:
## Summary
- Replace manual permission checking examples with recommended execute() pattern
- Improve documentation to align with architectural design from #76
- Update REST API controller permission check for consistency

### Documentation Improvements
- **docs/2.getting-started.md**: Replaced 22-line manual permission checking example with clean 8-line execute() pattern
- **docs/4.using-abilities.md**: Eliminated manual permission checking anti-patterns, added advanced error handling section showing how to handle specific error types

### REST API Controller Update
- **includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php**: Updated permission check to use strict comparison (`true !==`) for consistency with execute() pattern

## Discussion Point

As @gziolo noted, this isn't a security vulnerability since each ability's `execute()` method already performs comprehensive permission checking. The REST API controller creates a two-layer permission model:

1. **REST API layer**: Pre-check via `run_ability_permissions_check()`
2. **Ability execution layer**: Full check via `execute()`

The controller's permission callback is architecturally unique because it dynamically checks permissions for *any* ability. We could consider simplifying by removing the separate permission callback and relying entirely on `execute()`'s error handling.

Open to feedback on this approach.

## Test plan
- [ ] Verify documentation examples follow recommended execute() pattern
- [ ] Confirm no manual permission checking anti-patterns remain in docs
- [ ] Test REST API endpoint permission handling


gziolo:
> REST API security vulnerability

I don't think we are really experiencing a security vulnerability here, because every ability handles permissions check individually during execution as discussed in https://github.com/WordPress/abilities-api/pull/76#discussion_r2390340956. That being said, we can iterate on the handling because the run controller is unique, as its permissions check model differs for each ability.

gziolo:
This PR needs a rebase to resolve conflicts after https://github.com/WordPress/abilities-api/pull/94 landed. This would help to review the changes proposed in this PR.

Ref34t:
@gziolo on it

Ref34t:
@gziolo on it

Ref34t:
I've updated the PR title and description - you're right that this isn't a security vulnerability.

The reason: even if the REST API permission check had the WP_Error truthiness issue, `execute()` on line 136 would still block unauthorized access since it performs its own permission check internally.

This update improves code correctness by using strict comparison (`true !==`) to properly handle all return types from `check_permissions()`, but it's not fixing an actual security hole due to the two-layer permission model.

docs/4.using-abilities.md
```diff
@@ -1,8 +1,8 @@
-# 4. Using Abilities (`wp_get_ability`, `wp_get_abilities`)
+### 4. Using Abilities (`wp_get_ability`, `wp_get_abilities`)
 
 Once abilities are registered, they can be retrieved and executed using global functions from the Abilities API.
 
-## Getting a Specific Ability (`wp_get_ability`)
+**Getting a Specific Ability (`wp_get_ability`)**
```
justlevine: I've been known to misread diffs on mobile 🫣 but if all these reverted _to_ `**`, then it's a leak from #96 and should be changed back to semantic

Ref34t: resolved https://github.com/WordPress/abilities-api/pull/95/commits/0fd34481a42136bdb26ef9302e2d8ff6c831aadf

jonathanbossenger:
I'm in the process of updating documentation in preparation for the 6.9 merge request in #117, so once that's merged, I'll come back here to check what still needs to be included.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
```diff
@@ -163,7 +163,7 @@ public function run_ability_permissions_check( $request ) {
 		}
 
 		$input = $this->get_input_from_request( $request );
-		if ( ! $ability->check_permissions( $input ) ) {
+		if ( true !== $ability->check_permissions( $input ) ) {
```
felixarntz: Sharing relevant feedback that I initially mentioned in https://github.com/WordPress/wordpress-develop/pull/9410#discussion_r2440731279: I don't think we should replace a contextually more specific error (from the actual permission callback) with a generic "Sorry you can't do this" error.

docs/4.using-abilities.md
```diff
@@ -20,20 +20,20 @@ $site_info_ability = wp_get_ability( 'my-plugin/get-site-info' );
...
+		// Use result data
+		echo 'Site Name: ' . $result['name'];
 	}
 } else {
 	// Ability not found or not registered
 }
 ```
 
-## Getting All Registered Abilities (`wp_get_abilities`)
+## Getting All Registered Abilities (`wp_get_abilities`)**
```
jonathanbossenger: This addition is not necessary.

docs/4.using-abilities.md
```diff
@@ -74,10 +74,10 @@ $ability = wp_get_ability( 'my-plugin/get-site-info' );
 if ( $ability ) {
     $site_info = $ability->execute(); // No input required
```
jonathanbossenger: We should stick to one variable name, either $site_info, or $result.

jonathanbossenger (changes requested):
Some minor documentation suggestions.

Also, the docs folder structure has been updated in #117, so you might have to fetch those updates.

includes/rest-api/endpoints/class-wp-rest-abilities-run-controller.php
Ref34t: Good catch! I've updated the code to preserve specific error messages from the permission callback instead of replacing them with a generic error.

The permission check now:
1. Returns the `WP_Error` directly if `check_permissions()` returns one (preserves context)
2. Only falls back to the generic error if it returns `false`

resolved here https://github.com/WordPress/abilities api/pull/95/commits/27b7b1a8215023b128cbcc55a27f51ac23e8774c

docs/4.using-abilities.md
Ref34t: This file has been removed in trunk as part of the documentation reorganization.

docs/4.using-abilities.md
Ref34t: This file has been removed from trunk as part of the documentation reorganization.

gziolo:
It looks like the changes applied in this PR after the rebase are no longer visible. @Ref34t, is there still anything we should fix in the docs? All the issues raised for the permission checks in the code were fixed during merge to WordPress core, and they are now being synced with https://github.com/WordPress/abilities-api/pull/126.

PR #96, opened 2025-09-30
Docs/semantic headings

jonathanbossenger:
While working on the 6.9 Call for testing post, I realised some of the doc headings are not using semantic markup, making it harder to link to specific sections in the docs. This PR fixes that by applying the correct heading markup


gziolo (approved):
Looks good to me 👍🏻

PR #99, opened 2025-10-02
Proposal: Add ability anotations

### LINKED ISSUE #62: Add property for marking Ability hints like destructive, read-only, idempotent

gziolo:
Closes #62.

## What

> Oftentimes we want to give the AI different instructions for when it's about to perform a destructive action, such as framing what's about to happen and requesting confirmation to the user. It would be helpful, therefore, to have a property of an Ability that marks it as a destructive ability.

I took some inspiration from [ToolAnnotation in MCP](https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations). For starters, we will need `read_only` to replace the current logic that uses `meta.tool` to distinguish between HTTP method usage (i.e., `GET` and `POST`) in requests.

## Why

The best explanation of the root problem that is being addressed comes from @galatanovidiu in https://github.com/WordPress/mcp-adapter/pull/48#discussion_r2399844161

> It would be great to leverage the same metadata in MCP, but we need to address a fundamental terminology mismatch first.
>
> The core issue: WordPress Abilities API and MCP use "tools" and "resources" to mean completely different things:
>
> **WordPress Abilities API:**
>
> Tools = callable functions that modify data
> Resources = callable functions that read data
>
> **MCP:**
>
> Tools = callable functions that the AI invokes automatically when needed
> Resources = static context that users manually attach to conversations
>
> So WordPress's "tool vs resource" distinction doesn't map to MCP's "tool vs resource"
>
> **Concrete example:** A site-info ability could be implemented as:
>
> MCP tool: AI calls it when it needs current site data during conversation
> MCP resource: User pre-attaches site info as reference context
>
> The choice depends on the usage pattern (AI-driven vs. user-driven), not the type of action.

Here, we remove all references to `tool` and `resource` nomenclature. Instead, we perform detection of the `GET` vs `POST` method in the REST API calls based on annotations that describe what the ability does: read only vs updates data.

## Fute considerations

With more properties, we could also start using `DELETE` HTTP method as I explore `idempotent` and `destructive` annotations.  I'm also interested in including freeform `instructions` so devs could consist of usage examples for both the user and for consideration to pass to LLMs.

## Testing

Run test:

```bash
npm run test:php
npm run test:client
```

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -30,7 +30,7 @@ class WP_Ability {
 		// Instructions on how to use the ability.
 		'instructions' => '',
 		// If true, the ability does not modify its environment.
-		'read_only'    => false,
+		'readonly'     => false,
 		/*
 		 * If true, the ability may perform destructive updates to its environment.
 		 * If false, the ability performs only additive updates.
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -19,6 +19,30 @@ declare( strict_types = 1 );
  * @see WP_Abilities_Registry
  */
 class WP_Ability {
+	/**
+	 * The default ability annotations.
+	 * They are not guaranteed to provide a faithful description of ability behavior.
+	 *
+	 * @since n.e.x.t
+	 * @var array<string,(bool|string)>
+	 */
+	const DEFAULT_ANNOTATIONS = array(
+		// An instruction on how to use the ability.
+		'instruction' => '',
+		// If true, the ability does not modify its environment.
+		'read_only'   => false,
+		/*
+		 * If true, the ability may perform destructive updates to its environment.
+		 * If false, the ability performs only additive updates.
+		 */
+		'destructive' => true,
+		/*
+		 * If true, calling the ability repeatedly with the same arguments will have no additional effect
+		 * on its environment.
+		 */
+		'idempotent'  => false,
+
+	);
 
 	/**
 	 * The name of the ability, with its namespace.
@@ -78,6 +102,14 @@ class WP_Ability {
 	protected $permission_callback;
 
 	/**
+	 * The ability annotations.
+	 *
+	 * @since n.e.x.t
+	 * @var array<string,(bool|string)>
+	 */
+	protected $annotations = self::DEFAULT_ANNOTATIONS;
+
+	/**
 	 * The optional ability metadata.
 	 *
 	 * @since 0.1.0
@@ -99,7 +131,7 @@ class WP_Ability {
 	 * @param string              $name The name of the ability, with its namespace.
 	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should
 	 *                                  include `label`, `description`, `input_schema`, `output_schema`,
-	 *                                  `execute_callback`, `permission_callback`, and `meta`.
+	 *                                  `execute_callback`, `permission_callback`, `annotations`, and `meta`.
 	 */
 	public function __construct( string $name, array $args ) {
 		$this->name = $name;
@@ -190,6 +222,12 @@ class WP_Ability {
 			);
 		}
 
+		if ( isset( $args['annotations']) && ! is_array( $args['annotations'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties should provide a valid `annotations` array.' )
+			);
+		}
+
 		if ( isset( $args['meta'] ) && ! is_array( $args['meta'] ) ) {
 			throw new \InvalidArgumentException(
 				esc_html__( 'The ability properties should provide a valid `meta` array.' )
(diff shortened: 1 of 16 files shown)


emdashcodes (approved):
I like this a lot. I prefer the annotations over the current tool/resource pattern.

The `instructions` will also be useful for providing tool instructions that various things like the MCP Adapter or Agents can adapt.

includes/abilities-api/class-wp-ability.php
```diff
@@ -19,6 +19,29 @@
  * @see WP_Abilities_Registry
  */
 class WP_Ability {
+	/**
```
galatanovidiu, line 22: I want to suggest we align our internal naming conventions with the established MCP protocol standards to reduce developer cognitive load and maintain consistency across the systems.

**Current approach:**
- Internally: `read_only`, `destructive`, `idempotent`
- MCP: `readOnlyHint`, `destructiveHint`, `idempotentHint`

**Suggested approach:**
Since MCP is a well-established protocol with defined conventions, I think we should adopt the MCP naming directly in our abilities API:
- Use: `readOnlyHint`, `destructiveHint`, `idempotentHint` internally
- This eliminates the need for translation in the MCP adapter
- Developers working with both our API and MCP won't need to context-switch between naming conventions
- Minimizes developer confusion when working across both systems

**Additional consideration - Instructions handling:**
For the `instructions` field, since MCP uses the tool description as instructions, we can handle this in the MCP adapter by appending our `instructions` content to the description field.

gziolo: The counterpoint would be that WordPress uses [snake case convention](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#naming-conventions) for naming properties. REST API is a good reflection of that. It's also inconvenient with JavaScript that uses camel case similarly to the MCP protocol. Sometimes, we even mix the two when functionality spans all layers, as with block types that also support the JSON format, which again uses camel case.

@felixarntz, and @swissspidy, what are your thoughts on that aspect?

felixarntz: +1 on aligning with WordPress's snake case convention:

* For the reason @gziolo mentioned: We shouldn't discard WordPress Coding Standards when it's convenient :)
* Another counterpoint is that this is the Abilities API which was from the beginning decided to be decoupled from MCP, and rightfully so.

IMO it's perfectly reasonable to have a translation layer in place as part of the MCP adapter, we shouldn't change the Abilities API to tie in particularly well with MCP when it leads to other problems.

Aside: I do agree it's painful to deal with snake_case from the server in the client-side JavaScript world where everything is camelCase. My recommendation for that reason is to use single-word names (like `readonly`, `destructive`, `idempotent`) wherever possible, as those would be both valid snake_case and valid camelCase. But obviously this is not the problem discussed here - it doesn't allow us to align with the MCP naming.

gziolo: Happy to rename from `read_only` to `readonly`. I'm not even 100% sure whether "read-only" is the ultimate correct English spelling 😄 I definitely saw many times `readonly` in the context of programming, for example, in TypeScript.

galatanovidiu: Whatever you decide to do is ok with me. Also, adapting those keys to the MCP standards is not a big problem.

I also understand your points on WP coding standards and the abilities-api being decoupled from MCP.

One thing I want to avoid when creating abilities for MCP usage: 
```php
$annotations = array(
    'read_only' => true, // translated to readOnlyHint
    'destructive' => false, // translated to destructiveHint
    'idempotent' => true, // translated to idempotentHint
    'openWorldHint' => false,
    'lastModified' => '...',
    'title' => '...',
    'audience' => '...',
    'priority' => '...',
)
```

This is inconsistent, non-standard, and ugly. 

So, what do you think we should do to avoid this situation?

felixarntz: Completely agree, we shouldn't have inconsistencies like this.

Where does this example list of identifiers come from? As far as this PR is concerned, I think we can use `readonly` to work around the difference between snake_case and camelCase.

If there are terms where we _need_ multi-word identifiers, I'd say we need to use snake_case if their source of truth is on the server (i.e. in PHP).

If in some JavaScript use-cases we need to translate stuff to be camelCase, we could probably do that consistently as well.

JasonTheAdams: I agree with @felixarntz. We shouldn't build anything in Abilities around the MCP standard. The MCP Adapter is a consumer of the Abilities API and should handle any translations and such needed to work. Any parameters added by the MCP Adapter should follow the pattern established here.

As a grammatical note, it kills me every time I see "readonly" in various contexts. We're just lying to ourselves. 😆

galatanovidiu: The MCP annotations are documented here:
https://modelcontextprotocol.io/specification/2025-06-18/schema#annotations
https://modelcontextprotocol.io/specification/2025-06-18/schema#toolannotations

gziolo: I used the same docs to shape basic needs for abilities based on the discussion in https://github.com/WordPress/abilities-api/issues/62.

gziolo: Renamed `read_only` key to `readonly`, and adjusted docs to use "read-only" form in https://github.com/WordPress/abilities-api/pull/99/commits/9f3a00cd3f5e60321a3e83ace5159cb680d02934.

gziolo:
I opened a follow-up with extended docs in https://github.com/WordPress/abilities-api/pull/104.

PR #100, opened 2025-10-03
Include JS build files in the Composer package in v0.2.0 release

gziolo:
This PR improves the configuration for Composer package by ensuring only files from `packages/client/build` and `packages/client/build-types` get included.

Now, the question is whether we really need `build-types` if `build` folder contains code ready for use on production exposed through `wp.abilities` global.

@emdashcodes or @budzanowski, what would be the purpose of including `build-types` here? Should we include `package.json`, too?

On the higher level, It feels to me like we should keep the build files in the top-level folder mirroring Gutenberg like:

`build/client/`

In Gutenberg, we store different types of content inside local build folder:
- `build` is Common JS version of the package (`import`/`export` replaced with `require`/`module.exports`)
- `build-module` is more like ES Modules version of the code

These are modified sources that are compiled with Babel with small changes that ensure these files can run in all browsers.

## Testing instructions

Run `git archive --format=zip HEAD -o archive.zip` to check the contents of the Composer package:

Screenshot 2025-10-03 at 10 10 25


gziolo:
For comparison, I include [54d9765](https://github.com/WordPress/abilities-api/pull/100/commits/54d9765d1c05e99ecdbd88d8a218962843ba8319) that has `packages/client/build-type` ignored:

Screenshot 2025-10-03 at 11 22 47

budzanowski:
> @emdashcodes or @budzanowski, what would be the purpose of including build-types here? Should we include package.json, too?

__build-types__
Yes. TypeScript developers who import the package need these files for type checking and autocomplete.
Should you include package.json?

__package.json__ files
Yes. It tells bundlers and TypeScript where to find the code and type definitions. Without it, imports like import { getAbilities } from '@wordpress/abilities' won't work.

__Should you mirror Gutenberg's structure with build/client/?__
Yes, if you want consistency. It makes the package structure familiar to WordPress developers and follows established conventions. Is it necessary - __no__

emdashcodes (approved):
The types would be useful for anyone pulling this in to a TS project. It is not strictly required though.

They would need to configure their tsconfig at the moment to include something like 

```
"@wordpress/abilities": [
    "./vendor/wordpress/abilities-api/packages/client/build-types"
]
```

This would only be considered temporary until we publish the npm package which will include the types.

---

I don't have a strong preference for where things are stored. I need to look a little closer at how all the Gutenberg packages are stored.

budzanowski (approved):

gziolo:
> The types would be useful for anyone pulling this in to a TS project. It is not strictly required though.

I was contemplating how that could be done. Happy to remove the commit that excludes `build-types` as a temporary measure.

Issue #101, opened 2025-10-06
Feature Proposal: Categories for Abilities API


galatanovidiu:
## Summary

Add a required category system to the WordPress Abilities API to enable organizing, filtering, and discovering abilities by functional domain. Every ability must be assigned to a category.

---

## Problem

The Abilities API currently lacks a native way to:
- Organize abilities by functional area (content, users, e-commerce, etc.)
- Filter abilities in REST API queries
- Group related abilities for UI rendering and documentation
- Enable taxonomy-style organization

---

## Proposed Solution

Implement a **category system** with these characteristics:

- **Required property** - Every ability must have a category assigned
- **Required registration** - Categories must be registered before they can be assigned to abilities
- **First-class property** - Category stored as ability property, not in metadata
- **Registry pattern** - Category registry manages all available categories with labels and descriptions
- **REST API integration** - Native filtering via query parameters
- **WordPress patterns** - Follows post types/taxonomies conventions

---

### 1. Category Class

**New:** `WP_Ability_Category`

**Properties:**
- `slug` - Unique identifier (pattern: `^[a-z0-9-]+$`)
- `label` - Human-readable name
- `description` - Detailed description

### 2. Category Registry

**New:** `WP_Abilities_Category_Registry`

**Pattern:** Singleton registry (required for all categories)

**Functionality:**
- Register/unregister categories
- Retrieve categories by slug
- List all registered categories
- Check if category is registered (used during ability registration)
- Validate category before ability assignment

**Hook:** `abilities_api_category_registry_init`

### 3. Ability Integration

**Enhancement:** `WP_Ability` class

**New Property:** `category` (string, required)

**Validation:**
- Must be a non-empty string
- Enforce slug pattern: `^[a-z0-9]+(-[a-z0-9]+)*$`
- Lowercase alphanumeric characters and dashes only

### 4. Helper Functions

**New functions in `abilities-api.php`:**

**Registration:**
- `wp_register_ability_category()`
- `wp_unregister_ability_category()`

**Retrieval:**
- `wp_get_ability_category()`
- `wp_get_ability_categories()`

**Filtering:**
- `wp_get_abilities_by_category()`

### 5. REST API

**Schema Enhancement:**
- Add `category` field to ability response
- Type: string with pattern validation
- Context: view, edit, embed
- Readonly: true
- Required: true

**Query Parameter:**
- `GET /wp/v2/abilities?category=content`
- Filter abilities before pagination
- Return standard paginated response

[...]

PR #102, opened 2025-10-06
Feature: add categories system

galatanovidiu:
## Summary

Implements a comprehensive category system for organizing abilities. Each ability must now belong to exactly one category, improving discoverability and enabling filtering of abilities by their purpose.

---

### Required `category` Parameter

**All abilities must now specify a `category` when registering.**

**Before:**
```php
wp_register_ability( 'my-plugin/get-data', array(
    'label' => 'Get Data',
    'description' => 'Retrieves data',
    // ... other args
));
```

**After:**
```php
// First, register a category
add_action( 'abilities_api_category_registry_init', 'my_plugin_register_categories' );
function my_plugin_register_categories() {
    wp_register_ability_category( 'data-retrieval', array(
        'label' => 'Data Retrieval',
        'description' => 'Abilities that retrieve data',
    ));
}

// Then register ability with category
add_action( 'abilities_api_init', 'my_plugin_register_ability' );
function my_plugin_register_ability() {
    wp_register_ability( 'my-plugin/get-data', array(
        'label' => 'Get Data',
        'description' => 'Retrieves data',
        'category' => 'data-retrieval', // REQUIRED
        // ... other args
    ));
}
```

---

#### `WP_Ability_Category`
Encapsulates category properties (slug, label, description).

**Location**: `includes/abilities-api/class-wp-ability-category.php`

#### `WP_Abilities_Category_Registry`
Singleton registry managing all registered categories.

**Location**: `includes/abilities-api/class-wp-abilities-category-registry.php`

**Features**:
- Validates category slugs (lowercase alphanumeric + dashes only)
- Requires `label` and `description` for all categories
- Prevents duplicate category registration
- Fires `abilities_api_category_registry_init` hook on initialization
- Applies `register_ability_category_args` filter before registration

### Core Changes to `WP_Ability`

**File**: `includes/abilities-api/class-wp-ability.php`

1. Added `$category` property (required string)
2. Added `get_category()` method
3. Category validation in constructor:
   - Must be non-empty string
   - Must match slug format: `^[a-z0-9]+(-[a-z0-9]+)*$`
4. Category is now part of ability's required properties

### Changes to `WP_Abilities_Registry`

**File**: `includes/abilities-api/class-wp-abilities-registry.php`

1. Added `get_abilities_by_category( string $category )` method
2. Category validation during ability registration:
   - Checks if category exists before registering ability
   - Returns null and triggers `_doing_it_wrong()` if category not found
3. Ensures category registry initializes before ability registry

### New API Functions

**File**: `includes/abilities-api.php`

#### Category Management
```php
// Register a category
wp_register_ability_category( string $slug, array $args ): ?WP_Ability_Category

// Unregister a category
wp_unregister_ability_category( string $slug ): ?WP_Ability_Category

// Get a specific category
wp_get_ability_category( string $slug ): ?WP_Ability_Category

// Get all categories
wp_get_ability_categories(): array
```

#### Ability Filtering
```php
// Get abilities by category
wp_get_abilities_by_category( string $category ): WP_Ability[]
```

#### Action: `abilities_api_category_registry_init`
Fires when the category registry is initialized. This is the **required hook** for registering categories.

```php
add_action( 'abilities_api_category_registry_init', 'my_plugin_register_categories' );
function my_plugin_register_categories( $registry ) {
    wp_register_ability_category( 'my-category', array(
        'label' => 'My Category',
        'description' => 'Description of my category',
    ));
}
```

**Parameters**:
- `$registry` (`WP_Abilities_Category_Registry`) - The category registry instance

#### Filter: `register_ability_category_args`
Allows modification of category arguments before validation.

```php
add_filter( 'register_ability_category_args', 'my_modify_category_args', 10, 2 );
function my_modify_category_args( array $args, string $slug ): array {
    if ( 'my-category' === $slug ) {
        $args['label'] = 'Modified Label';
    }
    return $args;
}
```

**Parameters**:
- `$args` (`array`) - Category arguments (label, description)
- `$slug` (`string`) - Category slug being registered

### REST API Updates

**File**: `includes/rest-api/endpoints/class-wp-rest-abilities-list-controller.php`

#### New Features

1. **Category field in responses**:
   ```json
   {
     "name": "my-plugin/get-data",
     "label": "Get Data",
     "description": "Retrieves data",
     "category": "data-retrieval",
     "input_schema": {},
     "output_schema": {},
     "meta": {}
   }
   ```

2. **Category filtering parameter**:
   ```bash
   GET /wp/v2/abilities?category=data-retrieval
   ```

3. **Schema updates**:
   - `category` added to ability schema as required field
   - `category` marked as readonly
   - Available in all contexts (view, edit, embed)

### Test Coverage

**New Test File**: `tests/unit/abilities-api/wpAbilityCategory.php` (550 lines)

**Updated Test Files**:
- `wpAbilitiesRegistry.php` - Added category setup/teardown
- `wpAbility.php` - Added category property to tests
- `wpRegisterAbility.php` - Added category validation tests
- `wpRestAbilitiesListController.php` - Added category filtering tests (125+ lines)
- `wpRestAbilitiesRunController.php` - Updated for category support (56+ lines)

#### 1. `docs/1.intro.md`
- Added **Category** to Core Concepts
- Updated Registry definition to include category registry
- Updated example to show category registration

#### 2. `docs/3.registering-abilities.md` (59 new lines)
- Added `category` as **Required** parameter
- Added "Registering Categories" section with:
  - `wp_register_ability_category()` function signature
  - Category slug conventions
  - Example category registration code
  - Other category functions documentation
- Updated all 4 code examples to include `category` field

#### 3. `docs/5.rest-api.md`
- Added `category` field to Ability Object schema
- Added `category` filter parameter to List Abilities endpoint
- Updated all JSON examples to include category field

#### 4. `docs/6.hooks.md` (70 new lines)
- Added `abilities_api_category_registry_init` action documentation
- Added `register_ability_category_args` filter documentation
- Updated Quick Links navigation

---

### For Plugin Developers

**Step 1: Register Categories**

Create categories before registering abilities:

```php
add_action( 'abilities_api_category_registry_init', 'my_plugin_register_categories' );
function my_plugin_register_categories() {
    // Group related abilities into logical categories
    wp_register_ability_category( 'data-retrieval', array(
        'label' => __( 'Data Retrieval', 'my-plugin' ),
        'description' => __( 'Abilities that fetch and return data', 'my-plugin' ),
    ));

    wp_register_ability_category( 'data-modification', array(
        'label' => __( 'Data Modification', 'my-plugin' ),
        'description' => __( 'Abilities that modify or update data', 'my-plugin' ),
    ));
}
```

**Step 2: Update Ability Registrations**

Add the `category` parameter to all `wp_register_ability()` calls:

```php
add_action( 'abilities_api_init', 'my_plugin_register_abilities' );
function my_plugin_register_abilities() {
    wp_register_ability( 'my-plugin/get-posts', array(
        'label' => __( 'Get Posts', 'my-plugin' ),
        'description' => __( 'Retrieves WordPress posts', 'my-plugin' ),
        'category' => 'data-retrieval', // ADD THIS
        // ... rest of args
    ));
}
```

### Category Slug Naming Conventions

**Valid Slugs**:
- ✅ `data-retrieval`
- ✅ `user-management`
- ✅ `ecommerce`
- ✅ `analytics-123`

**Invalid Slugs**:
- ❌ `Data-Retrieval` (uppercase)
- ❌ `data_retrieval` (underscores)
- ❌ `data.retrieval` (dots)
- ❌ `data/retrieval` (slashes)
- ❌ `-data-retrieval` (leading dash)
- ❌ `data-retrieval-` (trailing dash)

## Related Issues

This PR implements the category system from issue #101.


docs/1.intro.md
```diff
@@ -35,11 +36,22 @@ It acts as a central registry, making it easier for different parts of WordPress
 ## Registration Example
 
 ```php
+// First, register a category
+add_action( 'abilities_api_category_registry_init', 'my_plugin_register_category');
```
gziolo: Nitpick, it could help to find a shorter name for the action
```suggestion
add_action( 'abilities_category_init', 'my_plugin_register_category');
```

or a bit differently structured than `abilities_api_init` but more in natural English:

```suggestion
add_action( 'ability_categories_init', 'my_plugin_register_category');
```

galatanovidiu: Yes, I like this idea.
I will apply the changes

galatanovidiu: In fact, what do you think about `abilities_api_categories_init`
I think it will better reflect the functionality

ability_categories sounds like we initialize categories/ability

galatanovidiu: So, I've replaced it with `abilities_api_categories_init`. If needed, we can easily search and replace it with another name.

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,240 @@
...
+	 *
+	 * @since 0.3.0
+	 * @var \WP_Ability_Category[]
+	 */
+	private $registered_categories = array();
+
+	/**
+	 * Registers a new category.
+	 *
+	 * Do not use this method directly. Instead, use the `wp_register_ability_category()` function.
```
JasonTheAdams, line 40: I'm curious as to why? Especially since we're passing the registry instance in the hook.

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,240 @@
...
+	 * @since 0.3.0
+	 * @var ?self
+	 */
+	private static $instance = null;
+
+	/**
+	 * Holds the registered categories.
+	 *
+	 * @since 0.3.0
+	 * @var \WP_Ability_Category[]
```
JasonTheAdams, line 33: Is `list<type>` allowed in WP core? Technically, this is a true PHP list (passing `array_is_list()`). Whereas `type[]` is a shorthand for `array<int, type>`, which doesn't mean it's truly a list (as the key can be any integer).

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,240 @@
...
+	 * Do not use this method directly. Instead, use the `wp_get_ability_category()` function.
+	 *
+	 * @since 0.3.0
+	 *
+	 * @see wp_get_ability_category()
+	 *
+	 * @param string $slug The slug of the registered category.
+	 * @return ?\WP_Ability_Category The registered category instance, or null if it is not registered.
+	 */
+	public function get_registered( string $slug ): ?WP_Ability_Category {
```
JasonTheAdams, line 187: This is a registry, so everything is registered, so what we're really getting here is a category.
```suggestion
	public function get_category( string $slug ): ?WP_Ability_Category {
```

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,240 @@
...
+	 *
+	 * Do not use this method directly. Instead, use the `wp_get_ability_categories()` function.
+	 *
+	 * @since 0.3.0
+	 *
+	 * @see wp_get_ability_categories()
+	 *
+	 * @return array<string,\WP_Ability_Category> The array of registered categories.
+	 */
+	public function get_all_registered(): array {
```
JasonTheAdams, line 159: Same as the other comment. I think this is more intuitively true.
```suggestion
	public function get_all_categories(): array {
```

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,240 @@
...
+
+	/**
+	 * Checks if a category is registered.
+	 *
+	 * @since 0.3.0
+	 *
+	 * @param string $slug The slug of the category.
+	 * @return bool True if the category is registered, false otherwise.
+	 */
+	public function is_registered( string $slug ): bool {
```
JasonTheAdams, line 171: I don't feel quite as strongly on this one, but this reads better to me: `$registry->has_category('woo')`
```suggestion
	public function has_category( string $slug ): bool {
```

includes/abilities-api/class-wp-ability-category.php
```diff
@@ -0,0 +1,152 @@
...
+	 * @param array<string,mixed> $args An associative array of arguments for the category.
+	 */
+	public function __construct( string $slug, array $args ) {
+		if ( empty( $slug ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The category slug cannot be empty.' )
+			);
+		}
+
+		$this->slug = sanitize_title( $slug );
```
JasonTheAdams: Should we validate this or silently sanitize it? 🤔 

I'd be inclined to do something like `if ( $slug !== sanitize_title( $slug ) )` to validate, more like how we're validating `$args`.

includes/abilities-api/class-wp-ability-category.php
```diff
@@ -0,0 +1,152 @@
...
+	 * @since 0.3.0
+	 *
+	 * @return string The detailed description for the category.
+	 */
+	public function get_description(): string {
+		return $this->description;
+	}
+
+	/**
+	 * Wakeup magic method.
```
JasonTheAdams, line 185: I'm guessing this is here for security purposes? If so, we should mention that.

tests/unit/abilities-api/wpAbilityCategory.php
```diff
@@ -0,0 +1,498 @@
...
+		// Uppercase characters not allowed.
+		$result = $this->register_category_during_hook(
+			'Test-Math',
+			array(
+				'label'       => 'Math',
+				'description' => 'Mathematical operations.',
+			)
+		);
+
+		$this->assertNull( $result );
```
JasonTheAdams, line 156: Is there a way to check for doing_it_wrong to have been called? Not a big deal if not.

JasonTheAdams (changes requested):
Great work, @galatanovidiu! I left a handful of small suggestions.

I'm still not a big fan of the init hooks occurring in the `get_instance()` methods, but that may just need to be where these things happen, as WP doesn't really have a formal bootsrapping system.

tests/unit/abilities-api/wpAbilityCategory.php
galatanovidiu: Done
I've added a `record_doing_it_wrong()` method that hooks into `doing_it_wrong_run` to capture all `_doing_it_wrong` calls. Tests can then use `assertDoingItWrongTriggered()` to verify the expected function was called, with optional message validation.

Please check: https://github.com/WordPress/abilities-api/pull/102/commits/110d403450ea5ba6ef5768dafeaf9b3336f7b0f6

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: I think it's fine as is because these names containing `registered` keyword are used in several places in WP core. 3rd party consumers won't use these names but function helpers instead.

includes/abilities-api/class-wp-ability-category.php
gziolo: Likely, it isn't documented this way in WP core, as it was added to harden security. However, we can add a note in this repo 👍🏻

tests/unit/abilities-api/wpAbilityCategory.php
gziolo: You should use `@expectedIncorrectUsage` and it will get automatically handled in WP core. We don't have the same PHPUnit setup here unfortunately so it won't work out of the box in this repo. Alternatively, you can use `$this->setExpectedIncorrectUsage( ... )`.

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: That's a hint for documentation purposes to promote usage of `wp_register_ability_category` helper. In WP Core, the tendency is to keep similar registries more like internal objects if possible.

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: PHP 7.2 is the minimum required version in WP core and it largely influences the state of linting. I didn't find any occurrence of `@var list` in production code, but there are some type hints in tests. 

The most important question is whether it's going to be compatible with the developer portal like this page:

https://developer.wordpress.org/reference/classes/wp_block_type_registry/get_all_registered/

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: Example: 

https://developer.wordpress.org/reference/classes/wp_block_type_registry/get_registered/

tests/unit/abilities-api/wpAbilityCategory.php
galatanovidiu: @gziolo 
The reason for having a custom `assertDoingItWrongTriggered` is that I also want to verify the correct message is triggered, as there are multiple `doing-it-wrong` messages/methods. 
`$this->setExpectedIncorrectUsage` is only checking for the correct function/method 

If you think this is too verbose, I can remove it and use `$this->setExpectedIncorrectUsage`. I will follow your lead on this one.

`@expectedIncorrectUsage` is used here. Without it, it will trigger an `Unexpected incorrect usage notice for ...` message.

gziolo: In the cases where it’s sufficient to verify the method name that was passed as the first argument, I would go with the PHPdoc annotation as it’s the most popular approach in tests. If you need more granular distinction then feel free to keep the extra logic if the benefits justify the complexity introduced.

gziolo: Maybe a separate method could also help to distinguish different paths in tests but only if this makes sense to extract method. It would be worse to let the tests dictate the class design.

includes/abilities-api.php
```diff
@@ -99,3 +100,67 @@ function wp_get_ability( string $name ): ?WP_Ability {
 function wp_get_abilities(): array {
 	return WP_Abilities_Registry::get_instance()->get_all_registered();
 }
+
+/**
+ * Registers a new ability category.
+ *
+ * @since 0.3.0
```
justlevine: here and elsewhere

```suggestion
 * @since n.e.x.t
```

includes/bootstrap.php
```diff
@@ -19,7 +19,7 @@
 
 // Version of the plugin.
 if ( ! defined( 'WP_ABILITIES_API_VERSION' ) ) {
-	define( 'WP_ABILITIES_API_VERSION', '0.2.0' );
+	define( 'WP_ABILITIES_API_VERSION', '0.3.0' );
```
justlevine: This deserves its own call out. This PR is focused on adding a categories system, not cutting a new release.

```suggestion
	define( 'WP_ABILITIES_API_VERSION', '0.2.0' );
```

JasonTheAdams: Good note. Versioning should be done in a subsequent PR.

includes/abilities-api/class-wp-ability.php
```diff
@@ -234,12 +243,26 @@ protected function prepare_properties( array $args ): array {
 			);
 		}
 
+		// VALIDATE: Category is required and must be a string.
+		if ( empty( $args['category'] ) || ! is_string( $args['category'] ) ) {
```
justlevine: What's the reasoning for making categories required?

(Trying to put aside my general reservations about 1) categories as the correct shape and 2) implemented as a top-level arg instead of as `meta`, but that could be making it harder for me to intuit the reason here from the PR alone )

JasonTheAdams: This was discussed on the call, but to document here: The number of abilities is highly likely to scale very quickly, which means in systems like MCP it will be important to have a way of grouping the abilities. This is necessary for techniques like Tool Layering. If we make this optional, or have a fallback "uncategorized" category, then we've not really solved the problem because many folks are likely to make it a dump site.

emdashcodes (commented):
For the client, I think we need a couple (small) things:

- getAbilities should include a way to filter by category via the REST endpoint
- We need to add `category` to the `Ability` type so people can pass in a category to their client side Abilities: https://github.com/WordPress/abilities-api/blob/a9039a30a64aac4cb2671faa99276e3476ce947e/packages/client/src/types.ts#L25

I think those would be small additions to this PR, but let me know if you need me to take a look @gziolo.

includes/abilities-api.php
galatanovidiu: Done

includes/bootstrap.php
galatanovidiu: Done

docs/6.hooks.md
```diff
@@ -5,13 +5,49 @@ The Abilities API provides [WordPress Action and Filter Hooks](https://developer
 ## Quick Links
 
 - [Actions](#actions)
+  - [`abilities_api_categories_init`](#abilities_api_categories_init)
```
gziolo, line 8: Outside of this PR, mostly note to myself and @jonathanbossenger, we are missing `abilities_api_init` covered here. That one is fundamental 😄

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: By the way, the same names, `get_all_registered`, `get_registered`, `is_registered` are also used in the abilities registry. So we can change it in one go for both registries, but I would keep it outside of this PR so it doesn't block the work here.

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: See https://github.com/WordPress/abilities-api/pull/102/files#r2416128992.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -219,6 +238,10 @@ public static function get_instance(): self {
 		if ( null === self::$instance ) {
 			self::$instance = new self();
 
+			// Ensure category registry is initialized first to allow categories to be registered
+			// before abilities that depend on them.
+			WP_Abilities_Category_Registry::get_instance();
```
gziolo, line 246: Mostly curious here, but won't it happen automatically, the first time you start validating the name of the category passed to the ability?

gziolo: Ah, it doesn't validate anymore after some intermediate feedback, before we concluded that the category has to be mandatory. So if we decide to check whether the category name is registered, then we might be able to remove this code here if I don't dream it will work in a proper sequence.

tests/unit/abilities-api/wpRegisterAbility.php
```diff
@@ -465,4 +491,27 @@ public function test_get_all_registered_abilities() {
...
+	/**
+	 * Tests registering an ability with non-existent category.
+	 *
+	 * @expectedIncorrectUsage WP_Abilities_Registry::register
+	 */
+	public function test_register_ability_nonexistent_category(): void {
+		do_action( 'abilities_api_init' );
+
+		// Ensure category doesn't exist.
+		if ( WP_Abilities_Category_Registry::get_instance()->is_registered( 'nonexistent' ) ) {
```
gziolo: Maybe assert here instead? It should never happen, so failing tests makes perfect sense.

includes/abilities-api/class-wp-ability-category.php
galatanovidiu: I've removed the sanitize_title.
We already [validate the slug](https://github.com/galatanovidiu/abilities-api/blob/7a5047bbba7bd1aae585047fdceb757d0309af0a/includes/abilities-api/class-wp-abilities-category-registry.php#L83) on `\WP_Abilities_Category_Registry::register()`

includes/abilities-api/class-wp-abilities-category-registry.php
galatanovidiu: I chose `get_all_registered()` to maintain consistency with the naming pattern used in `WP_Abilities_Category_Registry`

includes/abilities-api/class-wp-abilities-category-registry.php
galatanovidiu: Same as above, to maintain consistency :)

gziolo:
> For the client, I think we need a couple (small) things:

Good callout, @emdashcodes. This PR is big enough already, so let's tackle it seperately. For starters, we have a system in place that will ignore the category on the client, which isn't ideal, but things will continue to work. We will also have to sort out how to pass registered categories to the client in case someone wants to consume these registered labels and other metadata.

includes/abilities-api/class-wp-abilities-registry.php
galatanovidiu: Great question! While the category registry would technically auto-initialize during the first ability validation, this would cause the abilities_api_categories_init action to fire inside or during the abilities_api_init action, rather than before it.

The explicit initialization ensures a predictable execution order: abilities_api_categories_init always fires completely before abilities_api_init begins. This guarantees that all category registrations are complete before any ability registrations start.

Without this explicit initialization, we'd have a race condition where the hook order depends on whether the first ability registered happens to have a category or not (and based on the requirements, it must have one).

---

@gziolo I've just seen your last message after posting this :)

gziolo: Thank you for the additional explanation. It makes sense to me 👍🏻

tests/unit/abilities-api/wpRegisterAbility.php
galatanovidiu: Good catch! I've updated the test to use an assertion instead of the conditional check. 

The category 'nonexistent' should never be registered when this test runs, so if it does exist, it indicates a test isolation problem. Now the test will fail immediately with a clear message rather than silently working around the issue.

I also found and fixed two other similar cases:
- `test_ability_requires_existing_category` in wpAbilityCategory.php
- `test_filter_by_nonexistent_category` in wpRestAbilitiesListController.php

All three now use assertions to verify the test preconditions, which makes the tests more robust and the intent clearer.

Changes pushed in [6bde79a](https://github.com/WordPress/abilities-api/pull/102/commits/6bde79ad4648abb31c22b57a13f41d5afed62de3).

includes/abilities-api/class-wp-ability-category.php
galatanovidiu: Comment added

tests/unit/abilities-api/wpAbilityCategory.php
galatanovidiu: You are right, the tests should not dictate the class design. 
For now, I'll keep it as is. If we see that these assertions cause problems in time, we can always remove them.

includes/abilities-api/class-wp-abilities-category-registry.php
JasonTheAdams: Ah, right, this all gets turned into documentation. Fair enough! 😄

includes/abilities-api/class-wp-ability.php
```diff
@@ -243,6 +254,19 @@ protected function prepare_properties( array $args ): array {
...
+				esc_html__( 'The ability properties must contain a `category` string.' )
+			);
+		}
+
+		// Validate category slug format.
+		if ( ! preg_match( '/^[a-z0-9]+(-[a-z0-9]+)*$/', $args['category'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'Category slug must contain only lowercase alphanumeric characters and dashes.' )
+			);
+		}
```
justlevine: Why?

includes/abilities-api/class-wp-ability.php
```diff
@@ -136,8 +145,9 @@ class WP_Ability {
 	 * @see wp_register_ability()
 	 *
 	 * @param string              $name The name of the ability, with its namespace.
+
```
justlevine:
```suggestion
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -136,8 +145,9 @@ class WP_Ability {
...
 	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should include
-	 *                                  `label`, `description`, `input_schema`, `output_schema`, `execute_callback`,
+	 *                                  `label`, `description`, `category`, `input_schema`, `output_schema`, `execute_callback`,
```
justlevine: Does WPCS have a char limit per line? For the most part we've been breaking at 120ch but idk if its necessary.

includes/abilities-api/class-wp-ability.php
justlevine: Nit: 
```suggestion
		}

```

includes/abilities-api/class-wp-ability.php
```diff
@@ -243,6 +254,19 @@ protected function prepare_properties( array $args ): array {
 			);
 		}
 
+		// VALIDATE: Category is required and must be a string.
```
justlevine: Nit: seems weird that only the category validations have comments, but this particular one is AI and doesn't match our codebase.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -96,6 +97,24 @@ public function register( string $name, array $args ): ?WP_Ability {
...
+						esc_html__( 'Category "%1$s" is not registered. Please register the category before assigning it to ability "%2$s".' ),
+						esc_attr( $args['category'] ),
+						esc_attr( $name )
+					),
+					'n.e.x.t'
+				);
+				return null;
+			}
+		}
+
```
justlevine, line 119: This feels counterintuitive as 
1. category _always_ needs to be set
2. if a category is not registered then we'll never get to `WP_Ability::prepare_properties()`.

IMO should be treated like the rest of the ~meta~ args, and moved into `WP_Ability`

includes/abilities-api/class-wp-abilities-registry.php
justlevine: Not sure I follow 🤔

@galatanovidiu why do we need the hook in the first place? Anything preceding `abilities_api_init` would work just fine (see `register_block_pattern_category` https://developer.wordpress.org/block-editor/reference-guides/block-api/block-patterns/#register_block_pattern_category )

(Depending on your answer, I'll probably have follow-ups about why its not either in the constructor or before `new self()`, but the awkwardness of instantiating a different singleton inside a `::get_instance()` is a pretty big red flag that should cause us to rethink our assumptions, not architecture some faux-DI workaround )

justlevine: (Unresolving for visibility)

includes/abilities-api/class-wp-abilities-category-registry.php
justlevine: Slightly more relevant example for posterity (WP_Block_Type IRL more mirrors our WP_Ability): https://github.com/WordPress/wordpress-develop/blob/6.8.3/src/wp-includes/class-wp-block-pattern-categories-registry.php

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,240 @@
...
+		return self::$instance;
+	}
+
+	/**
+	 * Wakeup magic method.
+	 *
+	 * @since n.e.x.t
+	 * @throws \UnexpectedValueException If any of the registered categories is not an instance of WP_Ability_Category.
+	 */
+	public function __wakeup(): void {
```
justlevine, line 234: Is there a reason we'd be (un)serializing the registry?

includes/abilities-api/class-wp-ability-category.php
```diff
@@ -0,0 +1,196 @@
...
+declare( strict_types = 1 );
+
+/**
+ * Encapsulates the properties and methods related to a specific ability category.
+ *
+ * @since n.e.x.t
+ *
+ * @see WP_Abilities_Category_Registry
+ */
+class WP_Ability_Category {
```
justlevine: Any reason why we need to support extending this for WP6.9? Making the class final and/or the props private gives us more time to iterate and figure out what sort of categorization Abilities actually need.

```suggestion
final class WP_Ability_Category {
```

docs/3.registering-abilities.md
```diff
@@ -40,6 +41,60 @@ The `$id` parameter must follow the pattern `namespace/ability-name`:
 - **Convention:** Use your plugin slug as the namespace, like `my-plugin/ability-name`.
 - **Examples:** `my-plugin/update-settings`, `woocommerce/get-product`, `contact-form/send-message`, `analytics/track-event`
 
+## Registering Categories
```
justlevine: Does this deserve its own file? It would certainly be less disruptive than trying to fix the flow with the existing doc (e.g. the nested example vs `## Code Examples` section ).

cc @jonathanbossenger

includes/abilities-api/class-wp-ability.php
```diff
@@ -124,6 +132,7 @@ class WP_Ability {
 	 */
 	protected $show_in_rest = false;
 
+
```
justlevine: Not sure why our linter isnt picking up on these 🤔

```suggestion
```

includes/abilities-api/class-wp-ability-category.php
```diff
@@ -0,0 +1,196 @@
...
+	 */
+	protected $description;
+
+	/**
+	 * The optional category metadata.
+	 *
+	 * @since n.e.x.t
+	 * @var array<string,mixed>
+	 */
+	protected $meta = array();
```
justlevine, line 53: (Why) does an Ability Category need meta?

includes/abilities-api/class-wp-ability.php
galatanovidiu: Good catch, this is no longer required because the category is now mandatory, and the format is validated when the category is created.

includes/abilities-api/class-wp-ability.php
galatanovidiu: We can enforce this using `Generic.Files.LineLength` if we want.
For now, I will just reformat this

includes/abilities-api/class-wp-abilities-category-registry.php
galatanovidiu: I see no reason to do that.
This is here for consistency.

includes/abilities-api/class-wp-ability-category.php
galatanovidiu: I agree with this.
@gziolo @JasonTheAdams ?

includes/abilities-api/class-wp-ability-category.php
galatanovidiu: Categories can have optional data like an icon or plugin reference. The meta field lets us store that extra information without changing the core schema.

includes/abilities-api/class-wp-ability.php
gziolo: I don't think this is an issue in WordPress Core, because I ran most of the files on CI:

https://github.com/WordPress/wordpress-develop/pull/9410

We can revisit such details later in a separate PR with a bulk update if that is considered useful.

includes/abilities-api/class-wp-ability-category.php
gziolo: If we don't want it to be extended, then `final` makes sense. It's a familiar pattern there.

gziolo:
I see multiple merge conflicts after merging https://github.com/WordPress/abilities-api/pull/107. Maybe, let's also try merging https://github.com/WordPress/abilities-api/pull/104 before resolving conflicts to avoid additional hassle. If you need any help with resolving these conflicts, let me know. I hope it's only due to `category` moved out of `meta` in these files. Sorry about the trouble 😓

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: It's used in WP core most likely to prevent attacks like that: https://owasp.org/www-community/vulnerabilities/PHP_Object_Injection.

includes/abilities-api/class-wp-ability.php
galatanovidiu: The `Squiz.WhiteSpace.SuperfluousWhitespace.EmptyLines` rule only detects extra blank lines inside function bodies, not between class properties/methods.

I've tried `SlevomatCodingStandard.Classes.MethodSpacing` this one can enforce a set number of blank lines between class members also

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: That looks also applicable: https://www.sonarsource.com/blog/php-object-injection/

includes/abilities-api/class-wp-abilities-registry.php
galatanovidiu: Good points! Here's my reasoning:

**Why Registry over `prepare_properties()`:**

The category validation checks against `WP_Abilities_Category_Registry` - this is a cross-registry dependency, not a property validation concern. The Registry handles registration requirements (name format, duplicates, external references), while `prepare_properties()` validates argument types and required fields.

**Why the `isset()` check:**

Category validation happens in two stages:
1. **Registry**: "If provided, does it exist in the category registry?" (existence check)
2. **`prepare_properties()`**: "Is category provided at all?" (required field check)

This keeps all required-field validation together (label, description, category, callbacks) while letting the Registry handle cross-registry concerns. If the category is missing, `prepare_properties()` throws immediately. If provided but unregistered, the Registry catches it even earlier.

If you guys think it is appropriate, I can move it to `prepare_properties()`, or another place.

@gziolo @JasonTheAdams, what do you think?

includes/abilities-api/class-wp-ability.php
galatanovidiu: Done

includes/abilities-api/class-wp-ability.php
galatanovidiu: Done

includes/abilities-api/class-wp-ability.php
galatanovidiu: I've modified this particular line.
But I agree that we should review the entire code on a separate PR if required.

includes/abilities-api/class-wp-ability.php
galatanovidiu: Done

includes/abilities-api/class-wp-ability-category.php
galatanovidiu: Done

docs/3.registering-abilities.md
galatanovidiu: Yes, I agree that we need a new file instead of just a new section
Should we do this on a new PR?

galatanovidiu:
> I see multiple merge conflicts after merging https://github.com/WordPress/abilities-api/pull/107. Maybe, let's also try merging https://github.com/WordPress/abilities-api/pull/104 before resolving conflicts to avoid additional hassle. If you need any help with resolving these conflicts, let me know. I hope it's only due to category moved out of meta in these files. Sorry about the trouble 😓

Don't worry about it, I had to resolve some conflicts yesterday, too. This resulted in some formatting problems, which @justlevine caught.
I will wait until #104 is out

docs/3.registering-abilities.md
JasonTheAdams: Let's do it here since we're introducing categories.

includes/abilities-api/class-wp-ability.php
galatanovidiu: problem resolved

gziolo:
#104 landed. Let’s tackle this one next as soon as all feedback is addressed 😁

docs/3.registering-abilities.md
galatanovidiu: done

galatanovidiu:
> #104 landed. Let’s tackle this one next as soon as all feedback is addressed 😁

@gziolo, conflicts have been resolved.

includes/abilities-api/class-wp-abilities-category-registry.php
JasonTheAdams: Right, this is meant to protect against POP chains... but it won't. 😅 

The problem with `__wakeup()`, and why it's a deprecated magic method, is because it only fires _after_ the unserialization process has already occurred. So if the array being examined did in fact contain a malicious object, it was already instantiated and its malicious constructor code fire. Boo!

In PHP 7.4 there's an `__unserialize()` method introduced, which fires _before_ unserialization occurs. Buuuut, we can't support there here so that's not an option.

A slightly better way of locking it down is to do something like:
```php
public function __sleep(): array {
    throw new \LogicException('Registry must not be serialized.');
}
public function __wakeup(): void {
    throw new \LogicException('Registry must not be unserialized.');
}
```

Since we don't see a reason for the registry to be serialized, we should prevent it from being serialized in the first place. We should also just throw an exception if it is unserialized.

Is this bulletproof? Well... no, not at all. That's why these are deprecated. Someone still could manually create the serialized string themselves, and by the time `__wakeup()` throws its exception, it's already unserialized.

The only real way to prevent a POP chain attack in PHP 7.3 and below is by:
1. Avoiding passing untrusted strings to `unserialize()`
2. Use the `allowed_classes` parameter in `unserialize()`

There's no actual way within the class itself (without the PHP 7.4 magic methods) to prevent this attack from occurring. Yet another good reason for WP to update to PHP 7.4 when possible.

Thanks for coming to my TED Talk! 😆

galatanovidiu: Done. 
Added `__sleep()` and `__wakeup()` methods. They will both throw \LogicException.
Modifications done on `WP_Ability_Category` and `WP_Abilities_Category_Registry`

gziolo: @JasonTheAdams, much appreciate your feedbck here. This was very insightful!

docs/1.intro.md
```diff
@@ -35,11 +36,22 @@ It acts as a central registry, making it easier for different parts of WordPress
 ## Registration Example
 
 ```php
+// First, register a category
```
gziolo: Let's ensure it's clear that this is applicable only for new categories.
```suggestion
// First, register a category, or use one of the existing categories.
```

docs/7.registering-categories.md
```diff
@@ -0,0 +1,80 @@
...
+  - Leading/trailing dashes: `-data`, `data-`
+  - Double dashes: `data--retrieval`
+
+## Other Category Functions
+
+- `wp_unregister_ability_category( string $slug )` - Remove a registered category. Returns the unregistered category instance or `null` on failure.
+- `wp_get_ability_category( string $slug )` - Retrieve a specific category by slug. Returns the category instance or `null` if not found.
+- `wp_get_ability_categories()` - Get all registered categories as an associative array keyed by slug.
+
+## Hooks
```
gziolo: This is covered in Hooks, so maybe we should remove it from here.

docs/7.registering-categories.md
```diff
@@ -0,0 +1,80 @@
...
+- **Format:** Must contain only lowercase alphanumeric characters (`a-z`, `0-9`) and hyphens (`-`).
+- **Valid examples:** `data-retrieval`, `ecommerce`, `site-information`, `user-management`, `category-123`
+- **Invalid examples:**
+  - Uppercase: `Data-Retrieval`, `MyCategory`
+  - Underscores: `data_retrieval`
+  - Special characters: `data.retrieval`, `data/retrieval`, `data retrieval`
+  - Leading/trailing dashes: `-data`, `data-`
+  - Double dashes: `data--retrieval`
+
+## Other Category Functions
```
gziolo, line 57: This is something we could extend a bit in a follow-up.

includes/abilities-api.php
```diff
@@ -102,3 +103,67 @@ function wp_get_ability( string $name ): ?WP_Ability {
...
+ *
+ * @param string              $slug The unique slug for the category. Must contain only lowercase
+ *                                  alphanumeric characters and dashes.
+ * @param array<string,mixed> $args An associative array of arguments for the category. This should
+ *                                  include `label` and `description`.
+ * @return ?\WP_Ability_Category The registered category instance on success, null on failure.
+ *
+ * @phpstan-param array{
+ *   label?: string,
+ *   description?: string,
```
gziolo: What about meta, should it be included here, and in the description?

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,246 @@
...
+	 * @return ?\WP_Ability_Category The registered category instance on success, null on failure.
+	 *
+	 * @phpstan-param array{
+	 *   label?: string,
+	 *   description?: string,
+	 *   ...<string, mixed>
+	 * } $args
+	 */
+	public function register( string $slug, array $args ): ?WP_Ability_Category {
+		if ( ! doing_action( 'abilities_api_categories_init' ) ) {
```
gziolo, line 60: I put the same check for abilities inside the function `wp_register_ability()`, mostly to simplify the testing, as you don't have to enforce this action for the registry tests. In practice, we expect developers to use `wp_register_ability_category()` so this is where it matters the most. 

Definitely not a blocker, as I don't feel strongly about it, and we can also tackle it later in a follow-up if you agree with my reasoning, mostly for consistency.

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,246 @@
...
+	 *
+	 * @param string $slug The slug of the registered category.
+	 * @return ?\WP_Ability_Category The unregistered category instance on success, null on failure.
+	 */
+	public function unregister( string $slug ): ?WP_Ability_Category {
+		if ( ! $this->is_registered( $slug ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				/* translators: %s: Category slug. */
+				sprintf( esc_html__( 'Category "%s" not found.' ), esc_attr( $slug ) ),
```
gziolo: Small detail, but might be important for gaining the context when translating in the interface disconnected from the code:

```suggestion
				/* translators: %s: Ability category slug. */
				sprintf( esc_html__( 'Ability category "%s" not found.' ), esc_attr( $slug ) ),
```

includes/abilities-api/class-wp-abilities-category-registry.php
```diff
@@ -0,0 +1,246 @@
...
+	 *
+	 * @param string $slug The slug of the registered category.
+	 * @return ?\WP_Ability_Category The registered category instance, or null if it is not registered.
+	 */
+	public function get_registered( string $slug ): ?WP_Ability_Category {
+		if ( ! $this->is_registered( $slug ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				/* translators: %s: Category slug. */
+				sprintf( esc_html__( 'Category "%s" not found.' ), esc_attr( $slug ) ),
```
gziolo: Similar minor note regaring translations:
```suggestion
				/* translators: %s: Ability category slug. */
				sprintf( esc_html__( 'Ability category "%s" not found.' ), esc_attr( $slug ) ),
```

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -98,6 +99,24 @@ public function register( string $name, array $args ): ?WP_Ability {
...
 
+		// Validate category exists if provided (will be validated as required in WP_Ability).
+		if ( isset( $args['category'] ) ) {
+			$category_registry = WP_Abilities_Category_Registry::get_instance();
+			if ( ! $category_registry->is_registered( $args['category'] ) ) {
+				_doing_it_wrong(
+					__METHOD__,
+					sprintf(
+						/* translators: %1$s: category slug, %2$s: ability name */
+						esc_html__( 'Category "%1$s" is not registered. Please register the category before assigning it to ability "%2$s".' ),
```
gziolo: My consistent minor message about verbiage. Although here might be less necessary as the message is quite detailed:
```suggestion
						/* translators: %1$s: ability category slug, %2$s: ability name */
						esc_html__( 'Ability category "%1$s" is not registered. Please register the category before assigning it to ability "%2$s".' ),
```

includes/abilities-api/class-wp-abilities-registry.php
gziolo: So it moves most of the validation to the category registry, as the existence of the category name there is the strongest indicator that it has already been validated. No need for additional checks here. There is strong coupling with the category registry, which is fine to have at both levels. @galatanovidiu explained that the basic check for existence still happens inside `prepare_properties()`, so all checks necessary are covered. I'm fine keeping it here.

gziolo (commented):
I left a few minor notes for the production logic. This looks solid and I'm planning to approve the PR as soon as I review unit tests. Some of my feedback is perfectly suited as follow-up work as it focuses on code quality which might be even easier to review and discuss seperately. @galatanovidiu, can you collect the discussed code quality improvements so we have a good overview of what's left when making the final call? Again, I didn't disovered any blockers so far.

docs/7.registering-categories.md
```diff
@@ -0,0 +1,80 @@
+# 7. Registering Categories
```
gziolo, line 1: @jonathanbossenger, we really need to remove these numbers, as I would put this one next to registering categories as it fits better there. It's another instance where this ordering causes trouble 😅 

Let's also make sure to list this new document in README somwhere next to:

https://github.com/WordPress/abilities-api/blob/4dc57b3b64d6ec00caccb4ecff33a91f4554be9b/README.md?plain=1#L23

tests/unit/rest-api/wpRestAbilitiesListController.php
```diff
@@ -573,4 +622,85 @@ public function test_invalid_pagination_parameters( array $params ): void {
...
+		$data = $response->get_data();
+		$this->assertArrayHasKey( 'category', $data );
+		$this->assertEquals( 'math', $data['category'] );
+		$this->assertIsString( $data['category'], 'Category should be a string' );
+	}
+
+	/**
+	 * Test that category is in schema and marked as required.
+	 */
+	public function test_category_in_schema(): void {
```
gziolo: There is `test_get_schema` test in place. Let's move it there so it's easier to follow.

It might also help to assert the count of keys inside this existing `test_get_schema` test so it needs to get updated when new keys are added there helping to locate this test 😄

docs/7.registering-categories.md
galatanovidiu: I've removed the Hooks section

tests/unit/abilities-api/wpAbilityCategory.php
```diff
@@ -0,0 +1,768 @@
...
+		$this->doing_it_wrong_log = array();
+
+		add_action( 'doing_it_wrong_run', array( $this, 'record_doing_it_wrong' ), 10, 3 );
+	}
+
+	/**
+	 * Tear down after each test.
+	 */
+	public function tear_down(): void {
+		remove_action( 'doing_it_wrong_run', array( $this, 'record_doing_it_wrong' ), 10 );
```
gziolo: Nit, 10 is the defualt priority:
```suggestion
		remove_action( 'doing_it_wrong_run', array( $this, 'record_doing_it_wrong' ) );
```

More broadly, would it make sense to set this action once for the class with `setUpBeforeClass` and remove it once with `tearDownAfterClass` at the end?

includes/abilities-api.php
galatanovidiu: Good catch! Yes, `meta` should be included in the PHPStan annotation and the description. I've made the following updates:

**Changes made:**
- Added `meta?: array<string,mixed>` to the PHPStan type annotation (it's optional)
- Updated the docblock description to mention the `meta` parameter
- Fixed an issue where `label` and `description` were incorrectly marked as optional (`?`) when they're actually required fields

The PHPStan annotations now accurately reflect the implementation in `WP_Ability_Category::prepare_properties()`:
- `label` and `description` are **required** (throw exception if missing/empty)
- `meta` is **optional** (only validated if present)

These changes have been applied to both `includes/abilities-api.php` and `includes/abilities-api/class-wp-abilities-category-registry.php`.

Please check  2e2992c

docs/7.registering-categories.md
galatanovidiu: I've created a new Issue for it: https://github.com/WordPress/abilities-api/issues/110

tests/unit/abilities-api/wpAbilityCategory.php
```diff
@@ -0,0 +1,768 @@
...
+
+		$this->assertNull( $result );
+		$this->assertDoingItWrongTriggered( 'WP_Abilities_Category_Registry::register', 'already registered' );
+	}
+
+	/**
+	 * Test unregistering existing category.
+	 */
+	public function test_unregister_existing_category(): void {
+		$this->register_category_during_hook(
```
gziolo, line 280: So this is how you work around the enforced init hook during registration. That's fine, too 😄

tests/unit/abilities-api/wpAbilityCategory.php
```diff
@@ -0,0 +1,768 @@
...
+				array(
+					'label'       => 'System',
+					'description' => 'System operations.',
+				)
+			);
+		};
+
+		add_action( 'abilities_api_categories_init', $callback );
+		do_action( 'abilities_api_categories_init', WP_Abilities_Category_Registry::get_instance() );
+		remove_action( 'abilities_api_categories_init', $callback );
```
gziolo: This is precisely why I wanted to avoid checking for the init action in the registry itself.

tests/unit/abilities-api/wpAbilityCategory.php
```diff
@@ -0,0 +1,768 @@
...
+				},
+				'permission_callback' => '__return_true',
+			)
+		);
+
+		$this->assertInstanceOf( WP_Ability::class, $result );
+		$this->assertSame( 'test-math', $result->get_category() );
+
+		// Cleanup.
+		wp_unregister_ability( 'test/calculator' );
```
gziolo, line 449: Nit, I would do the cleanup before assertions to ensure it always happens in case assertions fail for some reason during development.
```suggestion
		// Cleanup.
		wp_unregister_ability( 'test/calculator' );

		$this->assertInstanceOf( WP_Ability::class, $result );
		$this->assertSame( 'test-math', $result->get_category() );
```

tests/unit/abilities-api/wpAbilityCategory.php
```diff
@@ -0,0 +1,768 @@
...
+	/**
+	 * Test category slug validation with valid formats.
+	 */
+	public function test_category_slug_valid_formats(): void {
+		$valid_slugs = array(
+			'test-simple',
+			'test-multiple-words',
+			'test-with-numbers-123',
+			'test-a',
+			'test-123',
```
gziolo: Nit: This is a perfect case for `@dataProvider`.

tests/unit/abilities-api/wpAbilityCategory.php
```diff
@@ -0,0 +1,768 @@
...
+		$invalid_slugs = array(
+			'Test-Uppercase',
+			'test_underscore',
+			'test.dot',
+			'test/slash',
+			'test space',
+			'-test-start-dash',
+			'test-end-dash-',
+			'test--double-dash',
+		);
```
gziolo: Nit: Similar perfect case for `@dataProvider`.

gziolo (approved):
I left a couple of additional nitpicks regarding the implementation of tests to consider before landing this PR.

**Overall, this is looking excellent functionality-wise from my perspective.** Let's make sure that other folks who left feedback are happy with the current shape and plan for merging. I would like to start the process of syncing the abilities registry and REST API layer to WordPress core, and this is the last missing piece of the puzzle I expected 🎉

includes/abilities-api/class-wp-abilities-category-registry.php
galatanovidiu: I agree that this is worth discussing for consistency.

There are two key differences in the approach I took for categories:

1. **Timing of the check**: For categories, I perform the `doing_action()` check directly in the `register()` method, whereas for abilities it's in the wrapper function `wp_register_ability()`.

2. **Enforcement level**: I use `doing_action()` to restrict category registration to the `abilities_api_categories_init` action only, while ability registration is more flexible using `did_action()`, which will let you register abilities even after the action.

I'll open a new issue so we can discuss these approaches and decide on the best path forward for consistency across the API.

galatanovidiu: Issue opened: https://github.com/WordPress/abilities-api/issues/112

tests/unit/abilities-api/wpAbilityCategory.php
galatanovidiu: Good catch on the priority parameter. Regarding class-level setup: we need per-test setup here because `setUpBeforeClass()` is static and can't access the instance method `record_doing_it_wrong()` or the instance property `$doing_it_wrong_log`.

includes/abilities-api/class-wp-abilities-category-registry.php
gziolo: Excellent, thank you very much. Let's decide in the issue 👍🏻

tests/unit/abilities-api/wpAbilityCategory.php
galatanovidiu: DOne please check [990e77a](https://github.com/WordPress/abilities-api/pull/102/commits/990e77aa17b34c8479d03baa7aba62a4ec676ee5)

tests/unit/abilities-api/wpAbilityCategory.php
galatanovidiu: DOne please check [990e77a](https://github.com/WordPress/abilities-api/pull/102/commits/990e77aa17b34c8479d03baa7aba62a4ec676ee5)

tests/unit/abilities-api/wpAbilityCategory.php
galatanovidiu: Good catch, I changed this one to use `register_category_during_hook()`

docs/7.registering-categories.md
gziolo: Moved the ordering aspect to a new issue [here](https://github.com/WordPress/abilities-api/issues/111#issuecomment-3397423261).

includes/abilities-api/class-wp-abilities-category-registry.php
JasonTheAdams: For consistency sake I'll agree. For readability, I'll kindly protest that `$registry->get_registered()` tells me nothing about what I'm getting. 😁

JasonTheAdams (approved):
Great work, @galatanovidiu! And great discussions, everyone! 😄

tests/unit/rest-api/wpRestAbilitiesListController.php
galatanovidiu: Done

includes/abilities-api/class-wp-abilities-registry.php
justlevine: Even though this PR was merged, I'd still appreciate some clarity here.

Issue #83, opened 2025-09-22
Roadmap to WP 6.9: defining and deciding


karmatosed:
This issue is meant to lead to the creation of a roadmap and clarification of the existing milestones. 

## Current milestones
There are 3 milestones so far and one is due the end of the month:

- v0.2.0: https://github.com/WordPress/abilities-api/milestone/3
- WP 6.9 Beta: https://github.com/WordPress/abilities-api/milestone/2
- Later: Anything after WP 6.9 but considered a priorty https://github.com/WordPress/abilities-api/milestone/4

## The roadmap to WP 6.9
Initially the roadmap is going to be heavily focused on 6.9, there can is a ‘later’ milestone though to use for anything outside.

### Triage and ensuring issues made
Whilst making sure the decisions are made, it’s a good time now to make sure everything to be decided for 6.9 has an issue and also if it doesn’t one is created. Along with this let’s review the milestones to be really sure what we have in them.

### Proposal to release to schedule
In order to make this easier for us to manage, I would propose that we have a two-week release cycle going up and during 6.9 or consider if we should release. If there are things to, we ship. This means we can rapidly bug fix in testing, it also means we can set some nice timed milestones and cadence for everyone.

Once WP 6.9 is locked in, we can stop this cadence or review if it indeed is working for everyone.

———————————

## Decisions needed
There are two outstanding decisions need discussion within the comments in order to ensure we have a full roadmap within milestones. They are:

- What core abilities to include: https://github.com/WordPress/ai/issues/40. 
- Client side capabilities. This has a PR: https://github.com/WordPress/abilities-api/pull/60. [ Decision: this will be included ]

Once those decisions are made they can be added to the milestones and issues updated if applicable.

## Next steps
Pinging a few people for input and then after this should have decisions by mid-week, to move to locking in milestones and focusing on shipping. Anyone is welcome to add input, but particularly pinging anyone that participated in the discussion during weekly sync.

@Jameswlepage @felixarntz @jeffpaul @swissspidy @gziolo @justlevine @galatanovidiu

annezazu:
>WP 6.2 Beta: https://github.com/WordPress/abilities-api/milestone/2
>Later: Anything after 6.2 but considered a priorty https://github.com/WordPress/abilities-api/milestone/4

Guessing you mean 6.9 here? :) Thanks for putting this together.

gziolo:
I updated the references from WP 6.2 to 6.9 😄

karmatosed:
[...]

Issue #38, opened 2025-08-22
Proposal: Add a convienient way to filter the list of all registered abilities


gziolo:
Figure out how to filter correctly all abilities when fetching them, given the expected high number of registered items by core and plugins.

Currently, it's only possible to find a single item or fetch all of them:

```php
// Retrieving all abilities.
$abilities = wp_get_abilities();

// Retrieving a specific ability.
$ability = wp_get_ability( 'example/add-numbers' );
```

The same applies to the REST API, which offers `WP_REST_Abilities_List_Controller` that handles listing abilities and retrieving individual abilities:

- GET `/wp-json/wp/v2/abilities` – list all registered abilities.
- GET `/wp-json/wp/v2/abilities/{name}` – Get a specific ability by name.

Let's discuss possible approaches. For reference, the WP Feature API prototype offers a special [`WP_Feature_Query`](https://github.com/Automattic/wp-feature-api/blob/trunk/includes/class-wp-feature-query.php) object that offers several filtering options:

https://github.com/Automattic/wp-feature-api/blob/82ff8af75ab607ec39dfbea3c1100807843922a3/includes/wp-feature.php#L51-L60

gziolo:
I assigned Copilot to explore whether it could be helpful for brainstorming. Initial proposal:
- https://github.com/WordPress/abilities-api/pull/63

galatanovidiu:
Thinking about [MCP adapter](https://github.com/WordPress/mcp-adapter) + abilities, it might be helpful to have an `expose_as_mcp` flag in ability metadata.

Proposal:
	- `false`: don’t expose at all on the default MCP server
	- '`tool`', '`resource`', '`prompt`': explicitly define how the ability should be exposed via MCP
	- The default can be '`tool`'

This gives us both a filter mechanism and a way to categorize abilities by type in the MCP adapter, which could help the adapter (or UIs) handle them appropriately.

--- 

For filtering abilities, we can use a simple filter. As an example:

```php
/**
 * Query abilities by meta criteria.
 *
 * Criteria supports:
 * - ['expose_as_mcp' => 'resource']                     // strict scalar equality
 * - ['expose_as_mcp' => ['tool','resource']]            // any-of (IN)
 * - ['audience' => ['internal','admin']]                // array intersect
 * - ['!expose_as_mcp' => false]                         // NOT equals
 * - ['?tags' => true]                                   // exists
 *
 * @param array      $criteria
[...]

PR #103, opened 2025-10-07
Add new `show_in_rest` arg to control REST API exposure

gziolo:
Noted in #83.

Introduces `show_in_rest` arg in the server-side registry in the case devs don't want their abilities to be discoverable in REST API endpoints. By default, abilities won't get exposed which aligns with many other fundamental pieces in WordPress like post types, taxonomies, meta, etc.

It also adds a new method on `WP_Ability` object `show_in_rest()` to allow checking whether the ability should be exposed through REST API.

Example

```php
// Ability that does not show in REST.
$ability = wp_register_ability(
	'test/not-show-in-rest',
	array(
		'label'               => 'Hidden from REST',
		'description'         => 'It does not show in REST.',
		'execute_callback'    => static function (): int {
			return 0;
		},
		'permission_callback' => '__return_true',
	)
);
// false === $ability->show_in_rest() 


// Ability that shows in REST.
$ability = wp_register_ability(
	'test/show-in-rest',
	array(
		'label'               => 'Show in REST',
		'description'         => 'It shows in REST.',
		'execute_callback'    => static function (): int {
			return 0;
		},
		'permission_callback' => '__return_true',
		'show_in_rest' => true,
	)
);
// true === $ability->show_in_rest() 
```

### Testing

Run `npm run test:php` to execute extended test coverage for the new functionality.

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -117,6 +117,14 @@ class WP_Ability {
 	protected $meta = array();
 
 	/**
+	 * Whether to show the ability in the REST API.
+	 *
+	 * @since n.e.x.t
+	 * @var bool
+	 */
+	protected $show_in_rest = false;
+
+	/**
 	 * Constructor.
 	 *
 	 * Do not use this constructor directly. Instead, use the `wp_register_ability()` function.
@@ -128,9 +136,9 @@ class WP_Ability {
 	 * @see wp_register_ability()
 	 *
 	 * @param string              $name The name of the ability, with its namespace.
-	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should
-	 *                                  include `label`, `description`, `input_schema`, `output_schema`,
-	 *                                  `execute_callback`, `permission_callback`, `annotations`, and `meta`.
+	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should include
+	 *                                  `label`, `description`, `input_schema`, `output_schema`, `execute_callback`,
+	 *                                  `permission_callback`, `annotations`, `meta`, and `show_in_rest`.
 	 */
 	public function __construct( string $name, array $args ) {
 		$this->name = $name;
@@ -180,6 +188,7 @@ class WP_Ability {
 	 *   output_schema?: array<string,mixed>,
 	 *   annotations?: array<string,mixed>,
 	 *   meta?: array<string,mixed>,
+	 *   show_in_rest?: bool,
 	 *   ...<string, mixed>,
 	 * } $args
 	 */
@@ -231,6 +240,12 @@ class WP_Ability {
 		if ( isset( $args['meta'] ) && ! is_array( $args['meta'] ) ) {
 			throw new \InvalidArgumentException(
 				esc_html__( 'The ability properties should provide a valid `meta` array.' )
+			);
+		}
+
+		if ( isset( $args['show_in_rest'] ) && ! is_bool( $args['show_in_rest'] ) ) {
+			throw new \InvalidArgumentException(
+				esc_html__( 'The ability properties should provide a valid `show_in_rest` boolean.' )
 			);
 		}
 
@@ -319,6 +334,17 @@ class WP_Ability {
 	 */
 	public function get_meta(): array {
 		return $this->meta;
+	}
+
+	/**
+	 * Checks whether the ability should be shown in the REST API.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool True if the ability should be shown in the REST API, false otherwise.
+	 */
+	public function has_show_in_rest(): bool {
+		return $this->show_in_rest;
 	}
 
 	/**
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -343,7 +343,7 @@ class WP_Ability {
 	 *
 	 * @return bool True if the ability should be shown in the REST API, false otherwise.
 	 */
-	public function has_show_in_rest(): bool {
+	public function show_in_rest(): bool {
 		return $this->show_in_rest;
 	}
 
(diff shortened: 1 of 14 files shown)


includes/abilities-api/class-wp-ability.php
```diff
@@ -266,6 +281,17 @@ public function get_meta(): array {
...
 	}
 
+	/**
+	 * Checks whether the ability should be shown in the REST API.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return bool True if the ability should be shown in the REST API, false otherwise.
+	 */
+	public function has_show_in_rest(): bool {
```
JasonTheAdams: I think the "has" is redundant in this case, as `$ability->show_in_rest()` is perfectly legible.
```suggestion
	public function show_in_rest(): bool {
```

tests/unit/abilities-api/wpAbilitiesRegistry.php
```diff
@@ -281,6 +282,21 @@ public function test_register_invalid_meta_type() {
...
+	 * @covers WP_Abilities_Registry::register
+	 * @covers WP_Ability::prepare_properties
+	 *
+	 * @expectedIncorrectUsage WP_Abilities_Registry::register
+	 */
+	public function test_register_invalid_show_in_rest_type() {
+		self::$test_ability_args['show_in_rest'] = 5;
+
+		$result = $this->registry->register( self::$test_ability_name, self::$test_ability_args );
+		$this->assertNull( $result );
```
JasonTheAdams, line 313: Looks like this should throw an `InvalidArgumentException`, right? Not just be null?

JasonTheAdams (changes requested):
Nice addition, @gziolo! One very minor suggestion.

includes/abilities-api/class-wp-ability.php
gziolo: Good point.

tests/unit/abilities-api/wpAbilitiesRegistry.php
gziolo: I think we catch it at this level, but it should throw in `WP_Ability` constructor. I will cover it accordingly, as we apparently don't have that part covered for other validated properties 👍🏻

gziolo: Addressed in[9f21064](https://github.com/WordPress/abilities-api/pull/103/commits/9f21064ca08f2ce50553f8fea8b0acd6948bd61c).

includes/abilities-api/class-wp-ability.php
gziolo: Addressed in[9f21064](https://github.com/WordPress/abilities-api/pull/103/commits/9f21064ca08f2ce50553f8fea8b0acd6948bd61c).

justlevine:
@gziolo what if we relocated this to our first official "meta" property: `$meta['show_in_rest']`?

This would allow us to drop the extra property + public getter + usage in favor of the existing `$ability->get_meta( 'show_in_rest' )`, keeping the api the same size while still supporting the same functionality. It also scales better for both known and unknown future requirements (e.g. `show_in_mcp`, [acp](https://agentclientprotocol.com/overview/introduction) and whatever else), and makes more sense from a semantic perspective (and therefore DX).

JasonTheAdams (approved):
Good work! LGTM!

JasonTheAdams:
I didn't see @justlevine's comment when I approved. Do we have a clear definition of what does and does not belong in meta?

gziolo:
@emdashcodes suggested in https://github.com/WordPress/abilities-api/issues/38#issuecomment-3382070226 that `show_in_rest` defaults to `true`. That’s the behavior before this PR, and since it’s a brand new concept I’m open to going with `true`. What other people think about it?

As for using `meta`, we didn’t arrive to conclusions but my thinking is it’s primarily for extenders to put additional details they need. When I put `show_in_rest` as top-level property I mirrored other concepts (post types, taxonomies, meta, options), but I don’t feel strongly about it. One thing important to keep in mind is we need a default value here and have it explicitly documented.

justlevine:
> @emdashcodes suggested in https://github.com/WordPress/abilities-api/issues/38#issuecomment-3382070226 that show_in_rest defaults to true. That’s the behavior before this PR, and since it’s a brand new concept I’m open to going with true. What other people think about it?

I'm good either way. Long term it's probably going to get inferred from a `public: bool` prop anyway, and ~since~ as long as  `permission_callback` is [for all intents and purposes](https://github.com/WordPress/abilities-api/pull/97#issuecomment-3362388352) required, there's not too much risk to defaulting to `true`

gziolo:
Based on the comment from @JasonTheAdams in https://github.com/WordPress/abilities-api/issues/106#issuecomment-3382733465, I will land this PR as is. However, I will immediately start working on a follow-up scoped to the following:
- Move `show_in_rest` to `meta.show_in_rest`.
- Update the default value to `true` as a separate commit, in case we decide to stay with `false`.

emdashcodes:
Thanks @gziolo! I feel pretty strongly that we should default to `true` for the same argument that we're making categories required. If it's set to `false`, most people aren't going to read the documentation and enable it, and a lot of these won't automatically be discoverable by the client. 

Having these opted-in by default allows for a lot of cool things like the command palette and automatically discovering them on the client and passing them to a client agent for example, and it would be nice if most of these were enabled "by default".

justlevine:
> Having these opted-in by default allows for a lot of cool things like the command palette

(My hunch is this too will more likely be encapsulated by a `show_in_cp` prop that also inherited from an eventual `public: bool` )

gziolo:
I have a follow-up ready for review based on my previous comment https://github.com/WordPress/abilities-api/pull/103#issuecomment-3384853695:
- https://github.com/WordPress/abilities-api/pull/107

gziolo:
Regarding Command Palette, the plan is to introduce the [Workflows concept](https://github.com/WordPress/gutenberg/issues/70710) as a facade for consuming Abilities. The point still holds that abilities need to be exposed on the client in the first place 👍🏻

PR #104, opened 2025-10-08
Move new `annotations` property under `meta` and update docs

gziolo:
Follow-up #99.

Addresses the feedback from https://github.com/WordPress/abilities-api/issues/106.

~Branch is currently based on https://github.com/WordPress/abilities-api/pull/107 to minimize the need for merge conflict resolution. Both PRs move properties to `meta`, so similar lines need to be updated.~

Refactoring ensures that `annotations` is no longer a top-level property in the `WP_Ability` class. Instead, it is now its group inside the `meta` property. This is similar to the previous changes applied to the `show_in_rest` property. As part of the refactoring `get_annotations()` public method got removed in favor of `get_meta_item( 'annotation' )` calls.

This PR expands on the documentation change necessary for understanding how the `annotation` property inside `meta` property on `WP_Ability` should be used.

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -110,14 +110,6 @@ class WP_Ability {
 	protected $permission_callback;
 
 	/**
-	 * The ability annotations.
-	 *
-	 * @since n.e.x.t
-	 * @var array<string,(bool|string)>
-	 */
-	protected $annotations = array();
-
-	/**
 	 * The optional ability metadata.
 	 *
 	 * @since 0.1.0
@@ -139,7 +131,7 @@ class WP_Ability {
 	 * @param string              $name The name of the ability, with its namespace.
 	 * @param array<string,mixed> $args An associative array of arguments for the ability. This should include
 	 *                                  `label`, `description`, `input_schema`, `output_schema`, `execute_callback`,
-	 *                                  `permission_callback`, `annotations`, and `meta`.
+	 *                                  `permission_callback`, and `meta`.
 	 */
 	public function __construct( string $name, array $args ) {
 		$this->name = $name;
@@ -187,8 +179,8 @@ class WP_Ability {
 	 *   permission_callback: callable( mixed $input= ): (bool|\WP_Error),
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   annotations?: array<string,(bool|string)>,
 	 *   meta?: array{
+	 *     annotations?: array<string,(bool|string)>,
 	 *     show_in_rest?: bool,
 	 *     ...<string, mixed>
 	 *   },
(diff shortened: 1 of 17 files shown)


JasonTheAdams:
As a quick note, I want to make sure that we're documenting the decision elsewhere that all annotations should be snake case.

gziolo:
Sure, snake case usage is a coding style convention in WordPress and we discussed that yesterday with shared understanding it’s not always convenient. I will link later to the thread.

gziolo:
Depends now on https://github.com/WordPress/abilities-api/issues/106.

gziolo:
I temporarily based this PR on https://github.com/WordPress/abilities-api/pull/107, and extended its scope based on the feedback from https://github.com/WordPress/abilities-api/issues/106.

includes/abilities-api/class-wp-ability.php
JasonTheAdams, line 1: What is the typical WordPress approach to redundancy? I personally wouldn't mind having a getter and setter for annotations, while still storing them in meta. Yes, one could use `get_meta()` to retrieve the annotations, but for built-in meta it's nice to have more declarative methods.

JasonTheAdams (approved):
Thanks for doing this, @gziolo! I left one question, but for me it's not a blocker to merging, so I'll go ahead and approve!

includes/abilities-api/class-wp-ability.php
gziolo: I’m very much open for adding a getter to simplify access to individual annotations. I don’t think WordPress core has many good reference points in that regard because older code is often composed of functions chained together. I can spin a follow up PR. To confirm we are on the same page, are we talking about something like:

```php
$this->get_annotation( ‘readonly’ );
```

JasonTheAdams: Yup! Exactly. Just convenience methods.

justlevine: If possible, I'd like to save convenience methods for 7.0. 

I know nested arrays are messy so I see why we might want to make an exception for annotations (at least in their [current shape](https://github.com/WordPress/abilities-api/issues/62#issuecomment-3385606922)), but FWIW @gziolo did just introduce `::get_meta_item()` in #107, so the method would only be at the convenience of `$ability->get_meta_item( 'annotations' )['readonly']`

JasonTheAdams: I don't see a reason to hesitate on these things unless you foresee something like annotations going away or moving.

justlevine: @JasonTheAdams more details in the comment and subsequent discussion I linked to above in #62 but tl;dr I expect at minimum "annotation" to be renamed to something that makes sense semanticaly in non-MCP contexts, as well as the shape/location/keys for some of the meta based on what gets decided in #106 . Hopefully in time for 6.9 so it too doesn't become cumulative tech debt.

But also if I  couldn't think of ways to possibly improve the semantics/dx, I'd hope the fact that we're not dogfooding the function internally would be reason enough to pause. We're already navelgazing these props into existence, it's counterproductive to keep expanding our API with compounding layers of conjecture, and slows down our ability to iterate since once something is in core we need to work around it forever.

(Like the linked comment it's not a big deal if we add it now in time for beta1 and then remove during core review. But I do feel it's pretty likely this method will be obviated before rc )

Issue #106, opened 2025-10-08
Determine what goes in Ability meta


JasonTheAdams:
The `wp_register_ability()` function has an `array $args` parameter. This includes data that is required and optional. It also has an `$args['meta']` array. The entire `$args` shape is part of the function signature, so changing it _is_ changing the signature of the function. As such, we should be clear and have a rule for what should be a root element of `$args`, versus what should belong in `$args['meta']`.

I propose these rules:
* Required data goes in `$args` root (e.g. $args['label'])
* Optional data goes in `$meta` which itself is optional

This is consistent with the [parameters as documented](https://github.com/WordPress/abilities-api/blob/trunk/docs/3.registering-abilities.md#parameters-explained).

As such, this would effect things like annotations and `show_in_rest`, which will become meta. As a note, meta is also the encouraged location for any further arbitrary data extended by developers — which will also be optional at least as far as registration is concerned.

JasonTheAdams:
We should make a decision on this quickly so we can document this and make any necessary changes. For PRs such as #103, I think we should just merge that PR and then make a subsequent PR for moving anything that's currently in the root which we want to move to meta.

gziolo:
> As such, this would effect things like annotations and show_in_rest, which will become meta.

Would you suggest merging meta with annotations or rather moving annotations as a group under meta?

Overall, I’m happy to make a final decision so we have clear path for future additions.

justlevine:
> I propose these rules:
> 
> * Required data goes in `$args` root (e.g. $args['label'])
> * Optional data goes in `$meta` which itself is optional

Not sure this is the best delineation since it just kicks the question to "what data should be required".

- Per this definition, it's not just `annotations` or `show_in_rest` that get relocated, but also the `input_schema`, `output_schema`, and `ability_class` that should be moved. To me that placement feels counterintuitive. (The documentation in https://github.com/WordPress/abilities-api/blob/trunk/docs/3.registering-abilities.md#parameters-explained seems out of date).
- `ability_class` technically makes _all_ properties optional, so the [polymorphic footgun](https://github.com/WordPress/abilities-api/pull/97#issuecomment-3371133652) now also affects DX semantics.

I propose the following rules instead:

* `$args`: Data required _to identify and execute the ability_ goes in  root
* Everything else is "additional metadata about the ability" and belongs in the optional `$args['meta']`.

This is imo more consistent with the current docs/implementation and also more intuitive:

| Element | Location | Reason |
|--------|------|------|
[...]

PR #107, opened 2025-10-09
Move `show_in_rest` to meta in the registration process

gziolo:
- Part of https://github.com/WordPress/abilities-api/issues/106

Move `show_in_rest` to `meta` in the registration process.

Example:

```php
// Ability that does not show in REST.
wp_register_ability(
	'test/not-show-in-rest',
	array(
		'label'               => 'Hidden from REST',
		'description'         => 'It does not show in REST.',
		'execute_callback'    => static function (): int {
			return 0;
		},
		'permission_callback' => '__return_true',
		'meta'                => array(
			'show_in_rest' => false,
		),
	)
);
```

~I also included the proposal in https://github.com/WordPress/abilities-api/pull/107/commits/8f937a66a66a7dc1f06792427c2c449570333e30 to change the default value to `true`, so all abilities are exposed in REST API by default.~ The decision was made in https://github.com/WordPress/abilities-api/pull/107#discussion_r2417261219 to keep `show_in_rest` defaulting to `false`.

I also replaced `show_in_rest()` method with a more generalized helper `get_meta_item()`. Example usage:

```php
$ability->get_meta_item( 'non_existing' ); // returns null
$ability->get_meta_item( 'non_existing', 'default_value' ); // returns 'default_value'
$ability->get_meta_item( 'show_in_rest' ); // returns false
```

### DIFF
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -25,7 +25,7 @@ declare( strict_types = 1 );
  *                                  alphanumeric characters, dashes and the forward slash.
  * @param array<string,mixed> $args An associative array of arguments for the ability. This should include
  *                                  `label`, `description`, `input_schema`, `output_schema`, `execute_callback`,
- *                                  `permission_callback`, `annotations`, `meta`, `show_in_rest`, and `ability_class`.
+ *                                  `permission_callback`, `annotations`, `meta`, and `ability_class`.
  * @return ?\WP_Ability An instance of registered ability on success, null on failure.
  *
  * @phpstan-param array{
@@ -37,7 +37,6 @@ declare( strict_types = 1 );
  *   output_schema?: array<string,mixed>,
  *   annotations?: array<string,mixed>,
  *   meta?: array<string,mixed>,
- *   show_in_rest?: bool,
  *   ability_class?: class-string<\WP_Ability>,
  *   ...<string, mixed>
  * } $args
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -35,8 +35,11 @@ declare( strict_types = 1 );
  *   permission_callback?: callable( mixed $input= ): (bool|\WP_Error),
  *   input_schema?: array<string,mixed>,
  *   output_schema?: array<string,mixed>,
- *   annotations?: array<string,mixed>,
- *   meta?: array<string,mixed>,
+ *   annotations?: array<string,(bool|string)>,
+ *   meta?: array{
+ *     show_in_rest?: bool,
+ *     ...<string,mixed>,
+ *   },
  *   ability_class?: class-string<\WP_Ability>,
  *   ...<string, mixed>
  * } $args
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -57,7 +57,7 @@ final class WP_Abilities_Registry {
 	 *   permission_callback?: callable( mixed $input= ): (bool|\WP_Error),
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   annotations?: array<string,mixed>,
+	 *   annotations?: array<string,(bool|string)>,
 	 *   meta?: array{
 	 *     show_in_rest?: bool,
 	 *     ...<string, mixed>
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -59,7 +59,6 @@ final class WP_Abilities_Registry {
 	 *   output_schema?: array<string,mixed>,
 	 *   annotations?: array<string,mixed>,
 	 *   meta?: array<string,mixed>,
-	 *   show_in_rest?: bool,
 	 *   ability_class?: class-string<\WP_Ability>,
 	 *   ...<string, mixed>
 	 * } $args
--- a/includes/abilities-api/class-wp-abilities-registry.php
+++ b/includes/abilities-api/class-wp-abilities-registry.php
@@ -58,7 +58,10 @@ final class WP_Abilities_Registry {
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
 	 *   annotations?: array<string,mixed>,
-	 *   meta?: array<string,mixed>,
+	 *   meta?: array{
+	 *     show_in_rest?: bool,
+	 *     ...<string, mixed>
+	 *   },
 	 *   ability_class?: class-string<\WP_Ability>,
 	 *   ...<string, mixed>
 	 * } $args
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -123,7 +123,7 @@ class WP_Ability {
 	 * @since 0.1.0
 	 * @var array<string,mixed>
 	 */
-	protected $meta = array();
+	protected $meta;
 
 	/**
 	 * Constructor.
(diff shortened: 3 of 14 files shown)


includes/abilities-api/class-wp-ability.php
```diff
@@ -19,6 +19,15 @@
...
  */
 class WP_Ability {
+
+	/**
+	 * The default value for the `show_in_rest` meta.
+	 *
+	 * @since n.e.x.t
+	 * @var bool
+	 */
+	protected const DEFAULT_SHOW_IN_REST = true;
```
JasonTheAdams: I'm surprised this is true by default. This would mean an ability is public by default, which I'm not sure is a good idea.

JasonTheAdams (commented):
Looking mostly good, just one question on the default, @gziolo.

includes/abilities-api/class-wp-ability.php
gziolo: See my note in the description:
> I also included the proposal in https://github.com/WordPress/abilities-api/commit/8f937a66a66a7dc1f06792427c2c449570333e30 to change the default value to true, so all abilities are exposed in REST API by default.

That was feedback shared by @emdashcodes. It’s how it works currently in v0.2.0.

gziolo: It largely depends on what we are comfortable to promote from start. In particular with mandatory `permission_check`, it should be safe. 

Aside, in Gutenberg many things have to be explicitly marked to show in REST. To give an example, very popular feedback was how to tweak custom fields registration so it is exposed by default.

gziolo: It’s a separate commit also so we can revert fast if we agree to stay with `false`.

JasonTheAdams: Given that abilities aren't presented as a REST-first feature, my inclination would be to have it be false by default. I don't think it's clear to someone, otherwise, that by virtue of making an ability they're also exposing it via REST.

felixarntz: I agree with @JasonTheAdams to have abilities not included in the REST API by default.

Enabling it by default assumes that every ability is useful in a REST context, but the code that interferes with REST and the code that interferes with abilities (e.g. MCP or a WordPress-internal agent) have very different implications and needs. For example, if there's an ability to create a post, that IMO doesn't make sense to have in the REST API because there's already a proper controller for that which is likely more suitable and certainly more RESTful than an ability implementation would likely be.

To be clear, I'm not against allowing abilities in the REST API, but I think that should be the exception not the norm, it should be enabled based on concrete needs for each given ability.

gziolo: @Jameswlepage, you might have better overview what’s been discussed initially in that regard that might help to make the decision. We need to settle on something here.

gziolo: I’m gravitating towards `false` as the default value based on the feedback received in this thread. I will hold off until tomorrow with reverting the commit changing the default to `true` (exposed through REST API) in case there is more feedback coming.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -59,7 +59,6 @@ final class WP_Abilities_Registry {
 	 *   output_schema?: array<string,mixed>,
 	 *   annotations?: array<string,mixed>,
 	 *   meta?: array<string,mixed>,
```
justlevine:
```suggestion
	 *   meta?: array{
	 *     show_in_rest?: bool,
	 *     ...<string, mixed>
	 *   },
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -116,14 +125,6 @@ class WP_Ability {
 	 */
 	protected $meta = array();
```
justlevine: We shouldn't need to give this (or `::$annotations` ) a value anymore since `::prepare_properties` always instantiates it.

includes/abilities-api/class-wp-ability.php
```diff
@@ -344,7 +350,7 @@ public function get_meta(): array {
 	 * @return bool True if the ability should be shown in the REST API, false otherwise.
 	 */
 	public function show_in_rest(): bool {
-		return $this->show_in_rest;
+		return $this->meta['show_in_rest'] ?? self::DEFAULT_SHOW_IN_REST;
```
justlevine: 1. Is defaulting here necessary, when `wp_parse_args()` ensures that `$this->meta['show_in_rest']` is set?
2. Is this `public function` necessary, when `$ability->get_meta()['show_in_rest']` will always be set and bool?

includes/abilities-api/class-wp-ability.php
```diff
@@ -188,7 +189,6 @@ public function __construct( string $name, array $args ) {
 	 *   output_schema?: array<string,mixed>,
 	 *   annotations?: array<string,mixed>,
```
justlevine: This should probably also be changed inside WP_Abilities_Registry

```suggestion
	 *   annotations?: array<string,(bool|string)>,
```

includes/abilities-api.php
```diff
@@ -37,7 +37,6 @@
  *   output_schema?: array<string,mixed>,
  *   annotations?: array<string,mixed>,
  *   meta?: array<string,mixed>,
```
justlevine:
```suggestion
 *   annotations?: array<string,(bool|string)>,
 *   meta?: array{
 *     show_in_rest?: bool,
 *     ...<string,mixed>,
 *   },
```

includes/abilities-api/class-wp-ability.php
gziolo: Commit removed, the default for `meta.show_in_rest` restored to `false`.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -58,8 +58,10 @@ final class WP_Abilities_Registry {
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
 	 *   annotations?: array<string,mixed>,
```
gziolo:
```suggestion
	 *   annotations?: array<string,(bool|string)>,
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -186,9 +187,8 @@ public function __construct( string $name, array $args ) {
 	 *   permission_callback: callable( mixed $input= ): (bool|\WP_Error),
 	 *   input_schema?: array<string,mixed>,
 	 *   output_schema?: array<string,mixed>,
-	 *   annotations?: array<string,mixed>,
+	 *   annotations?: array<string,(bool|string)>,
 	 *   meta?: array<string,mixed>,
```
gziolo:
```suggestion
	 *   meta?: array{
	 *     show_in_rest?: bool,
	 *     ...<string, mixed>
	 *   },
```

includes/abilities-api/class-wp-ability.php
gziolo:
```suggestion
	protected $meta;
```

includes/abilities-api/class-wp-ability.php
```diff
@@ -344,7 +353,7 @@ public function get_meta(): array {
 	 * @return bool True if the ability should be shown in the REST API, false otherwise.
 	 */
 	public function show_in_rest(): bool {
-		return $this->show_in_rest;
+		return $this->meta['show_in_rest'] ?? false;
```
gziolo:
```suggestion
		return $this->meta['show_in_rest'];
```

includes/abilities-api/class-wp-ability.php
gziolo: Feedback addressed in https://github.com/WordPress/abilities-api/pull/107/commits/fee5c3c8dd51805c82fbd061949db50c010b0d5a. I introduced `get_meta_item( 'show_in_rest' )` helper replacing too specific `show_in_rest()`.

gziolo:
The decision was made in https://github.com/WordPress/abilities-api/pull/107#discussion_r2417261219 to keep `show_in_rest` defaulting to false. I addressed all other feedback, too.

justlevine (approved):

Issue #30, opened 2025-08-19
Proposal: Add common instruction files for AI assisted coding tools


jonathanbossenger:
This is a proposal to add support for the most commonly used AI Coding tools.

- [GitHub Copilot](https://docs.github.com/en/copilot/how-tos/configure-custom-instructions/add-repository-instructions?tool=vscode)
- [Cursor](https://docs.cursor.com/en/context/rules)
- [Claude Code](https://docs.anthropic.com/en/docs/claude-code/memory)

This will allow developers who use these tools to ramp up when contributing to this project quickly.

gziolo:
Is there any prior art in existing WordPress repositories that we could use for inspiration?

jonathanbossenger:
Not that I'm aware of. We might pioneer the space.

gziolo:
I was using locally GitHub Copilot code agent this week, and it was able to quickly sort out how to run tests, linting and fix issues after some hints I shared. However, it would be great to have that stored in instructions so these tools can discover it automatically 👍🏻

justlevine:
This makes me _very_ nervous. 

I need to delete WPGraphQL's (canonical but not in `WordPress/*`) [files](https://github.com/wp-graphql/wp-graphql/tree/develop/.cursor) before I contribute because they produce worse results than no files at all. 

These files are immediate tech debt, notoriously temperamental, and even if you were to theoretically tweak and test them per model/code-tool release (or unversioned background changes) what's the point? Currently they're not much more effective than pointing the agent to the actual docs. And unlike vibe-coded slop or external plugins, this is a WordPress _core_ project. We _have_ docs, coding standards, tools to enforce, style guides.

If we want to pioneer the space, then perhaps experimenting, ideating and publicly sharing personal instruction files is a better way to start

(Also:

<img width="768" alt="Image" src="https://github.com/user-attachments/assets/68f35a83-4036-4c0f-aaca-56b2c020cecc" />

Source: https://x.com/shadcn/status/1955254151807635602
)

jonathanbossenger:
> I need to delete WPGraphQL's (canonical but not in WordPress/*) [files](https://github.com/wp-graphql/wp-graphql/tree/develop/.cursor) before I contribute because they produce worse results than no files at all.

[...]

PR #108, opened 2025-10-10
Add initial core abilities for WordPress 6.9

### LINKED ISSUE #105: Core Abilities for WordPress 6.9
This issue tracks the initial set of Abilities to ship in WordPress 6.9, based on the broader discussion in WordPress/ai#40.

The goal for 6.9 is to deliver a minimal, stable foundation of Abilities in core, while continuing to develop the broader catalog in the Abilities API plugin for future merges.

## Todo

- [x] Decide which core abilities we're including for 6.9
- [x] Ship them

Jameswlepage:
## Summary

Implements 4 foundational core abilities to provide a stable foundation for WordPress 6.9, addressing issue #105.

### Abilities Included

1. **`core/get-bloginfo`** - Retrieve individual site information fields (name, description, url, wpurl, admin_email, charset, language, version)
2. **`core/get-current-user-info`** - Get current authenticated user data (id, display_name, user_nicename, user_login, roles, locale)
3. **`core/get-environment-type`** - Get WordPress environment type (production, staging, development, local)
4. **`core/find-abilities`** - Discover available abilities with optional namespace and show_in_rest filtering

### Design Philosophy

These abilities were selected to provide essential foundational operations while following atomic design principles:

- **Single-purpose operations**: Each ability performs one clearly-defined task (e.g., `core/get-bloginfo` retrieves one field at a time rather than bulk data)
- **Composability**: Atomic operations allow AI agents and clients to request exactly what they need
- **Minimal surface area**: Focused on read-only discovery and introspection to establish a stable foundation
- **WordPress alignment**: Mirrors existing WordPress API patterns (get_bloginfo, wp_get_current_user, etc.)

The atomic approach keeps the initial release simple and predictable, with room to add bulk or convenience operations in future releases based on real-world usage patterns.

### Implementation Details

- All abilities use the new `meta` structure (PRs #107, #104)
- `annotations` and `show_in_rest` properly nested inside `meta`
- All abilities marked as `readonly: true`, `destructive: false`, `idempotent: true`
- Comprehensive test coverage (6 new tests, 149 total tests passing)
- REST API fully functional and tested with authentication
- Abilities are registered via `abilities_api_init` hook
- Can be disabled in test environments or controlled via `abilities_api_register_core_abilities` filter

### Testing

All 149 PHPUnit tests pass (345 assertions)
REST API endpoints validated with authentication
All 8 bloginfo fields tested and working
Input validation working correctly
Permission callbacks enforced properly

### Architecture Decisions

- **Single-field design for `core/get-bloginfo`**: Keeps abilities atomic and composable, following Unix philosophy and WordPress REST API patterns
- **Authentication required for `core/find-abilities`**: Requires `read` capability to prevent anonymous enumeration while still accessible to all authenticated users

## Addresses

Closes #105

## Checklist

- [ ] Tests added and passing
- [ ] Code follows WordPress coding standards
- [ ] Documentation updated in code
- [x] REST API tested with authentication
- [x] Rebased on latest trunk
- [x] Uses new meta structure from PRs #107 and #104

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -321,6 +321,27 @@ class WP_Ability {
 	}
 
 	/**
+	 * Applies the defined input default when no input is provided.
+	 *
+	 * @since 0.4.0
+	 *
+	 * @param mixed $input Optional. The raw input provided for the ability. Default `null`.
+	 * @return mixed The input with the schema default applied when available.
+	 */
+	public function normalize_input( $input = null ) {
+		if ( null !== $input ) {
+			return $input;
+		}
+
+		$input_schema = $this->get_input_schema();
+		if ( ! empty( $input_schema ) && array_key_exists( 'default', $input_schema ) ) {
+			return $input_schema['default'];
+		}
+
+		return null;
+	}
+
+	/**
 	 * Retrieves the output schema for the ability.
 	 *
 	 * @since 0.1.0
@@ -436,6 +457,7 @@ class WP_Ability {
 	 * @return bool|\WP_Error Whether the ability has the necessary permission.
 	 */
 	public function check_permissions( $input = null ) {
+		$input    = $this->normalize_input( $input );
 		$is_valid = $this->validate_input( $input );
 		if ( is_wp_error( $is_valid ) ) {
 			return $is_valid;
@@ -522,6 +544,7 @@ class WP_Ability {
 	 * @return mixed|\WP_Error The result of the ability execution, or WP_Error on failure.
 	 */
 	public function execute( $input = null ) {
+		$input           = $this->normalize_input( $input );
 		$has_permissions = $this->check_permissions( $input );
 		if ( true !== $has_permissions ) {
 			if ( is_wp_error( $has_permissions ) ) {
--- a/includes/abilities/class-wp-core-abilities.php
+++ b/includes/abilities/class-wp-core-abilities.php
@@ -140,7 +140,7 @@ final class WP_Core_Abilities {
 	}
 
 	/**
-	 * Registers the `core/get-current-user-info` ability.
+	 * Registers the `core/get-user-info` ability.
 	 *
 	 * @since 0.3.0
 	 *
@@ -148,9 +148,9 @@ final class WP_Core_Abilities {
 	 */
 	protected static function register_get_current_user_info(): void {
 		wp_register_ability(
-			'core/get-current-user-info',
+			'core/get-user-info',
 			array(
-				'label'               => __( 'Get Current User Information' ),
+				'label'               => __( 'Get User Information' ),
 				'description'         => __( 'Returns basic profile details for the current authenticated user to support personalization, auditing, and access-aware behavior.' ),
 				'category'            => self::CATEGORY_USER,
 				'output_schema'       => array(
--- a/includes/abilities/class-wp-core-abilities.php
+++ b/includes/abilities/class-wp-core-abilities.php
@@ -1,0 +1,345 @@
+<?php
+/**
+ * Core Abilities registration.
+ *
+ * @package WordPress
+ * @subpackage Abilities_API
+ * @since n.e.x.t
+ */
+
+declare( strict_types = 1 );
+
+/**
+ * Registers the default core abilities that ship with the Abilities API.
+ *
+ * @since n.e.x.t
+ */
+class WP_Core_Abilities {
+	/**
+	 * Registers the default core abilities.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
+	 */
+	public static function register(): void {
+		self::register_get_bloginfo();
+		self::register_get_current_user_info();
+		self::register_get_environment_type();
+		self::register_find_abilities();
+	}
+
+	/**
+	 * Registers the `core/get-bloginfo` ability.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
+	 */
+	protected static function register_get_bloginfo(): void {
+		$fields = array(
+			'name',
+			'description',
+			'url',
+			'wpurl',
+			'admin_email',
+			'charset',
+			'language',
+			'version',
+		);
+
+		wp_register_ability(
+			'core/get-bloginfo',
+			array(
+				'label'               => __( 'Get Blog Information', 'abilities-api' ),
+				'description'         => __( 'Returns a single site information field from get_bloginfo().', 'abilities-api' ),
+				'input_schema'        => array(
+					'type'                 => 'object',
+					'properties'           => array(
+						'field' => array(
+							'type'        => 'string',
+							'enum'        => $fields,
+							'description' => __( 'The site information field to retrieve.', 'abilities-api' ),
+						),
+					),
+					'required'             => array( 'field' ),
+					'additionalProperties' => false,
+				),
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'required'             => array( 'field', 'value' ),
+					'properties'           => array(
+						'field' => array(
+							'type'        => 'string',
+							'description' => __( 'The requested site information field.', 'abilities-api' ),
+						),
+						'value' => array(
+							'type'        => 'string',
+							'description' => __( 'The value returned by get_bloginfo().', 'abilities-api' ),
+						),
+					),
+					'additionalProperties' => false,
+				),
+				'execute_callback'    => static function ( array $input ): array {
+					$field = $input['field'];
+					$value = get_bloginfo( $field );
+
+					return array(
+						'field' => $field,
+						'value' => (string) $value,
+					);
+				},
+				'permission_callback' => '__return_true',
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Retrieves a single site property by passing an allowed field to get_bloginfo().', 'abilities-api' ),
+						'readonly'     => true,
+						'destructive'  => false,
+						'idempotent'   => true,
+					),
+					'show_in_rest' => true,
+				),
+			)
+		);
+	}
+
+	/**
+	 * Registers the `core/get-current-user-info` ability.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
+	 */
+	protected static function register_get_current_user_info(): void {
+		wp_register_ability(
+			'core/get-current-user-info',
+			array(
+				'label'               => __( 'Get Current User Information', 'abilities-api' ),
+				'description'         => __( 'Returns basic information about the current authenticated user.', 'abilities-api' ),
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'required'             => array( 'id', 'display_name', 'locale' ),
+					'properties'           => array(
+						'id'           => array(
+							'type'        => 'integer',
+							'description' => __( 'The user ID.', 'abilities-api' ),
+						),
+						'display_name' => array(
+							'type'        => 'string',
+							'description' => __( 'The display name of the user.', 'abilities-api' ),
+						),
+						'user_nicename' => array(
+							'type'        => 'string',
+							'description' => __( 'The URL-friendly name for the user.', 'abilities-api' ),
+						),
+						'user_login'   => array(
+							'type'        => 'string',
+							'description' => __( 'The login username for the user.', 'abilities-api' ),
+						),
+						'roles'        => array(
+							'type'        => 'array',
+							'description' => __( 'The roles assigned to the user.', 'abilities-api' ),
+							'items'       => array(
+								'type' => 'string',
+							),
+						),
+						'locale' => array(
+							'type'        => 'string',
+							'description' => __( 'The locale string for the user, such as en_US.', 'abilities-api' ),
+						),
+					),
+					'additionalProperties' => false,
+				),
+				'execute_callback'    => static function (): array {
+					$current_user = wp_get_current_user();
+
+					return array(
+						'id'           => $current_user->ID,
+						'display_name' => $current_user->display_name,
+						'user_nicename' => $current_user->user_nicename,
+						'user_login'   => $current_user->user_login,
+						'roles'        => $current_user->roles,
+						'locale'       => get_user_locale( $current_user ),
+					);
+				},
+				'permission_callback' => static function (): bool {
+					return is_user_logged_in();
+				},
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Retrieves information about the current authenticated user.', 'abilities-api' ),
+						'readonly'     => true,
+						'destructive'  => false,
+						'idempotent'   => true,
+					),
+					'show_in_rest' => true,
+				),
+			)
+		);
+	}
+
+	/**
+	 * Registers the `core/get-environment-type` ability.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
+	 */
+	protected static function register_get_environment_type(): void {
+		wp_register_ability(
+			'core/get-environment-type',
+			array(
+				'label'               => __( 'Get Environment Type', 'abilities-api' ),
+				'description'         => __( 'Returns the current WordPress environment type (e.g. production or staging).', 'abilities-api' ),
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'required'             => array( 'environment' ),
+					'properties'           => array(
+						'environment' => array(
+							'type'        => 'string',
+							'description' => __( 'The environment type returned by wp_get_environment_type().', 'abilities-api' ),
+						),
+					),
+					'additionalProperties' => false,
+				),
+				'execute_callback'    => static function (): array {
+					return array(
+						'environment' => wp_get_environment_type(),
+					);
+				},
+				'permission_callback' => '__return_true',
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Retrieves the current WordPress environment type.', 'abilities-api' ),
+						'readonly'     => true,
+						'destructive'  => false,
+						'idempotent'   => true,
+					),
+					'show_in_rest' => true,
+				),
+			)
+		);
+	}
+
+	/**
+	 * Registers the `core/find-abilities` ability.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @return void
+	 */
+	protected static function register_find_abilities(): void {
+		wp_register_ability(
+			'core/find-abilities',
+			array(
+				'label'               => __( 'Find Abilities', 'abilities-api' ),
+				'description'         => __( 'Returns a list of abilities that are exposed through the registry.', 'abilities-api' ),
+				'input_schema'        => array(
+					'type'                 => array( 'object', 'null' ),
+					'properties'           => array(
+						'namespace' => array(
+							'type'        => 'string',
+							'description' => __( 'Optional namespace prefix to filter abilities (e.g. "core/").', 'abilities-api' ),
+						),
+						'show_in_rest' => array(
+							'type'        => 'boolean',
+							'description' => __( 'Whether to limit results to abilities exposed in REST. Defaults to true.', 'abilities-api' ),
+						),
+					),
+					'additionalProperties' => false,
+				),
+				'output_schema'       => array(
+					'type'       => 'object',
+					'properties' => array(
+						'abilities' => array(
+							'type'  => 'array',
+							'items' => array(
+								'type'       => 'object',
+								'properties' => array(
+									'name'        => array(
+										'type'        => 'string',
+										'description' => __( 'The ability name.', 'abilities-api' ),
+									),
+									'label'       => array(
+										'type'        => 'string',
+										'description' => __( 'The human readable label.', 'abilities-api' ),
+									),
+									'description' => array(
+										'type'        => 'string',
+										'description' => __( 'The detailed description.', 'abilities-api' ),
+									),
+									'meta'        => array(
+										'type'        => 'object',
+										'description' => __( 'Additional metadata associated with the ability.', 'abilities-api' ),
+									),
+									'annotations' => array(
+										'type'        => 'object',
+										'description' => __( 'Annotations describing ability behavior.', 'abilities-api' ),
+									),
+									'show_in_rest' => array(
+										'type'        => 'boolean',
+										'description' => __( 'Whether the ability is exposed in REST.', 'abilities-api' ),
+									),
+								),
+								'required'   => array( 'name', 'label', 'description', 'meta', 'annotations', 'show_in_rest' ),
+							),
+						),
+					),
+					'additionalProperties' => false,
+				),
+				'execute_callback'    => static function ( array $input = array() ): array {
+					$namespace     = $input['namespace'] ?? null;
+					$filter_rest   = array_key_exists( 'show_in_rest', $input ) ? (bool) $input['show_in_rest'] : true;
+					$abilities     = wp_get_abilities();
+					$filtered_list = array();
+
+					foreach ( $abilities as $ability ) {
+						$show_in_rest = $ability->get_meta_item( 'show_in_rest', false );
+						if ( $filter_rest && ! $show_in_rest ) {
+							continue;
+						}
+
+						if ( $namespace && ! str_starts_with( $ability->get_name(), $namespace ) ) {
+							continue;
+						}
+
+						$filtered_list[] = array(
+							'name'         => $ability->get_name(),
+							'label'        => $ability->get_label(),
+							'description'  => $ability->get_description(),
+							'meta'         => $ability->get_meta(),
+							'annotations'  => $ability->get_meta_item( 'annotations', array() ),
+							'show_in_rest' => $show_in_rest,
+						);
+					}
+
+					/**
+					 * Filters the abilities returned by the `core/find-abilities` ability.
+					 *
+					 * @since n.e.x.t
+					 *
+					 * @param array<string,mixed>[] $abilities An array of abilities exposed by the ability.
+					 * @param array<string,mixed>   $input      The input arguments passed to the ability.
+					 */
+					$filtered_list = apply_filters( 'abilities_api_core_find_abilities_results', $filtered_list, $input );
+
+					return array(
+						'abilities' => array_values( $filtered_list ),
+					);
+				},
+				'permission_callback' => static function (): bool {
+					return current_user_can( 'read' );
+				},
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Lists abilities from the registry. Optional namespace filter is supported.', 'abilities-api' ),
+						'readonly'     => true,
+						'destructive'  => false,
+						'idempotent'   => true,
+					),
+					'show_in_rest' => true,
+				),
+			)
+		);
+	}
+}
--- a/includes/abilities/class-wp-core-abilities.php
+++ b/includes/abilities/class-wp-core-abilities.php
@@ -117,7 +117,7 @@ class WP_Core_Abilities {
 						'value' => (string) $value,
 					);
 				},
-				'permission_callback' => static function ( $input = array() ): bool {
+				'permission_callback' => static function (): bool {
 					// Site information can expose sensitive details; require admin capability.
 					return current_user_can( 'manage_options' );
 				},
--- a/includes/abilities/class-wp-core-abilities.php
+++ b/includes/abilities/class-wp-core-abilities.php
@@ -150,7 +150,7 @@ class WP_Core_Abilities {
 				'category'            => 'user',
 				'output_schema'       => array(
 					'type'                 => 'object',
-					'required'             => array( 'id', 'display_name', 'locale' ),
+					'required'             => array( 'id', 'display_name', 'user_nicename', 'user_login', 'roles', 'locale' ),
 					'properties'           => array(
 						'id'            => array(
 							'type'        => 'integer',
--- a/includes/abilities/class-wp-core-abilities.php
+++ b/includes/abilities/class-wp-core-abilities.php
@@ -53,7 +53,7 @@ final class WP_Core_Abilities {
 	/**
 	 * Registers the default core abilities.
 	 *
- 	* @since 0.3.0
+	* @since 0.3.0
 	 *
 	 * @return void
 	 */
@@ -64,7 +64,7 @@ final class WP_Core_Abilities {
 	}
 
 	/**
-	 * Registers the `wp/get-site-info` ability.
+	 * Registers the `core/get-site-info` ability.
 	 *
 	 * @since 0.3.0
 	 *
@@ -83,7 +83,7 @@ final class WP_Core_Abilities {
 		);
 
 		wp_register_ability(
-			'wp/get-site-info',
+			'core/get-site-info',
 			array(
 				'label'               => __( 'Get Site Information' ),
 				'description'         => __( 'Returns a single site information field configured in WordPress (e.g., site name, URL, version) for display or diagnostics.' ),
@@ -140,7 +140,7 @@ final class WP_Core_Abilities {
 	}
 
 	/**
-	 * Registers the `wp/get-current-user-info` ability.
+	 * Registers the `core/get-current-user-info` ability.
 	 *
 	 * @since 0.3.0
 	 *
@@ -148,7 +148,7 @@ final class WP_Core_Abilities {
 	 */
 	protected static function register_get_current_user_info(): void {
 		wp_register_ability(
-			'wp/get-current-user-info',
+			'core/get-current-user-info',
 			array(
 				'label'               => __( 'Get Current User Information' ),
 				'description'         => __( 'Returns basic profile details for the current authenticated user to support personalization, auditing, and access-aware behavior.' ),
@@ -215,7 +215,7 @@ final class WP_Core_Abilities {
 	}
 
 	/**
-	 * Registers the `wp/get-environment-info` ability.
+	 * Registers the `core/get-environment-info` ability.
 	 *
 	 * @since 0.3.0
 	 *
@@ -223,7 +223,7 @@ final class WP_Core_Abilities {
 	 */
 	protected static function register_get_environment_info(): void {
 		wp_register_ability(
-			'wp/get-environment-info',
+			'core/get-environment-info',
 			array(
 				'label'               => __( 'Get Environment Info' ),
 				'description'         => __( 'Returns core details about the site\'s runtime context for diagnostics and compatibility (environment, PHP runtime, database server info, WordPress version).' ),
--- a/includes/abilities/class-wp-core-abilities.php
+++ b/includes/abilities/class-wp-core-abilities.php
@@ -4,7 +4,7 @@
  *
  * @package WordPress
  * @subpackage Abilities_API
- * @since n.e.x.t
+ * @since 0.3.0
  */
 
 declare( strict_types = 1 );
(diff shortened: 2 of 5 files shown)


includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,345 @@
...
+										'description' => __( 'Additional metadata associated with the ability.', 'abilities-api' ),
+									),
+									'annotations' => array(
+										'type'        => 'object',
+										'description' => __( 'Annotations describing ability behavior.', 'abilities-api' ),
+									),
+									'show_in_rest' => array(
+										'type'        => 'boolean',
+										'description' => __( 'Whether the ability is exposed in REST.', 'abilities-api' ),
+									),
```
JasonTheAdams: We should keep these inside of `meta` to be consistent with its structure.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,345 @@
...
+							continue;
+						}
+
+						$filtered_list[] = array(
+							'name'         => $ability->get_name(),
+							'label'        => $ability->get_label(),
+							'description'  => $ability->get_description(),
+							'meta'         => $ability->get_meta(),
+							'annotations'  => $ability->get_meta_item( 'annotations', array() ),
+							'show_in_rest' => $show_in_rest,
```
JasonTheAdams: I think it would be a good idea to make a `WP_Ability::to_array()` for a more declarative option.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,345 @@
...
+					),
+					'additionalProperties' => false,
+				),
+				'output_schema'       => array(
+					'type'       => 'object',
+					'properties' => array(
+						'abilities' => array(
+							'type'  => 'array',
+							'items' => array(
+								'type'       => 'object',
```
JasonTheAdams: I'm seriously considering a `WP_Ability::to_json_schema()` method to generate this structure. I could see this being used in a few places, including JS, and having a declarative way of getting the schema along with `::to_array()` would be useful.

JasonTheAdams (changes requested):
Nice work, @Jameswlepage! I left a few thoughts and suggestions. I didn't exhaustively review it, yet, but noticed these.

Jameswlepage:
- Updated the filter hook name from `abilities_api_core_find_abilities_results` to `wp_find_abilities_results` to follow WordPress core naming conventions (simpler and more consistent with other core hooks)
- Removed all text domains from translation functions since this is core code (WordPress core uses empty/default text domain)
- Fixed execute callback signatures to use optional parameters to match the PHPStan type definitions
- Cleaned up bootstrap.php to inline the test environment check, removing an unnecessary variable

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,345 @@
...
+					'type'                 => array( 'object', 'null' ),
+					'properties'           => array(
+						'namespace' => array(
+							'type'        => 'string',
+							'description' => __( 'Optional namespace prefix to filter abilities (e.g. "core/").', 'abilities-api' ),
+						),
+						'show_in_rest' => array(
+							'type'        => 'boolean',
+							'description' => __( 'Whether to limit results to abilities exposed in REST. Defaults to true.', 'abilities-api' ),
+						),
```
JasonTheAdams: So... there's a bit of a chick and egg issue here. Hahah!

1. This ability is marked as shown in REST
2. Therefore, it can be called via REST
3. The other abilities also control whether they show in REST
4. This limits the abilities that show up in the REST schema
5. But this ability can have this parameter set to false... which shows all abilities, regardless of their setting

The only way I can think of to circumvent this is to check in the execution callback of this ability if it's being called in the REST context. If it is, then it should _always_ honor if an ability is marked as shown in REST. But then that makes this parameter useless in a REST context, and there's not a simple way of removing this input in that context. We'd have to do one of:

1. Add a contextual parameter to inputs that shows/hides them in certain contexts
2. Make this parameter support a callback so it can be conditional to the context

No problem, right? 😆

gziolo: I would rather see it as a meta issue, since we have a REST API layer that lists all abilities and excludes those with `show_in_rest` set to `false`. Now, if we have the ability, that list abilities we need to take account a few things:
- Should this ability list also include itself?
- Does it make sense to have this ability on the server while everyone can call `wp_get_abilities`? The plan is to add filtering to this method anyway, so again, there is going to be duplication.
- When calling on the server, we can ignore `show_in_rest`, but when fetching this ability through REST API, should the list take into account `show_in_rest` for the listed items?

I'm having hard time jusifying why we would want to promote in WordPress core such a complex ability that replicates most of the existing logic on all three layers:
- server: `wp_get_abilities()` once it supports filtering
- REST API `wp/v2/abilities`  support for categories is coming, it works with pagination already
- client: `wp.abilities.getAbilities()` - support for filtering would help like categories that is coming soon, but is able to fetch paginated results when there are more than 50 items

includes/abilities/class-wp-core-abilities.php
gziolo: I think it's a consequence of what I explain in https://github.com/WordPress/abilities-api/pull/108#discussion_r2425586846. This replicates the existing REST API endpoint. This has far-reaching consequences as noted in https://github.com/WordPress/abilities-api/pull/109#pullrequestreview-3330388191, too.

gziolo:
Thank you for stating it. I personally would be pretty happy to include these two abilities with some minor tweaks:

> `core/get-bloginfo` - Retrieve individual site information fields (name, description, url, wpurl, admin_email, charset, language, version)

The method is pretty old, but its description is more up to date at https://developer.wordpress.org/reference/functions/get_bloginfo/:
> Retrieves information about the current site.

To align with that and the other ability name, I propose we go with: `core/get-current-site-info`. The implementation proposed returns a single requested field, based on the enum field. This doesn't provide enough information to LLMs. How do they differentiate between `url` and `wpurl` without descriptions? Not to mention that it might be better to return a list similar to what 'core/get-current-user-info' does.

`core/get-current-user-info` - Get current authenticated user data (id, display_name, user_nicename, user_login, roles, locale)

This is shaped nicely, as you could see from my references above.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,347 @@
...
+	 */
+	protected static function register_get_current_user_info(): void {
+		wp_register_ability(
+			'core/get-current-user-info',
+			array(
+				'label'               => __( 'Get Current User Information' ),
+				'description'         => __( 'Returns basic information about the current authenticated user.' ),
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'required'             => array( 'id', 'display_name', 'locale' ),
```
gziolo: This is a bit surprising, because there are more fields in the response, so it needs some adjustment to meet the actual implementation.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,347 @@
...
+						),
+						'locale'        => array(
+							'type'        => 'string',
+							'description' => __( 'The locale string for the user, such as en_US.' ),
+						),
+					),
+					'additionalProperties' => false,
+				),
+				'execute_callback'    => static function (): array {
+					$current_user = wp_get_current_user();
```
gziolo, line 220: This is going to be interesting to see how it plays out in practice with AI clients that use some access tokens. It might happen that this will report back as an agent user 😄

gziolo:
`core/get-environment-type` would be very low-level. It only returns a simple string for the env type like `production` vs `development`. How about making it more generic, and mirror the pattern from other abilities to make it something like `core/get-environment-info` and maybe provide some debug info folks might need to reason about the site that lives in `wp-config.php`. Although in that case, we would be cautious about the permission check.

Jameswlepage:
I think the call to remove the meta ability to find abilities is a good one, and will do this. 

I like that concept (find/execute), and I know there have been experiments by @galatanovidiu with the MCP adapter. Let's continue to experiment, and understand real world usage there and in the AI experiments plug-in

Jameswlepage:
@gziolo, interesting points on the get blog info feedback. I intentionally had it only return one to keep things very atomic/one-shot. But as I think about this further, I don't see any negative in allowing there to be bulk actions. 

I do wonder if we should name it something that deviates from get blog info. 

> core/get-environment-type

This too was intentionally as simple as it could get. I approached it considering how it could be a potential of the most basic ability, that also returned helpful information, and could then be layered into something like get site info. 

But then if you consider a real world use case, I think there's merit to having a full environment info ability.

includes/abilities/class-wp-core-abilities.php
JasonTheAdams: Resolving this as we've removed the ability.

Jameswlepage:
And the sea of failing tests are reminding me that the category system is now here! Noting that I am registering and categorizing these abilities as `core`. Any thoughts here?

JasonTheAdams:
> And the sea of failing tests are reminding me that the category system is now here! Noting that I am registering and categorizing these abilities as `core`. Any thoughts here?

I would propose:
* `wp/` or `wordpress/` namespace
* `site` category for the site info ability
* `user` category for the user info ability

The reason being that I can definitely see a future wherein there are more abilities tied to interacting with site data (e.g. changing the site title), as well as interacting with user data (e.g. setting their name). I think they're categorically distinct.

JasonTheAdams:
> This too was intentionally as simple as it could get. I approached it considering how it could be a potential of the most basic ability, that also returned helpful information, and could then be layered into something like get site info.

I think it could be useful to make this a touch more generic as `get-environment-info` with info such as:
* Environment type
* PHP version
* MySQL version
* WordPress version
* Database type (e.g. MySQL vs MariaDB)

Trying not to get into sensitive data. But I think it's useful for the AI to know details like that.

Jameswlepage:
An interesting consideration when thinking about naming **get blog info** versus **get site info,** get site info is not a function in WordPress - Because of that, I could see hallucinations being introduced when using very small LLMs to fill and run abilities. If we mirror the getBlogInfo ability to the schema of getBlogInfo, I would assume that the accuracy is higher, especially on lower parameter models (because that function is in the training data). 

First, this is only something that applies to TINY LLMS, which almost all users probably won't deal with. However, these _are_ the LLMs that would power on browser, on device, or on server AI. 

This had me interested, so I generated a basic experiment with ChatGPT and used the OSS Google Gemma 1b and 4b param models (because they power Chrome's built-in AI) to to a very quick, very nonscientific test.

https://github.com/Jameswlepage/abilities-llm-benchmarks/blob/main/FINAL_REPORT_wordpress_naming_analysis.md (this is a summary made by Claude Code from the outputs generated with LM Studio and the Gemma models. noting again that this is very quick and messy)

> TLDR: This benchmark tested whether naming WordPress abilities after their underlying PHP functions (like core/get-bloginfo) versus modern semantic names (like core/get-current-site-info) affects LLM accuracy and hallucination rates. Testing gemma-3-1b and gemma-3-4b models revealed that the 1B model failed with both approaches (25-38% accuracy), while the 4B model achieved perfect accuracy (8/8) with function-aligned naming but failed one critical test with modern naming—when asked to "call get_bloginfo()", it couldn't map the function name to the available modern ability name. The key insight is that function-aligned naming eliminates a semantic translation gap in domain-specific systems like WordPress where documentation, training data, and user prompts heavily reference the underlying function names, and we recommend minimum 4B parameter models for production ability-based systems.

In regards to actually renaming things, I agree with your point here, Greg. 

1. The Abilities API is an abstraction layer, not a 1:1 mirror of WordPress functions. If we tie naming too closely to legacy function names, we lose the opportunity to create a cleaner, more consistent API surface, so I think it's a good thing to showcase with the first abilities. WordPress has also moved from "blog" terminology to "site" terminology over the years. 
2. Schemas provide the context LLMs need - The description, label, and detailed input_schema/output_schema fields give LLMs sufficient context to understand what the ability does, regardless of whether it matches an internal function name. From my non-scientific tests, it looks like that works even for some of the smallest models out there.

The above consideration is probably just that - larger hosted models won't have trouble with this.

So - I've updated the original implementation of get blog info to a site specific implementation accordingly, but these are interesting considerations as we really determine how abilities, AI, and WordPress Core will all interact in the future.

gziolo:
The site/blog info might be a good opportunity to test `meta.annotations.instructions` to provide additional hints to LLM like: “This plays similar role to calling get_blog_info() in WordPress codebase.” So it can connect the dots.

justlevine:
wholly unrelated to the rest of this conversation but wanted to give this a special callout:

> so I generated a basic experiment with ChatGPT and used the OSS Google Gemma 1b and 4b param models (because they power Chrome's built-in AI) to to a very quick, very nonscientific test.

@Jameswlepage thank you for testing and sharing 🤩!
Even this sort of "nonscientific" rubber ducking is a big step forward from pure conjecture - surfacing in hopes we'll be more easily to find and point to a more practical example than [claude-cookbook](https://github.com/anthropics/claude-cookbooks/blob/main/tool_evaluation/tool_evaluation.ipynb) (e.g. for #30 )

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,294 @@
...
+	 * Registers the core abilities categories.
+	 *
+	 * @since 0.3.0
+	 *
+	 * @return void
+	 */
+	public static function register_category(): void {
+		// Site-related capabilities
+		wp_register_ability_category(
+			'site',
```
JasonTheAdams: Let's use a class constant here, same with `user`, for reuse in the abilities.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,294 @@
...
+						'value' => (string) $value,
+					);
+				},
+				'permission_callback' => static function (): bool {
+					// Site information can expose sensitive details; require admin capability.
+					return current_user_can( 'manage_options' );
+				},
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Retrieves a single site property by passing an allowed field to get_bloginfo().' ),
```
JasonTheAdams: I think I missed this annotation. What's this for? How is it different than the ability description?

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,294 @@
...
+					return is_user_logged_in();
+				},
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Retrieves information about the current authenticated user.' ),
+						'readonly'     => true,
+						'destructive'  => false,
+						'idempotent'   => true,
+					),
+					'show_in_rest' => true,
```
JasonTheAdams: I think we want to avoid duplicate purpose REST endpoints. There's already the `/wp/v2/users/me` endpoint that retrieves the current user's information. So I suggest we make this false.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,294 @@
...
+						'roles'         => $current_user->roles,
+						'locale'        => get_user_locale( $current_user ),
+					);
+				},
+				'permission_callback' => static function (): bool {
+					return is_user_logged_in();
+				},
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Retrieves information about the current authenticated user.' ),
```
JasonTheAdams: Same question here: How is this different from the ability description?

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,294 @@
...
+					if ( stripos( $db_version, 'mariadb' ) !== false ) {
+						$type = 'mariadb';
+					}
+
+					return array(
+						'environment'   => (string) $env,
+						'php_version'   => (string) $php_version,
+						'mysql_version' => (string) $db_version,
+						'wp_version'    => (string) $wp_version,
+						'database_type' => (string) $type,
```
JasonTheAdams: Why are we explicitly casting all of these? Aren't most or all of them strings already?

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,294 @@
...
+					);
+				},
+				'permission_callback' => static function (): bool {
+					// Environment information is restricted to administrators.
+					return current_user_can( 'manage_options' );
+				},
+				'meta'                => array(
+					'annotations'  => array(
+						'instructions' => __( 'Retrieves environment information such as environment type, PHP, database, and WordPress versions.' ),
```
JasonTheAdams: Last one. Still curious. 😂

JasonTheAdams (changes requested):
Good work, @Jameswlepage! Left some suggestions/questions. 👍

includes/abilities/class-wp-core-abilities.php
Jameswlepage: I think we drop instructions in this; was exploring how AI could interact with something like (but this one isn't any diff from the description).

includes/abilities/class-wp-core-abilities.php
Jameswlepage: Good note everything is a string other than the db version, and that "DB version" got me thinking as well. There are some artifacts from previous thoughts here. It shouldn't be the MySQL version because it could be Maria. And that is clarified in the database type. So going to move database type above the database version, and use db_server_info instead of this implementation.

includes/abilities/class-wp-core-abilities.php
Jameswlepage: This was in response to one of Greg's notes about using annotations to steer models should they get confused by the naming convention (ie; hey AI, the underlying function powering this is actually get_bloginfo()). This description could probably be more helpful. And I actually don't think we should add instructions in the (first) core implementation anymore - folks could extend if they want

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,286 @@
...
+				'category'            => self::CATEGORY_SITE,
+				'input_schema'        => array(
+					'type'                 => 'object',
+					'properties'           => array(
+						'field' => array(
+							'type'        => 'string',
+							'enum'        => $fields,
+							'description' => __( 'The site information field to retrieve.' ),
+						),
+					),
```
JasonTheAdams: I'm going to be very curious to see how well the AI uses this. If each field was a distinct property with its own description, that would probably work the best, but I get why we're doing this. If it doesn't work well, then we could have each field as a boolean, where the AI determines which fields it wants and which it doesn't, returning multiple at once.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,286 @@
...
+							'type'        => 'string',
+							'description' => __( 'The site\'s runtime environment classification (e.g., production, staging, development).' ),
+						),
+						'php_version'    => array(
+							'type'        => 'string',
+							'description' => __( 'The PHP runtime version executing WordPress.' ),
+						),
+						'db_server_info' => array(
+							'type'        => 'string',
+							'description' => __( 'The database server vendor and version string reported by the driver (e.g., “8.0.34”, “10.11.6-MariaDB”).' ),
```
JasonTheAdams: This is a good opportunity to use the "examples" JSON Schema keyword to hint what to expect
```suggestion
							'description' => __( 'The database server vendor and version string reported by the driver.' ),
							'examples' => ['8.0.34', '10.11.6-MariaDB']
```

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,286 @@
...
+				'label'               => __( 'Get Environment Info' ),
+				'description'         => __( 'Returns core details about the site\'s runtime context for diagnostics and compatibility (environment, PHP runtime, database server info, WordPress version).' ),
+				'category'            => self::CATEGORY_SITE,
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'required'             => array( 'environment', 'php_version', 'db_server_info', 'wp_version' ),
+					'properties'           => array(
+						'environment'    => array(
+							'type'        => 'string',
+							'description' => __( 'The site\'s runtime environment classification (e.g., production, staging, development).' ),
```
JasonTheAdams, line 266: We can use examples here, too.

JasonTheAdams (approved):
Nicely done, @Jameswlepage! I have a few more comments, but they're not blockers to merging for me, so I'll go ahead and approve. Your choice if you'd like to implement any of the suggestions.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,288 @@
...
+
+declare( strict_types = 1 );
+
+/**
+ * Registers the default core abilities that ship with the Abilities API.
+ *
+ * @since 0.3.0
+ */
+// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedClassFound -- Core class intended for WordPress core.
+class WP_Core_Abilities {
```
justlevine: Keeps iteration future-compat

```suggestion
final class WP_Core_Abilities {
```

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,288 @@
...
+	/**
+	 * Category slugs for core abilities.
+	 *
+	 * @since 0.3.0
+	 */
+	public const CATEGORY_SITE = 'site';
+	public const CATEGORY_USER = 'user';
```
justlevine, line 25: Are these meant to be consumed outside this class? If not, I'd keep them private for now.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,288 @@
...
+			'url',
+			'wpurl',
+			'admin_email',
+			'charset',
+			'language',
+			'version',
+		);
+
+		wp_register_ability(
+			'wp/get-site-info',
```
justlevine: should our prefix be `core` to align block namespaces? (personally think `wp` is fine)

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,288 @@
...
+	/**
+	 * Registers the `wp/get-current-user-info` ability.
+	 *
+	 * @since 0.3.0
+	 *
+	 * @return void
+	 */
+	protected static function register_get_current_user_info(): void {
+		wp_register_ability(
+			'wp/get-current-user-info',
```
justlevine: I'm wondering if it makes sense to generalize this to `wp/get-user-info` and then add an optional `id: int` input field (which in the future could be expanded with an `idType` and polymorphism or separate `login_name`, `slug`, `email` fields) that defaults to the current user if not supplied.

I don't think we need _both_, and implementing it this way will make it easier to handle the "agent user" pattern that @gziolo brought up in https://github.com/WordPress/abilities-api/pull/108#discussion_r2425630892

includes/abilities/class-wp-core-abilities.php
justlevine: Related: https://github.com/WordPress/abilities-api/pull/108#discussion_r2437776398

Switching to `current_user_can()` per the above would make this useful for both the agent user, and the user interacting with the agent.

includes/abilities/class-wp-core-abilities.php
justlevine: (In 7.0 I want to to try and refactor that endpoint to consume this ability 🤞)

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,288 @@
...
+			'charset',
+			'language',
+			'version',
+		);
+
+		wp_register_ability(
+			'wp/get-site-info',
+			array(
+				'label'               => __( 'Get Site Information' ),
+				'description'         => __( 'Returns a single site information field configured in WordPress (e.g., site name, URL, version) for display or diagnostics.' ),
```
justlevine: Similarly to https://github.com/WordPress/abilities-api/pull/108#discussion_r2437776398, I'm wondering if we'd get more mileage out of this if it returned an entire array of blog_info / site health, and instead used a `fields: string[]` if a user only wants a subset.

The obvious use case is LLMs/REST where you don't want a bunch of round trips for basic data, but I think it's helpful beyond that. 

Semantically, it lets '-info' align better with the other two abilities.

Also, a shaped output_schema with the actual properties and matching types is better for both LLMs, traditional tools, and humans to reason with vs an `array{field:string,value:mixed}`. (We see the effects of this a lot in GraphQL/headless land)


Thoughts?

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,288 @@
...
+					'label'       => __( 'User' ),
+					'description' => __( 'Abilities that retrieve or modify user information and settings.' ),
+				)
+			);
+	}
+
+	/**
+	 * Registers the default core abilities.
+	 *
+ * @since 0.3.0
```
justlevine:
```suggestion
 	* @since 0.3.0
```

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,288 @@
...
+			);
+
+		// User-related capabilities
+			wp_register_ability_category(
+				self::CATEGORY_USER,
+				array(
+					'label'       => __( 'User' ),
+					'description' => __( 'Abilities that retrieve or modify user information and settings.' ),
+				)
+			);
```
justlevine:
```suggestion
		// Site-related capabilities
		wp_register_ability_category(
			self::CATEGORY_SITE,
			array(
				'label'       => __( 'Site' ),
				'description' => __( 'Abilities that retrieve or modify site information and settings.' ),
			)
		);

		// User-related capabilities
		wp_register_ability_category(
			self::CATEGORY_USER,
			array(
				'label'       => __( 'User' ),
				'description' => __( 'Abilities that retrieve or modify user information and settings.' ),
			)
		);
```

justlevine (commented):
Code LGTM 🙌

Left a couple thoughts re robustness/future-compat below. Nonblocking, but I do think making the abilities more generic for 6.9 will be very helpful for both ourselves in the other building blocks and downstream adopters (and aligns with the feedback on https://github.com/WordPress/ai/issues/40 )

includes/abilities/class-wp-core-abilities.php
justlevine: Related: https://github.com/WordPress/wordpress-develop/pull/9410#discussion_r2437614400

includes/abilities/class-wp-core-abilities.php
gziolo: I think it’s perfectly fine to start with supporting only the current user because it has simple permission check model. There isn’t anything sensitive about asking for your info. If any, we could drop the `current` part from the ability unique name in case we feel it would help making it more extensible to support also fetching details about other users later. Filtering output by field name(s) seems like enhancement worth exploring but I would keep it for later because right now there isn’t much to filter.

includes/abilities/class-wp-core-abilities.php
gziolo: I’m biased and prefer `core/` based on my familiarity with the Gutenberg project where every aspect uses that convention: block types, block patterns, block bindings, data stores, etc.

includes/abilities/class-wp-core-abilities.php
gziolo: However I guess you are comparing it to the site info where such filtering by field is possible. I left my comment earlier, the downside of the shape proposed for site info is that the input and output schema describes more the shape of the response rather what individual fields represent. It’s much richer for the user as every field has description explaining their role.

includes/abilities/class-wp-core-abilities.php
JasonTheAdams: It was `core/` and I had suggested `wp/`. I didn't think of the convention already being established elsewhere. It's confusing to me that WP sometimes prefixes with `wp_` and other times uses `core`. 🤷‍♂️

includes/abilities/class-wp-core-abilities.php
JasonTheAdams: Good scope thought! It is possible for others to use these categories in their abilities, so I think it's worth keeping public.

Jameswlepage: I agree, I think they should be public because ideally the category system, especially those bundled with core, is extended to include additional core functionality that we simply don't include in 6. 9 here.

includes/abilities/class-wp-core-abilities.php
Jameswlepage: If Gutenberg uses core as a prefix, we probably should follow that convention because it's the more modern implementation on WordPress. If anybody has strong feelings, let me know. 

Will roll it back to core, but not super opinionated here.

includes/abilities/class-wp-core-abilities.php
Jameswlepage: Renamed it - good optionality now, but I like starting as direct/upfront/atomic as possible.

includes/abilities/class-wp-core-abilities.php
Jameswlepage: Good point. It was the first ability I wrote, so missing consistency _and_ not self documenting which I think is a big unlock. Kind of 'layered' but I think that's OK in this situation.

Jameswlepage: However, if we make this change, then REST would require a weird empty `?input` field given how the current controller works.

GET /run?input  # All fields
GET /run?input[fields][]=name  # Filtered

This could be a nonbreaking improvement in the future and imo shouldn't block

gziolo: I think the problem is that when you provide empty `input`, then the REST API sanitizes it and sets it to an empty array. Otherwise, `execute` will use its default value `null`. So, the solution could be allowing `null` in the input schema to address it. There might be better alternatives to consider, but I can't think of a simpler approach.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,315 @@
...
+		);
+
+		wp_register_ability(
+			'core/get-site-info',
+			array(
+				'label'               => __( 'Get Site Information' ),
+				'description'         => __( 'Returns site information configured in WordPress. By default returns all fields, or optionally a filtered subset.' ),
+				'category'            => self::CATEGORY_SITE,
+				'input_schema'        => array(
+					'type'                 => 'object',
```
gziolo, line 92: This might address the issue raised about the quirks with calling REST API with no input:
```suggestion
					'type'                 => array( 'object', 'null, ),
```

justlevine: @gziolo `wp-env` is down so I cant test easily, but I think `null` means an _explicit_ `null` value needs to be passed.
If so, I believe another way to work around this would be to coerce `null|unset` to an empty array (IIRC this was the approach I was using when I was advocating to keep `input` an `array<string.mixed>|null` for 6.9)

includes/abilities/class-wp-core-abilities.php
```diff
@@ -151,7 +153,8 @@ protected static function register_get_site_info(): void {
 
 					return $result;
 				},
-				'permission_callback' => static function (): bool {
+				'permission_callback' => static function ( $input = null ): bool {
+					unset( $input );
```
Jameswlepage: this is due to phpcs, feels hacky, maybe to ignore?

JasonTheAdams: Huh? What's PHPCS complaining about if you don't unset it?

Jameswlepage: It's an unused var warning

JasonTheAdams: Why not just remove the parameter from the callback, then? Ignoring parameters by not listing them doesn't hurt anything. 🙂

Jameswlepage: 🤦

includes/abilities-api/class-wp-ability.php
```diff
@@ -320,6 +320,27 @@ public function get_input_schema(): array {
...
 
+	/**
+	 * Applies the defined input default when no input is provided.
+	 *
+	 * @since 0.4.0
+	 *
+	 * @param mixed $input Optional. The raw input provided for the ability. Default `null`.
+	 * @return mixed The input with the schema default applied when available.
+	 */
+	public function normalize_input( $input = null ) {
```
gziolo, line 331: I ran out of time to include that in the patch against WordPress `trunk`. Is there anyone willing to handle the bug fix PR in core just for this change? 😄

includes/abilities/class-wp-core-abilities.php
gziolo: `normalize_input` helper fixes the issue 👍🏻

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,317 @@
...
+
+declare( strict_types = 1 );
+
+/**
+ * Registers the default core abilities that ship with the Abilities API.
+ *
+ * @since 0.3.0
+ */
+// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedClassFound -- Core class intended for WordPress core.
+final class WP_Core_Abilities {
```
aaronjorbin, line 18: What is the value in putting these in a class? It feels like this could all be done in more straightforward functions.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,317 @@
...
+			'description',
+			'url',
+			'wpurl',
+			'admin_email',
+			'charset',
+			'language',
+			'version',
+		);
+
+		wp_register_ability(
```
aaronjorbin, line 85: This seems to be duplicating the [site settings REST endpoint](https://developer.wordpress.org/rest-api/reference/settings/).  What is the value of exposing this here when there is already a way to expose this information.

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,317 @@
...
+						),
+						'user_login'    => array(
+							'type'        => 'string',
+							'description' => __( 'The login username for the user.' ),
+						),
+						'roles'         => array(
+							'type'        => 'array',
+							'description' => __( 'The roles assigned to the user.' ),
+							'items'       => array(
+								'type' => 'string',
```
aaronjorbin, line 209: Roles is an array, not a string.  See:

https://core.trac.wordpress.org/browser/trunk/src/wp-includes/class-wp-user.php#L82

includes/abilities/class-wp-core-abilities.php
```diff
@@ -0,0 +1,317 @@
...
+				'label'               => __( 'Get Environment Info' ),
+				'description'         => __( 'Returns core details about the site\'s runtime context for diagnostics and compatibility (environment, PHP runtime, database server info, WordPress version).' ),
+				'category'            => self::CATEGORY_SITE,
+				'output_schema'       => array(
+					'type'                 => 'object',
+					'required'             => array( 'environment', 'php_version', 'db_server_info', 'wp_version' ),
+					'properties'           => array(
+						'environment'    => array(
+							'type'        => 'string',
+							'description' => __( 'The site\'s runtime environment classification (e.g., production, staging, development).' ),
```
aaronjorbin, line 266: This can only be one of four values.  I think it's valueable to make that clearer.

aaronjorbin (commented):
I know this was merged, but here are a handful of things to address before a core patch

includes/abilities/class-wp-core-abilities.php
justlevine: Great Q!

If you're asking from a more philosophical POV ("why do Abilities even exist"), then the answer is similar to "What is the value of having REST endpoints when there's already a php function and ajax?". The tl;dr goal of the Abilities API is to create a sort of "Generic API" abstraction + registry, the idea being that instead of needed to reregister/expose the same functionality to various last-mile APIs (REST, admin-ajax, graphql,  PHP in theme/plugins, MCP/A2A, Command Palette, and whatever future tech integrations that come up with). It's standardized and reliable, like hooks but for functionality, so all anybody needs is a generic adapter and it'l work without any domain knowledge of the ability itself.

In fact, I'd hope and expect to see the internals of many of our existing REST endpoints updated to internally use the Abilities API in 7.0 and beyond.

If you're asking from a more practical POV (e.g. regarding the shape of the input/output schemas), I think we're open to evolution here. I believe we want something that it generic to make sense in multiple context, and not just a single one of either REST, MCP, or Command Palette.

includes/abilities/class-wp-core-abilities.php
gziolo: It should be perfectly fine to introduce two functions or static methods and attach them inside `src/wp-includes/default-filters.php` in WP core, for example:
```php
add_filter( 'wp_abilities_api_categories_init', 'register_core_ability_categories' );
add_filter( 'wp_abilities_api_init', 'register_core_abilities' );
```

includes/abilities-api/class-wp-ability.php
gziolo: https://github.com/WordPress/wordpress-develop/pull/10395 - started working on the patch. Unit tests are still missing at the time of writing.

includes/abilities/class-wp-core-abilities.php
jorgefilipecosta: Hi @aaronjorbin, what we have seems correct, roles is of type array with each item being a string. Feel free to correct me if I'm missing anything.

includes/abilities/class-wp-core-abilities.php
gziolo: `enum` would fit perfectly here 👍🏻

PR #114, opened 2025-10-13
Add category support to client package

emdashcodes:
This PR adds some basic category registration and filtering functionality to the `@wordpress/abilities` client package, allowing developers to filter abilities by category through the JavaScript API. This addresses some of the feedback highlighted in https://github.com/WordPress/abilities-api/pull/102#pullrequestreview-3316543742.

## What's Changed

- Added required `category: string` field to the `Ability` interface and store
- Updated `getAbilities()` API function to accept an optional `category` parameter through `args` object
- Implemented client-side category filtering in store selectors

### Tests

- Updated all existing tests to include category field
- Added tests for category filtering in selectors
- Added tests for empty/non-existent category handling

### Documentation updates

- Added missing `registerAbility` documentation
- Added category parameter documentation
- Added example for category filtering

## Not Addressed in this PR

- One limitation of this PR is that we do not load categories from the server for validation, because we would need to add a REST API endpoint for fetching categories. Is this something we would be interested in? If we introduce categories over REST and a category does not opt into the REST API, it wouldn't get exposed to the client anyway, so that is something to consider. I can work on this in a follow-up, but want to make sure that's the direction we want to take first. 
- This PR also does not introduce any client-specific methods for registering new categories. Because of the larger client/server separation, it's not clear to me how these two registries are supposed to work together anymore, as we basically end up with two separate registries. In my original mental model of how things worked, these registries were one and the same.

## Testing Instructions

- Run `npm run test:client`
- Make sure all tests pass
- Make sure that you have a few test abilities registered, server-side and/or client-side, which you can see examples of in the previous client PRs: https://github.com/WordPress/abilities-api/pull/60). To correctly expose server-side abilities, make sure to set `meta.show_in_rest` to `true`
```php
		'meta'        => array(
			'show_in_rest' => true,
		),
```
- Update any of those to pass in a category.
- Now test running and filtering by various different parameters depending on the abilities you have registered.
For example, I have some server and client abilities, and I can filter by navigation: 
```js
  const nav = await wp.abilities.getAbilities({category: 'navigation'});
  console.log('Navigation:', nav.map(a => a.name));
```
- Output all your registered categories and abilities
```js
const summary = {};
const abilities = await wp.abilities.getAbilities();

for (const ability of abilities) {
  const category = ability.category;

  if (!summary[category]) {
      summary[category] = {
          Category: category,
          Abilities: []
      };
  }

  summary[category].Abilities.push(ability.name);
}

for (const [category, data] of Object.entries(summary)) {
  console.log(`\n${category}:`);
  console.log(data.Abilities.join(', '));
}
```

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -75,13 +75,18 @@ function MyComponent() {
 
 ### Functions
 
-#### `getAbilities(): Promise<Ability[]>`
+#### `getAbilities(category?: string): Promise<Ability[]>`
 
-Returns all registered abilities. Automatically handles pagination to fetch all abilities across multiple pages if needed.
+Returns all registered abilities. Optionally filter by category slug. Automatically handles pagination to fetch all abilities across multiple pages if needed.
 
 ```javascript
+// Get all abilities
 const abilities = await getAbilities();
 console.log( `Found ${ abilities.length } abilities` );
+
+// Get abilities in a specific category
+const dataAbilities = await getAbilities( 'data-retrieval' );
+console.log( `Found ${ dataAbilities.length } data retrieval abilities` );
 ```
 
 #### `getAbility(name: string): Promise<Ability | null>`
@@ -93,6 +98,32 @@ const ability = await getAbility( 'my-plugin/create-post' );
 if ( ability ) {
   console.log( `Found ability: ${ ability.label }` );
 }
+```
+
+#### `registerAbility(ability: Ability): void`
+
+Registers a client-side ability. Client abilities are executed locally in the browser and must include a callback function and a valid category.
+
+```javascript
+import { registerAbility } from '@wordpress/abilities';
+
+registerAbility( {
+  name: 'my-plugin/navigate',
+  label: 'Navigate to URL',
+  description: 'Navigates to a URL within WordPress admin',
+  category: 'navigation',
+  input_schema: {
+    type: 'object',
+    properties: {
+      url: { type: 'string' }
+    },
+    required: [ 'url' ]
+  },
+  callback: async ( { url } ) => {
+    window.location.href = url;
+    return { success: true };
+  }
+} );
 ```
 
 #### `executeAbility(name: string, input?: Record<string, any>): Promise<any>`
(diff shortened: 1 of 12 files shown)


packages/client/README.md
```diff
@@ -75,13 +75,18 @@ function MyComponent() {
...
-Returns all registered abilities. Automatically handles pagination to fetch all abilities across multiple pages if needed.
+Returns all registered abilities. Optionally filter by category slug. Automatically handles pagination to fetch all abilities across multiple pages if needed.
 
 ```javascript
+// Get all abilities
 console.log( `Found ${ abilities.length } abilities` );
+
+// Get abilities in a specific category
+const dataAbilities = await getAbilities( 'data-retrieval' );
```
gziolo: We still don’t have this functionality present on the server, but I would use an object to pass the category name:

```suggestion
const dataAbilities = await getAbilities( { category: 'data-retrieval' } );
```

This way we can gradually introduce more filters.

gziolo: See: https://github.com/WordPress/abilities-api/pull/115.

API proposed:

```php
// Get abilities in a specific category
$abilities = wp_get_abilities( array( 'category' => 'data-retrieval' ) );
```

gziolo: Addressed my feedback in [951d33c](https://github.com/WordPress/abilities-api/pull/114/commits/951d33c26848516b77820539a64d03fd8e78abe5).

gziolo (approved):
@emdashcodes, everything looks great from my perspective. This bring the client and server implementation closer to each other after the mandatory category got introduced. As noted in the feedback, @galatanovidiu is working in #115 on adding filtering capabilities on the server, so I added a commit to account on the same shape on the client. This way we should have the interfaces open for further extension with not risk of backward incompatibilities.

I'm approving this PR, but since I added new commits that change some code paths substantially, I would appreciate some sanity check.

emdashcodes (commented):
Your two changes look good to me. I did some testing on it this morning as well. Thanks for handling it!

PR #115, opened 2025-10-14
Add Filtering Capabilities for Registered Abilities

### LINKED ISSUE #38: Proposal: Add a convienient way to filter the list of all registered abilities

galatanovidiu:
## Summary

Closes https://github.com/WordPress/abilities-api/issues/38.

Adds querying and filtering capabilities to the Abilities API, allowing developers to efficiently find and retrieve specific abilities from potentially thousands of registered abilities.

This PR introduces the WP_Abilities_Query class and extends wp_get_abilities() to accept filtering arguments, enabling server-side filtering by category, namespace, meta properties, search terms, and more.

 ## Changes

**New Features**

  - New `WP_Abilities_Query` class - WordPress-standard query interface for filtering abilities
  - Enhanced `wp_get_abilities()` function - Now accepts optional `$args` parameter for filtering
  - Multiple filter types:
    - category - Filter by category slug (single or array)
    - namespace - Filter by namespace (single or array)
    - search - Search in name, label, and description
    - meta - Filter by meta properties (show_in_rest, annotations, custom meta)
    - orderby - Sort by name, label, or category
    - order - Sort direction (ASC/DESC)
    - limit/offset - Pagination support

 **Implementation Details**

  - All filters use AND logic between different filter types
  - Multiple values in category or namespace use OR logic
  - Meta filters support nested arrays with AND logic
  - Single-pass filtering for optimal performance

## Documentation

  - New comprehensive documentation in `docs/8.querying-filtering-abilities.md`
  - Updated `docs/4.using-abilities.md` with filtering examples
  - Includes a quick reference table and detailed examples
  
## Backward Compatibility

Fully backward compatible - existing code calling `wp_get_abilities() `without arguments continues to work unchanged.

### DIFF
--- a/docs/8.querying-filtering-abilities.md
+++ b/docs/8.querying-filtering-abilities.md
@@ -1,0 +1,213 @@
+# 8. Querying and Filtering Abilities
+
+The Abilities API provides powerful filtering capabilities through the `WP_Abilities_Query` class and the `wp_get_abilities()` function. This allows you to efficiently find and retrieve specific abilities from potentially thousands of registered abilities.
+
+`wp_get_abilities()` returns an array of `WP_Ability` objects, keyed by ability name.
+
+When no abilities match the query, an empty array is returned.
+
+Invalid query parameters are normalized to safe defaults.
+
+## Overview
+
+When working with large numbers of abilities registered by core and various plugins, the query system provides optimized server-side filtering. You can pass query arguments to `wp_get_abilities()` to filter abilities by category, namespace, meta properties, and more.
+
+```php
+// Get all abilities
+$abilities = wp_get_abilities();
+
+// Filter by category
+$abilities = wp_get_abilities( array( 'category' => 'data-retrieval' ) );
+
+// Combine multiple filters
+$abilities = wp_get_abilities( array(
+    'namespace' => 'my-plugin',
+    'category'  => 'communication',
+    'orderby'   => 'label',
+) );
+```
+
+## Quick Reference: Query Parameters
+
+| Parameter | Type | Default | Description |
+|-----------|------|---------|-------------|
+| `category` | string\|array | `''` | Filter by category slug. Accepts single value or array for multiple categories (OR logic) |
+| `namespace` | string\|array | `''` | Filter by namespace. Accepts single value or array for multiple namespaces (OR logic) |
+| `search` | string | `''` | Search in name, label, and description (case-insensitive) |
+| `meta` | array | `array()` | Filter by meta properties using nested arrays (AND logic between multiple meta filters) |
+| `orderby` | string | `''` | Sort by: `'name'`, `'label'`, or `'category'`. Empty = no sorting |
+| `order` | string | `'ASC'` | Sort direction: `'ASC'` or `'DESC'` |
+| `limit` | int | `-1` | Maximum results to return. `-1` = unlimited |
+| `offset` | int | `0` | Number of results to skip |
+
+## Basic Filtering
+
+### Filter by Category
+
+```php
+// Get all data retrieval abilities
+$abilities = wp_get_abilities( array( 'category' => 'data-retrieval' ) );
+
+// Get all communication abilities
+$abilities = wp_get_abilities( array( 'category' => 'communication' ) );
+
+// Get abilities from multiple categories (OR logic)
+$abilities = wp_get_abilities( array( 
+    'category' => array( 'math', 'communication', 'data-retrieval' ),
+) );
+```
+
+### Filter by Namespace
+
+Namespace is the part before the `/` in the ability name (e.g., `my-plugin` in `my-plugin/send-email`).
+
+```php
+// Get all abilities from your plugin
+$abilities = wp_get_abilities( array( 'namespace' => 'my-plugin' ) );
+
+// Get all WooCommerce abilities
+$abilities = wp_get_abilities( array( 'namespace' => 'woocommerce' ) );
+
+// Get abilities from multiple namespaces (OR logic)
+$abilities = wp_get_abilities( array(
+    'namespace' => array( 'my-plugin', 'woocommerce', 'jetpack' ),
+) );
+```
+
+### Search Abilities
+
+Search is case-insensitive and searches in name, label, and description:
+
+```php
+// Find all abilities related to email
+$abilities = wp_get_abilities( array( 'search' => 'email' ) );
+
+// Find payment-related abilities
+$abilities = wp_get_abilities( array( 'search' => 'payment' ) );
+```
+
+### Combine Multiple Filters
+
+Filters are cumulative (AND logic):
+
+```php
+// Get communication abilities from my-plugin only
+$abilities = wp_get_abilities( array(
+    'category'  => 'communication',
+    'namespace' => 'my-plugin',
+) );
+
+// Search for email abilities in the communication category
+$abilities = wp_get_abilities( array(
+    'category' => 'communication',
+    'search'   => 'email',
+) );
+```
+
+## Meta Filtering
+
+Filter abilities by their meta properties, including `show_in_rest`, `annotations`, and custom meta keys. All meta filters use AND logic — abilities must match all specified criteria.
+
+### Filter by REST API Visibility
+
+```php
+// Get abilities exposed in REST API
+$abilities = wp_get_abilities( array(
+    'meta' => array(
+        'show_in_rest' => true,
+    ),
+) );
+
+// Get abilities hidden from REST API
+$abilities = wp_get_abilities( array(
+    'meta' => array(
+        'show_in_rest' => false,
+    ),
+) );
+```
+
+### Filter by Annotations
+
+```php
+// Get readonly abilities
+$abilities = wp_get_abilities( array(
+    'meta' => array(
+        'annotations' => array(
+            'readonly' => true,
+        ),
+    ),
+) );
+
+// Get non-destructive, idempotent abilities
+$abilities = wp_get_abilities( array(
+    'meta' => array(
+        'annotations' => array(
+            'destructive' => false,
+            'idempotent'  => true,
+        ),
+    ),
+) );
+```
+
+### Combine Multiple Meta Filters
+
+When you specify multiple meta conditions, abilities must match all of them (AND logic):
+
+```php
+// Get readonly abilities that are exposed in REST API
+$abilities = wp_get_abilities( array(
+    'meta' => array(
+        'show_in_rest' => true,
+        'annotations'  => array(
+            'readonly' => true,
+        ),
+    ),
+) );
+
+// Get REST-enabled, readonly, idempotent abilities
+$abilities = wp_get_abilities( array(
+    'meta' => array(
+        'show_in_rest' => true,
+        'annotations'  => array(
+            'readonly'   => true,
+            'idempotent' => true,
+        ),
+    ),
+) );
+```
+
+### Filter by Custom Meta
+
+You can filter by any custom meta keys you've added to abilities:
+
+```php
+$abilities = wp_get_abilities( array(
+    'meta' => array(
+        'custom_key'   => 'custom_value',
+        'another_meta' => 'another_value',
+    ),
+) );
+```
+
+## Ordering and Pagination
+
+```php
+// Order by name, label, or category
+$abilities = wp_get_abilities( array(
+    'orderby' => 'label',
+    'order'   => 'ASC', // or 'DESC'
+) );
+
+// Paginate results
+$abilities = wp_get_abilities( array(
+    'limit'  => 10,
+    'offset' => 0,
+) );
+```
+
+## See Also
+
+- [Registering Abilities](3.registering-abilities.md) - How to register abilities with proper metadata
+- [Using Abilities](4.using-abilities.md) - Basic ability usage, execution, and permissions
+- [Registering Categories](7.registering-categories.md) - How to register ability categories
+- [REST API](5.rest-api.md) - REST API endpoints for abilities
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -95,13 +95,38 @@ function wp_get_ability( string $name ): ?WP_Ability {
  * Retrieves all registered abilities using Abilities API.
  *
  * @since 0.1.0
+ * @since n.e.x.t Added optional $args parameter for filtering abilities.
  *
  * @see WP_Abilities_Registry::get_all_registered()
+ * @see WP_Abilities_Query
  *
+ * @param array<string,mixed> $args Optional. Arguments to filter abilities. Default empty array.
+ *                                   Accepts 'category', 'namespace', 'search', 'meta', 'orderby',
+ *                                   'order', 'limit', and 'offset'.
+ *                                   'category' and 'namespace' accept string or array for multi-value filtering.
+ *                                   All filters use AND logic between different filter types and meta properties.
  * @return \WP_Ability[] The array of registered abilities.
+ *
+ * @phpstan-param array{
+ *   category?: string|string[],
+ *   namespace?: string|string[],
+ *   search?: string,
+ *   meta?: array<string,mixed>,
+ *   orderby?: string,
+ *   order?: string,
+ *   limit?: int,
+ *   offset?: int,
+ *   ...<string, mixed>
+ * } $args
  */
-function wp_get_abilities(): array {
-	return WP_Abilities_Registry::get_instance()->get_all_registered();
+function wp_get_abilities( array $args = array() ): array {
+	if ( empty( $args ) ) {
+		// Backward compatibility: return all abilities if no args.
+		return WP_Abilities_Registry::get_instance()->get_all_registered();
+	}
+
+	$query = new WP_Abilities_Query( $args );
+	return $query->get_abilities();
 }
 
 /**
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -94,10 +94,11 @@ function wp_get_ability( string $name ): ?WP_Ability {
 /**
  * Retrieves all registered abilities using Abilities API.
  *
+ * Uses WP_Abilities_Query to retrieve abilities from the registry with optional filtering.
+ *
  * @since 0.1.0
  * @since n.e.x.t Added optional $args parameter for filtering abilities.
  *
- * @see WP_Abilities_Registry::get_all_registered()
  * @see WP_Abilities_Query
  *
  * @param array<string,mixed> $args Optional. Arguments to filter abilities. Default empty array.
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -100,6 +100,7 @@ function wp_get_ability( string $name ): ?WP_Ability {
  * @since n.e.x.t Added optional $args parameter for filtering abilities.
  *
  * @see WP_Abilities_Query
+ * @see WP_Abilities_Query::AbilityQueryArgs for the full type definition.
  *
  * @param array<string,mixed> $args Optional. Arguments to filter abilities. Default empty array.
  *                                   Accepts 'category', 'namespace', 'search', 'meta', 'orderby',
--- a/includes/abilities-api.php
+++ b/includes/abilities-api.php
@@ -108,8 +108,8 @@ function wp_get_ability( string $name ): ?WP_Ability {
  * @return \WP_Ability[] The array of registered abilities.
  *
  * @phpstan-param array{
- *   category?: string|string[],
- *   namespace?: string|string[],
+ *   category?: string|array<string>,
+ *   namespace?: string|array<string>,
  *   search?: string,
  *   meta?: array<string,mixed>,
  *   orderby?: string,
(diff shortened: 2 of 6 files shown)


includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,515 @@
...
+ */
+class WP_Abilities_Query {
+
+	/**
+	 * Constant representing no limit on results.
+	 *
+	 * @since n.e.x.t
+	 * @var int
+	 */
+	private static $no_limit = -1;
```
JasonTheAdams, line 47: Why is this a static property instead of a constant? Same for the static properties below.

includes/abilities-api.php
```diff
@@ -95,13 +95,38 @@ function wp_get_ability( string $name ): ?WP_Ability {
...
+ *   limit?: int,
+ *   offset?: int,
+ *   ...<string, mixed>
+ * } $args
  */
-function wp_get_abilities(): array {
-	return WP_Abilities_Registry::get_instance()->get_all_registered();
+function wp_get_abilities( array $args = array() ): array {
+	if ( empty( $args ) ) {
+		// Backward compatibility: return all abilities if no args.
```
JasonTheAdams: We don't need to be worrying about BC at this point. 😄 

It looks like `WP_Abilities_Query` returns all abilities of `$args` is empty, so I think we can just use that in any case.

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,515 @@
...
+	 *   category?: string|array<string>,
+	 *   namespace?: string|array<string>,
+	 *   search?: string,
+	 *   meta?: array<string,mixed>,
+	 *   orderby?: string,
+	 *   order?: string,
+	 *   limit?: int,
+	 *   offset?: int,
+	 *   ...<string, mixed>
+	 * } $args
```
JasonTheAdams: Let's define this as a type in the class docblock, and then use `@phpstan-import-type` in the `wp_get_abilities` function. We can then set that as the `@param` type in this method, too.

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,515 @@
...
+			'limit'     => self::$no_limit,
+			'offset'    => 0,
+		);
+
+		$this->query_vars = wp_parse_args( $args, $defaults );
+
+		$this->validate_meta_arg();
+		$this->validate_orderby();
+		$this->validate_order();
+		$this->validate_pagination_args();
```
JasonTheAdams: Hmm, these validation methods are interesting. They "validate" in the sense that they check if the provided value is valid, but then fallback on a default instead of notifying of the issue in any way. Is this how the `WP_Query` class works, so we're doing this for consistency? I feel like we should at least be calling `doing_it_wrong` so people know why there's unexpected behavior.

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,515 @@
...
+
+		$getter_method    = $getter_map[ $orderby ];
+		$order_multiplier = 'DESC' === $order ? - 1 : 1;
+
+		$abilities = array_values( $abilities );
+
+		usort(
+			$abilities,
+			static function ( $a, $b ) use ( $getter_method, $order_multiplier ) {
+				return strcasecmp( $a->$getter_method(), $b->$getter_method() ) * $order_multiplier;
```
JasonTheAdams, line 481: Ohh, I like the use of `$order_multiplier` here! Clever! 😃

JasonTheAdams (changes requested):
Great work, @galatanovidiu! Left a few suggestions!

includes/abilities-api/class-wp-abilities-query.php
gziolo: Here should be fine to use `const` like here:

https://github.com/WordPress/abilities-api/blob/c1b24b8aeee6d1f67aeb2f190db38efb2377c370/includes/abilities-api/class-wp-ability.php#L29

For arrays, PHP 7.2 is unhappy, so here:

https://github.com/WordPress/abilities-api/blob/c1b24b8aeee6d1f67aeb2f190db38efb2377c370/includes/abilities-api/class-wp-ability.php#L38

It's a complex env to support WP core 😅

JasonTheAdams: I'm not sure what you mean. Why would PHP 7.2 be unhappy for arrays as constant values? I believe arrays as scalar values was introduced in PHP 5.6.

gziolo: In that case, it’s linter that is very strict and prohibits usage. That would be similar case to array short syntax.

JasonTheAdams: Can we update the linter? `const FOO = array( 'bar' => 123 );` is supported, and reducing our architecture quality to appease linting feels lame. Hahah!

includes/abilities-api/class-wp-abilities-query.php
galatanovidiu: I've tried this, but it looks like `@phpstan-import-type` doesn't work in function docblocks (only in class docblocks)

includes/abilities-api/class-wp-abilities-query.php
gziolo: As far as I’m aware that was discussed at WC US, and someone has to champion the proposal and adjustments to the codebase.

galatanovidiu: Good point! I initially implemented these as constants, but our PHP linter was flagging them with errors. To maintain compatibility with our linting rules (which appear to be stricter than PHP 7.2's actual capabilities), I converted them to private static properties.

includes/abilities-api.php
galatanovidiu: Done on [c89f085](https://github.com/WordPress/abilities-api/pull/115/commits/c89f085f8692c391193328579257f012f105e174) and [de10fa7](https://github.com/WordPress/abilities-api/pull/115/commits/de10fa7457cc796078adfbc1a9c02501e871b93f)

includes/abilities-api/class-wp-abilities-query.php
galatanovidiu: I should have used `sanitize` instead of `validate`
WP_Query also only sanitizes the args.
I can implement validation if you want, but I'd prefer to keep it simple.

Please check [68fea1b](https://github.com/WordPress/abilities-api/pull/115/commits/68fea1b8e806570f09ae65173cbca93202499818)

includes/abilities-api/class-wp-abilities-query.php
JasonTheAdams: Correct! It would go like this:
1. Add an AbilityQueryArgs type in the class docblock
2. Use it as the `@param` type for this method
3. Import and use it in the function

includes/abilities-api/class-wp-abilities-query.php
JasonTheAdams: Ah, gotcha. I'm _inclined_ to let people know when they're straight up doing things wrong (I've personally felt the consternation of figuring out that I have a typo in a query), but won't push it if it's not recommended otherwise.

includes/abilities-api/class-wp-abilities-query.php
JasonTheAdams: That's unfortunate. Is this the right place to propose the change? https://github.com/WordPress/WordPress-Coding-Standards?

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,527 @@
...
+
+/**
+ * Query class for filtering registered abilities.
+ *
+ * Provides a WordPress-standard query interface for filtering abilities by various criteria,
+ * including category, namespace, search terms, meta properties, and annotations.
+ *
+ * @since n.e.x.t
+ */
+class WP_Abilities_Query {
```
justlevine: From the docs it seems this class isn't meant to be consumed/extended directly (which IMO makes sense from a 6.9 iterative POV).

If so, can we mark the class `final` and add an `@internal` + a note pointing to `wp_get_abilities( $args )` in the doc-block? Would give us more freedom to iterate in future versions.

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,527 @@
...
+		$this->parse_query( $args );
+	}
+
+	/**
+	 * Parses and sanitizes query arguments.
+	 *
+	 * @since n.e.x.t
+	 *
+	 * @param array<string,mixed> $args Query arguments.
+	 *
```
justlevine: Nit: I'm assuming the blank end-lines in the comments throughout this file are unintentional

```suggestion
```

docs/8.querying-filtering-abilities.md
```diff
@@ -0,0 +1,213 @@
...
+    'orderby' => 'label',
+    'order'   => 'ASC', // or 'DESC'
+) );
+
+// Paginate results
+$abilities = wp_get_abilities( array(
+    'limit'  => 10,
+    'offset' => 0,
+) );
+```
```
justlevine, line 206: Is there a specific need for pagination?

I see that you're trying to mirror `WP_Query` et al (versus adding sorting/filtering as registry methods) but abilities are in a PHP array, not a database. Do we really need a shortcut for what's essentially an `array_slice()`?

justlevine (commented):
I'm concerned about using the `WP_*_Query` when that pattern is
1. used by core for DB lookups
2. often encouraged to be used directly versus getters (e.g. `new WP_Term_Query()` versus `get_terms()`. )

Neither afaik are true here and [docs](https://github.com/WordPress/abilities-api/pull/115#discussion_r2430520924) can only help so much.

From a code POV, we could also simplify implementation significantly if we weren't trying to mimic `WP_Query`. E.g. we don't need 011y, query getters, or pagination when we're just sorting/filtering an array in PHP.

includes/abilities-api/class-wp-abilities-query.php
JasonTheAdams: This is a really fair point, @justlevine. Thinking on this, I wonder if this would be a good use for introducing an `Abilities_Collection` class, instead of a querying system like this. It would implement `IteratorAggregate`, `ArrayAccess`, and such, and be what the registry stores all the abilities into.

From there, it could have methods such as `Abilities_Collection::with_category(...$categories)` which returns an array (or another `Abilities_Collection` subset). If it's the latter, then folks can chain:
```php
$abilities = wp_get_abilities()
  ->with_namespace('woo')
  ->with_category('orders');
```

This way, to your point, we're treating it more like an in-memory array rather than querying against an external system. Constructing a query is most important for efficiently retrieving the data in one go. But for something like this it's not really necessary.

What do you think, including @galatanovidiu?

includes/abilities-api/class-wp-abilities-query.php
gziolo: I found the [Core Committer meeting notes](https://make.wordpress.org/core/2025/09/18/core-committer-meeting-notes-from-wordcamp-us-2025/), which include an entire section on the need for changes in coding standards. You can try in that GitHub repository, or raise it as a topic for the weekly dev chat ([next one today](https://make.wordpress.org/core/2025/10/14/dev-chat-agenda-october-15-2025/)).

includes/abilities-api/class-wp-abilities-query.php
gziolo: If you feel this needs more time to define the API’s shape, we will have to postpone it to WordPress 7.0. This is a rather substantial change that needs some longer conversation, so contributors can have time to chime in and share their perspective on how the codebase evolves, so they can later replicate better patterns.

docs/8.querying-filtering-abilities.md
gziolo: Filtering abilities are the most crucial aspect from my perspective. Ordering and pagination are nice-to-haves, but I want to acknowledge that they fit well together in the proposed implementation, as they allow easy selection of a precise number of abilities that meet specific criteria, which might be useful when passing data to an LLM with limited context. @galatanovidiu might have a better overview of the real needs based on the work he has done for the MCP Adapter.

includes/abilities-api/class-wp-abilities-query.php
galatanovidiu: I agree with all of you.
I know we are in a hurry to have this functionality out and ready for 6.9, but I also like the idea of the Abilities Collection a lot. I've used Laravel collections a lot, and I know how natural and DX-friendly that is.

Because of that, I've prepared a POC here: https://github.com/WordPress/abilities-api/pull/119

gziolo: It's also important to consider performance. The proposal for collections will loop through once for the current set of abilities per filter. In contrast, in this PR, there is a single pass over the entire set, and all filters are applied to a single item at once. There is also an extensibility story that needs to be taken into account. `WP_Query` is at extreme with 44 occurrences of `apply_filters()` and `apply_filters_ref_array()`, but some sort of filtering will eventually need to be added, so custom filters can be executed, for example, when fetching abilities through the REST API.

gziolo:
One thing I want to note is that some filters are hard to express in this array notation. For example, a `category` or `namespace` let you provide an array of options where the existence of one of them means a match (OR strategy). In the case of `meta` where the structure is more complex, the approach used (AND strategy) makes more sense, but this comes at the cost of the developer having to be very careful when studying the docs to ensure they express their intent appropriately. However, for `meta` it isn't currently possible to request an OR condition in case at least one of several meta conditions is met.

includes/abilities-api/class-wp-abilities-query.php
justlevine: > This is a rather substantial change that needs some longer conversation,

I second @gziolo, but probably take it further than he.

While I'm a big fan of fluent APIs in general (and love how much simpler #119 reads), I don't think that us introducing the concept of a `*_Collection` into WordPress on the cusp of 6.9 is a great idea. It's also possible that the `WP_Query` parrten will turn out to be ideal in 7.x after we find a reason to have abilities be backed by something in the DB (a la plugins/themes/fonts), in which case preemptively going with a Collection pattern would be hard to untangle.

I'm hoping we can take the simplified methods from @galatanovidiu 's #119 but move them to `private function`s inside `WP_Abilities_Registry`. That would give us all the surface area we need for MVP filtering support in 6.9 (albeit via passed `$args`), while still keeping our hands free for whatever pattern we ultimately decide is best (when we have some unpressured time to reach the decision holistically).

JasonTheAdams: Thanks for weighing in, folks! And for throwing together the draft so quickly, @galatanovidiu! 😄 

Overall, I agree with the desire to not rush this. If we start with simple arrays in 6.9 with no built-in filtering (folks can always get the abilities and filter them themselves), I'm fine with that.

Regarding the introduction of the `*_Collection` concept, in my opinion, WordPress errs too much on the side of using "we haven't done that before" argument as a reason not to do things. Concepts like Collections are tried and true patterns across multiple systems. We're not creating a new concept entirely, we're porting one over. I do appreciate you raising the point, @justlevine, as it is a good one to be mindful of. I would love to see more tried-and-true practices and patterns brought _tastefully_ into WordPress over time.

With regards to performance, @galatanovidiu, in this case I think it's negligible — absolutely nothing like performance from a database perspective. And even chaining things like `with_x` is working from a smaller. subset each time, so the iteration performance jumps each time. Frankly, this isn't too unlike how databases often filter things down.

In short, I'm in favor of closing this PR to explore a better API. I really like the draft (which I need to dig into more), and would like to explore that sort of direction further. 😄

includes/abilities-api/class-wp-abilities-query.php
galatanovidiu: This is exactly what I tried and is not working.

`@phpstan-import-type` only works in class docblocks, not in function docblocks. PHPStan throws an error: "Parameter $args has invalid type AbilityQueryArgs".

The solution is to keep the full `@phpstan-param` array shape in both places:
1. Define `@phpstan-type AbilityQueryArgs` in the class docblock as the source of truth
2. Duplicate the full array shape using `@phpstan-param` in the function docblock

This reduces duplication within the class but we still need to duplicate it for standalone functions since `@phpstan-import-type` doesn't work outside of class contexts.

Please check [d3275fd](https://github.com/WordPress/abilities-api/pull/115/commits/d3275fddfdff45a9e82c36f053a9b18b9f89ddb0)

includes/abilities-api/class-wp-abilities-query.php
galatanovidiu: Done

swissspidy:
> One thing I want to note is that some filters are hard to express in this array notation. For example, a `category` or `namespace` let you provide an array of options where the existence of one of them means a match (OR strategy). In the case of `meta` where the structure is more complex, the approach used (AND strategy) makes more sense, but this comes at the cost of the developer having to be very careful when studying the docs to ensure they express their intent appropriately. However, for `meta` it isn't currently possible to request an OR condition in case at least one of several meta conditions is met.

For such cases, would it be useful being able to provide a filter callback instead of an array of options?

JasonTheAdams:
> For such cases, would it be useful being able to provide a filter callback instead of an array of options?

@swissspidy These sort of nuanced needs is what increasingly leads me to believe that having Collections in WP would be so helpful for the DX:
```php
$woo_abilities = wp_get_abilities_collection()
  ->withNamespace('woo')
  ->withCategory('woo_products')
  ->filter(static function($ability) { return $ability; });
```

Ignore the method chaining. That's not my point. Just being able to clearly retrieve subsets of abilities would be so nice to do declaratively, without making the `wp_get_abilities()` signature really complicated.

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,531 @@
...
+		$this->query_vars['limit']  = (int) $this->query_vars['limit'];
+		$this->query_vars['offset'] = (int) $this->query_vars['offset'];
+
+		// Ensure offset is non-negative.
+		if ( $this->query_vars['offset'] < 0 ) {
+			$this->query_vars['offset'] = 0;
+		}
+
+		// Ensure limit is either -1 (no limit) or positive.
+		if ( $this->query_vars['limit'] >= self::$no_limit ) {
```
jorgefilipecosta, line 179: It accepts a limit of 0, which I'm not sure if is expected.

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,531 @@
...
+	protected function filter_by_meta( WP_Ability $ability ): bool {
+		$filters = $this->query_vars['meta'];
+
+		if ( empty( $filters ) ) {
+			return true;
+		}
+
+		$ability_meta = $ability->get_meta();
+
+		[ $flat_filters, $nested_filters ] = $this->separate_meta_filters( $filters );
```
jorgefilipecosta, line 311: Normally in WordPoress core, list is used on these cases, I'm not sure if in core this array Destructuring is allowed, but I did not found any case there so I think it is not allowed.

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,531 @@
...
+			return;
+		}
+
+		$this->query_vars['limit'] = self::$no_limit;
+	}
+
+	/**
+	 * Retrieves the filtered abilities based on query arguments.
+	 *
+	 * @return \WP_Ability[] Array of filtered abilities.
```
jorgefilipecosta, line 189: @return normally appears after @since

includes/abilities-api/class-wp-abilities-query.php
```diff
@@ -0,0 +1,531 @@
...
+		);
+
+		if ( ! isset( $getter_map[ $orderby ] ) ) {
+			return $abilities;
+		}
+
+		$getter_method    = $getter_map[ $orderby ];
+		$order_multiplier = 'DESC' === $order ? - 1 : 1;
+
+		$abilities = array_values( $abilities );
```
jorgefilipecosta, line 476: I think this array_values is not needed, usort just sorts the values and ignores keys.

gziolo:
@swissspidy and @jorgefilipecosta, I see you started reviewing this PR, so it'd be important to mention that work on this PR stalled when an alternative API was proposed:
- https://github.com/WordPress/abilities-api/pull/119

The next step would be to evaluate both approaches and decide on the best path forward.

PR #121, opened 2025-10-15
Add category listing and validation to client package

emdashcodes:
This PR adds improved category support to the client package. It adds methods for listing categories (`getAbilityCategories`, `getAbilityCategory`), methods for registering and unregistering categories (`registerAbilityCategory`, `unregisterAbilityCategory`), and it validates ability categories during ability registration.

Follows up on #102 (categories system), #114 (basic category support), and #120 (REST endpoints).

## Why

In the previous client PR we only addressed filtering abilities by category slug but had no way to discover available categories or validate that a category exists before registering an ability. There was also no way to register a category only on the client side.

## How

- Added `getAbilityCategories()` function to fetch all categories
- Added `getAbilityCategory(slug)` function to fetch a single category
- Added `registerAbilityCategory` to register a category on the client side
- Added `unregisterAbilityCategory` to remove a category on the client side
- Added resolvers that automatically fetch categories from the REST API
- Enhanced validation to check category existence, not just format
- Added unit tests for all new selectors, resolvers, and actions
- Updated README with usage examples for all new functions

## Breaking Change

- **Breaking Change** - `registerAbility()` now returns `Promise<void>` instead of `void`. This is. to. ensure that categories are loaded on first `registerAbility()` call if store is empty. Since categories are already a breaking change, I feel like this is OK. It does still work without awaiting, but it's possible the categories are not loaded before the register / validation check.

## Testing Steps

1. Run `npm run test:client`
2. Register a few test categories and abilities
3. Open WordPress admin in browser with dev tools console
4. Test `getAbilityCategories()`:

   ```javascript
   const categories = await wp.abilities.getAbilityCategories();
   console.log(categories);
   ```

5. Test `getAbilityCategory()`:

   ```javascript
   const nav = await wp.abilities.getAbilityCategory('navigation');
   console.log(nav);

   const missing = await wp.abilities.getAbilityCategory('does-not-exist');
   console.log(missing);
   ```

6. Test `registerAbility()` validation:

Test with a valid category:

   ```javascript
   await wp.abilities.registerAbility({
     name: 'test/my-ability',
     label: 'Test Ability',
     description: 'A test ability',
     category: 'navigation', // Should validate successfully
     callback: () => ({ success: true })
   });
   ```

Test with an invalid category:

   ```javascript
   await wp.abilities.registerAbility({
     name: 'test/invalid',
     label: 'Test',
     description: 'Test',
     category: 'does-not-exist', // Should throw validation error
     callback: () => ({})
   });
   // Expected error: 'Ability "test/invalid" references non-existent category "does-not-exist". Please register the category first.'
   ```

7. Test registering a category and assigning some abilities to it:

```javascript
  // Register category
  await wp.abilities.registerAbilityCategory('block-editor', {
    label: 'Block Editor',
    description: 'Abilities for interacting with the block editor',
    meta: { icon: 'dashicons-editor-code' }
  });

  // Register multiple abilities
  await wp.abilities.registerAbility({
    name: 'my-plugin/insert-paragraph',
    label: 'Insert Paragraph',
    description: 'Inserts a paragraph block',
    category: 'block-editor',
    callback: async ({ content }) => {
      return { success: true, content };
    }
  });

  await wp.abilities.registerAbility({
    name: 'my-plugin/insert-heading',
    label: 'Insert Heading',
    description: 'Inserts a heading block',
    category: 'block-editor',
    callback: async ({ level, text }) => {
      return { success: true, level, text };
    }
  });

  // Get all abilities in this category
  const allAbilities = await wp.abilities.getAbilities({ category: 'block-editor' });
  console.log('Abilities in block-editor category:', allAbilities);
```

8. Test unregistering a category

```javascript
  wp.abilities.unregisterAbilityCategory('block-editor');
  console.log('Category unregistered!');
```


JasonTheAdams (commented):
This is looking pretty good! One question comes to mind: Should it be possible to register categories here the same way abilities can? If a user is registering abilities on the front-end, is it also possible they may be introducing new categories of abilities? I would think so, actually, in such cases as interacting with the block editor.

A category here obviously shouldn't conflict with a category already registered from the server. But I think that's about it.

emdashcodes:
> One question comes to mind: Should it be possible to register categories here the same way abilities can? If a user is registering abilities on the front-end, is it also possible they may be introducing new categories of abilities? I would think so, actually, in such cases as interacting with the block editor.

Yeah, I think that we should allow for this, and that there's going to be a lot of use cases where a set of abilities is just all client-side. 

I went ahead and added support for this here: https://github.com/WordPress/abilities-api/pull/121/commits/cc55e76eabe5e7d29028fd3e6439cdeb536ac9b7. I have also updated the docs and the PR description to include the new changes.

packages/client/src/store/selectors.ts
```diff
@@ -45,3 +53,29 @@ export function getAbility(
...
 	return state.abilitiesByName[ name ] || null;
 }
+
+/**
+ * Returns all registered ability categories.
+ *
+ * @param state Store state.
+ * @return Array of categories.
+ */
+export function getAbilityCategories(
```
gziolo: Similar to `getAbilities` selector, this one should also be wrapped with `createSelector` to memoize result and prevent re-rendering in React components.

packages/client/src/store/actions.ts
```diff
@@ -26,19 +31,33 @@ export function receiveAbilities( abilities: Ability[] ) {
...
+ * REST API if they haven't been loaded yet.
  *
  * @param  ability The ability to register.
  * @return Action object or function.
  * @throws {Error} If validation fails.
  */
 export function registerAbility( ability: Ability ) {
 	// @ts-expect-error - registry types are not yet available
-	return ( { select, dispatch } ) => {
+	return async ( { select, dispatch } ) => {
```
gziolo, line 60: This reminds me that on the server the same name of the ability should not be registered twice. How does it work on the client? In theory, we could allow overriding server abilities to include different execute handling, but this might require additional permissions handling on the client. I don’t think we have that in place. Maybe the simplest way forward for now id to disallow registration twice under the same name with error handling. I have some recollection that it’s correctly handled in the reducer.

packages/client/src/store/actions.ts
```diff
@@ -81,6 +98,28 @@ export function registerAbility( ability: Ability ) {
 			);
 		}
 
+		// Ensure categories are loaded before validating
+		let categories = select.getAbilityCategories();
```
gziolo: Would it be possible to use `hasFinishedResolution` or is it internal? I need to check.

gziolo: I’m not entirely sure it’s the best approach to check for the value and then await for the result. Isn’t `resolveSelect` a blocking version of the selector that returns data?

gziolo (commented):
This looks close to ready. I did first pass of review of production code and it looks solid 👍

packages/client/src/store/actions.ts
emdashcodes: > How does it work on the client? 

At the moment it matches the server. We return an error if you try to register a duplicate Ability: 

https://github.com/WordPress/abilities-api/blob/118cd0a741dc83fff1a8602c8e1a16c61a96afb7/packages/client/src/store/actions.ts#L98

>  In theory, we could allow overriding server abilities to include different execute handling, but this might require additional permissions handling on the client. I don’t think we have that in place.

We don't at the moment. It was mentioned as a potential feature in the original issue (https://github.com/WordPress/abilities-api/issues/41) but we didn't add that.

gziolo: Ok, perfect. It’s hard to identify the logic with these four components in the store 😅

I will file an enhancement issue for the client overrides if I don’t forget.

packages/client/src/store/actions.ts
emdashcodes: Addressed this in https://github.com/WordPress/abilities-api/pull/121/commits/da86104662ef10c0a3f11bed8574c74f58b6921c

packages/client/src/store/selectors.ts
emdashcodes: Addressed this in https://github.com/WordPress/abilities-api/pull/121/commits/da86104662ef10c0a3f11bed8574c74f58b6921c

emdashcodes:
@gziolo Thanks for the review! I've pushed a couple final changes here.

gziolo:
I’ll have another look later. The follow up changes look good after quick check 👍

packages/client/src/store/actions.ts
```diff
@@ -119,3 +152,94 @@ export function unregisterAbility( name: string ) {
...
+		}
+
+		// Ensure categories are loaded before checking for duplicates
+		await resolveSelect( STORE_NAME ).getAbilityCategories();
+		const existingCategory = select.getAbilityCategory( slug );
+		if ( existingCategory ) {
+			throw new Error(
+				sprintf( 'Category "%s" is already registered.', slug )
+			);
+		}
```
JasonTheAdams, line 192: @emdashcodes Just double-checking: It's this `await resolveSelect` here that ensures the server categories are loaded prior to checking for a collision, right?

emdashcodes: Yup! See https://developer.wordpress.org/block-editor/reference-guides/packages/packages-data/#resolveselect. The promise here will resolve once the [resolvers fetch from the entity store](https://github.com/WordPress/abilities-api/pull/121/files#diff-438b0daec0a76675839e22930b599072aa368c76cbde4ecdba4694d1a5f14b61), which will fetch from the REST API if needed.

packages/client/README.md
```diff
@@ -100,14 +100,39 @@ if ( ability ) {
...
+
+```javascript
+const category = await getAbilityCategory( 'data-retrieval' );
+if ( category ) {
+  console.log( `Found category: ${ category.label }` );
+  console.log( `Description: ${ category.description }` );
+}
+```
+
+#### `registerAbility( ability: Ability ): Promise<void>`
```
gziolo, line 128: We will need an update in another document the example with missing `await`, too:

https://github.com/WordPress/abilities-api/blob/trunk/docs/7.javascript-client.md#registerability-ability-

A larger question for later is how to avoid redundancy between these two documents. Is this one auto-generated from JSDoc?

packages/client/src/store/resolvers.ts
```diff
@@ -74,3 +74,56 @@ export function getAbility( name: string ) {
...
+/**
+ * Resolver for getAbilityCategory selector.
+ * Fetches a specific category from the server if not already in store.
+ *
+ * @param slug Category slug.
+ */
+export function getAbilityCategory( slug: string ) {
+	// @ts-expect-error - registry types are not yet available
+	return async ( { dispatch, registry, select } ) => {
+		const existingCategory = select.getAbilityCategory( slug );
```
gziolo, line 122: If I follow it correctly, this makes sense for the category registered on the client, so it doesn't necessarily trigger a network request. It could use a better explanation, as it took me some time to conclude.

packages/client/src/store/resolvers.ts
```diff
@@ -74,3 +74,56 @@ export function getAbility( name: string ) {
...
+/**
+ * Resolver for getAbilityCategories selector.
+ * Fetches all categories from the server.
+ *
+ * The resolver only fetches once and stores all categories.
+ */
+export function getAbilityCategories() {
+	// @ts-expect-error - registry types are not yet available
+	return async ( { dispatch, registry, select } ) => {
+		const existingCategories = select.getAbilityCategories();
```
gziolo, line 87: It's probably similar to https://github.com/WordPress/abilities-api/pull/121/files#r2437266219, but here I have some concerns. If someone registers first a single category from the client, then this resolver will never hit the server, as it will find an entry. 

One solution would be to have two selectors, one private (or explicitly tied to the server) for network requests without any conditions letting the framework handle caching. Another selector wouldn't have the matching resolver, but would combine categories from the client and the server together.

Another option, filter out client-only categories in the check.

gziolo (commented):
It's a bit complicated to handle entities for abilities, and ability categories, while it's possible to register more items on the client for both registries. We probably need a small tweak as outlined in https://github.com/WordPress/abilities-api/pull/121/files#r2437279266, but it also raises a larger question what would be the ideal design to take advantage of the fact that these entities are already nicely abstracted, and maybe, we could avoid using these resolvers, and instead make selectors more complex with direct calls that resolvers do today. I have no idea if that is feasible, but I am leaving a note to revisit later.

packages/client/src/store/resolvers.ts
emdashcodes: See https://github.com/WordPress/abilities-api/pull/121/commits/91d8bf8fd7131bedb38400a2f2199291c72d6442. I opted for the 3rd option, `filter out client-only categories in the check` which matches how we are handling `getAbilities` as well.

packages/client/src/store/resolvers.ts
emdashcodes: I added a clarifying comment here.

packages/client/README.md
emdashcodes: Thanks, I updated this document with all the new methods and updated the await in the example.

> A larger question for later is how to avoid redundancy between these two documents. Is this one auto-generated from JSDoc?

This one is not auto-generated. I'm not sure the best way to keep them in sync to be honest.

I do think it's valuable for the client package README to have a function definition and such. Especially if it gets split off into Gutenberg. But we also don't want to duplicate stuff in multiple places obviously..

gziolo: Gutenberg has tooling in place that allows auto-generating the list of public methods based on the JSDoc, so that would cover the README file nicely, as these methods are very well documented.

gziolo (approved):
No more feedback from my side. I'm glad to see we reached feature parity with the server implementaiton for categories 🎉

PR #122, opened 2025-10-15
Adding ability_class information and example

jonathanbossenger:
WIP, needs to be tested to verify.

### DIFF
--- a/docs/php-api.md
+++ b/docs/php-api.md
@@ -384,55 +384,18 @@ function my_plugin_register_send_email_ability() {
 
 #### Registering an Ability with a Custom Ability Class
 
-The `ability_class` parameter allows you to use a custom class that extends `WP_Ability`. This is useful when you need to add custom methods to your ability or override the default behavior of the base `WP_Ability` class.
-
-**When to use a custom ability class:**
-- You need to add custom helper methods specific to your ability
-- You want to override how the ability executes (e.g., add logging, caching, or custom validation)
-- You need to store additional state or configuration beyond what the standard metadata provides
-- You want to encapsulate complex business logic within the ability class itself
+The `ability_class` parameter allows you to use a custom class that extends `WP_Ability`. This is useful when you want to extend the default behavior of the base `WP_Ability` class.
 
 **Example: Creating a custom ability class with additional methods**
 
 ```php
 /**
- * Custom ability class that adds a validation helper method.
+ * Custom ability class that adds logging.
  *
  * This example shows how to extend WP_Ability to add custom behavior
  * while still leveraging all the standard ability functionality.
  */
 class My_Plugin_Post_Validator_Ability extends WP_Ability {
-
-    /**
-     * Custom method to check if a post ID is valid and accessible.
-     *
-     * This helper method can be called from the execute callback or permission callback
-     * to perform validation logic that's specific to this ability.
-     *
-     * @param int $post_id The post ID to validate.
-     * @return true|WP_Error True if valid, WP_Error if invalid.
-     */
-    public function validate_post_id( $post_id ) {
-        // Check if the post exists
-        $post = get_post( $post_id );
-        if ( ! $post ) {
-            return new WP_Error(
-                'invalid_post',
-                __( 'The specified post does not exist.', 'my-plugin' )
-            );
-        }
-
-        // Check if the post is published
-        if ( 'publish' !== $post->post_status ) {
-            return new WP_Error(
-                'post_not_published',
-                __( 'The specified post is not published.', 'my-plugin' )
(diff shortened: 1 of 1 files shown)


gziolo (commented):
I read the changes proposed and everything seemed to be correct. Nice understanding of the use cases it tries to cover. It’s also primarily for projects offering collections of abilities that share similar characteristics like logging, debugging, social permission checks, etc.

docs/php-api.md
```diff
@@ -381,6 +382,176 @@ function my_plugin_register_send_email_ability() {
...
+    ));
+}
+```
+
+**Important notes about custom ability classes:**
+
+- Your custom class **must** extend `WP_Ability`
+- The custom class is only used to instantiate the ability - the `ability_class` parameter is not stored as a property of the ability
+- You can override protected methods like `do_execute()`, `validate_input()`, or `validate_output()` to customize behavior
+- You can add public methods to provide additional functionality specific to your ability
```
justlevine: They can add any sort of method, not sure if we want to encourage public methods with an explicit callout.
```suggestion
- You can add custom methods to provide additional functionality specific to your ability
```

docs/php-api.md
```diff
@@ -381,6 +382,176 @@ function my_plugin_register_send_email_ability() {
...
+    /**
+     * Custom method to check if a post ID is valid and accessible.
+     *
+     * This helper method can be called from the execute callback or permission callback
+     * to perform validation logic that's specific to this ability.
+     *
+     * @param int $post_id The post ID to validate.
+     * @return true|WP_Error True if valid, WP_Error if invalid.
+     */
+    public function validate_post_id( $post_id ) {
```
justlevine: This example seems counterintuitve at a glance. I wanna give it some more thought in the morning but this either should be in the permission callback or in our overloaded `do_execute()`. If we want to give an example of a public method being called on the instance inside of inside our extended class, we might want to think of a use case that better justifies it, maybe something stateful?

docs/php-api.md
```diff
@@ -381,6 +382,176 @@ function my_plugin_register_send_email_ability() {
...
+    }
+
+    /**
+     * Override the do_execute method to add custom logging.
+     *
+     * This demonstrates how you can override methods from WP_Ability
+     * to customize behavior before or after the standard execution.
+     *
+     * @param mixed $input Optional. The input data for the ability.
+     * @return mixed|WP_Error The result of the ability execution.
```
justlevine:
```suggestion
     * @return mixed|\WP_Error The result of the ability execution.
```

docs/php-api.md
```diff
@@ -381,6 +382,176 @@ function my_plugin_register_send_email_ability() {
...
+class My_Plugin_Post_Validator_Ability extends WP_Ability {
+
+    /**
+     * Custom method to check if a post ID is valid and accessible.
+     *
+     * This helper method can be called from the execute callback or permission callback
+     * to perform validation logic that's specific to this ability.
+     *
+     * @param int $post_id The post ID to validate.
+     * @return true|WP_Error True if valid, WP_Error if invalid.
```
justlevine:
```suggestion
     * @return true|\WP_Error True if valid, WP_Error if invalid.
```

docs/php-api.md
```diff
@@ -381,6 +382,176 @@ function my_plugin_register_send_email_ability() {
...
+     */
+    public function validate_post_id( $post_id ) {
+        // Check if the post exists
+        $post = get_post( $post_id );
+        if ( ! $post ) {
+            return new WP_Error(
```
justlevine:
```suggestion
            return new \WP_Error(
```

docs/php-api.md
```diff
@@ -381,6 +382,176 @@ function my_plugin_register_send_email_ability() {
...
+                'invalid_post',
+                __( 'The specified post does not exist.', 'my-plugin' )
+            );
+        }
+
+        // Check if the post is published
+        if ( 'publish' !== $post->post_status ) {
+            return new WP_Error(
```
justlevine:
```suggestion
            return new \WP_Error(
```

docs/php-api.md
justlevine: Alternatively, this could be a lovely place to demonstrate `{@inheritDoc}` and leave off the types annotations for better future compat

justlevine (commented):
Did a quick scan from my phone. mostly nits

jonathanbossenger:
@gziolo @justlevine sorry, this wasn't mean to be reviewed yet, still a WIP.

jonathanbossenger:
Merging this into the proper docs/prepare-6.9 branch so it can all be reviewed there.

PR #123, opened 2025-10-16
Switch to Classmap for bootstrap autoloading

JasonTheAdams:
It was pointed out that using `require_once` in `boostrap.php` was causing this plugin to not work properly with the Jetpack Autoloader, which relies on the classmap to control loading the latest version when both are present (0.2.0 and 0.3.0, for instance).

This PR addresses this by switching to using a Classmap for the autoloading of classes. I'm not terribly familiar with the nuances of how Jetpack autloader works, so I _think_ I got this right. I added a check for using the built-in autoloader as well, so this can still work standalone. I'm hoping that doesn't cause issues with the Jetpack autoloader.

### DIFF
--- a/abilities-api.php
+++ b/abilities-api.php
@@ -26,5 +26,9 @@
  */
 define( 'WP_ABILITIES_API_DIR', plugin_dir_path( __FILE__ ) );
 
+// Load Composer autoloader if not already loaded (for standalone plugin usage).
+if ( file_exists( WP_ABILITIES_API_DIR . 'vendor/autoload.php' ) ) {
+	require_once WP_ABILITIES_API_DIR . 'vendor/autoload.php';
+}
 
 require_once WP_ABILITIES_API_DIR . 'includes/bootstrap.php';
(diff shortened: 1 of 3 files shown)


gziolo:
I’m curious how this is going to work when WP core 6.9 beta once it defined the same classes and functions. The current defensive approach tries to prevent redeclaration issues. So probably worth testing it all together.

JasonTheAdams:
> I’m curious how this is going to work when WP core 6.9 beta once it defined the same classes and functions. The current defensive approach tries to prevent redeclaration issues. So probably worth testing it all together.

That's a very good point. We're discussing this in [the Core AI Slack meet, now](https://wordpress.slack.com/archives/C08TJ8BPULS/p1760631618839829).

JasonTheAdams:
I'm going to close this for now as I think it's become a moot point. I won't delete the branch for now in case we want to revive it.

gziolo:
I was too quick in deleting the branch. I used the UI to restore it 😄

PR #126, opened 2025-10-22
Sync changes from WordPress core after 6.9 beta 1

### LINKED ISSUE #112: Align doing_action() and did_action() usage for category and ability registration in the Abilities API
There’s an inconsistency in how the Abilities API handles the action checks for category and ability registration.  

**Current behavior:**
- **Categories:**  
  - The `doing_action( 'abilities_api_categories_init' )` check is performed **directly** inside the `register()` method.  
  - This enforces registration **only** during the `abilities_api_categories_init` action.
- **Abilities:**  
  - The check is performed in the **wrapper** function `wp_register_ability()`.  
  - It uses `did_action()` instead, allowing registration even **after** the action has fired.

**Discussion points:**
- Should both follow the same pattern for consistency across the API?  
- Should registration be restricted strictly to the init action (`doing_action()`), or remain flexible using `did_action()`?  
- Which level (core `register()` method vs wrapper function) should enforce this restriction?

_Originally posted by @gziolo in https://github.com/WordPress/abilities-api/pull/102#discussion_r2425780133_

gziolo:
Closes https://github.com/WordPress/abilities-api/issues/112 – action names changes, and the strategy got unified.
Closes https://github.com/WordPress/abilities-api/pull/97 – PHPStan was removed as WP core doesn't support it.

I did an initial pass for syncing back all the changes applied to WordPress Core:
- https://github.com/WordPress/wordpress-develop/commit/cd95bd2338a6bbdfe474c01373af39034e1d2a44 (https://github.com/WordPress/wordpress-develop/pull/9410)
- https://github.com/WordPress/wordpress-develop/commit/871fd0e997e411221b1b63352f753ab5653eb6f0 (https://github.com/WordPress/wordpress-develop/pull/10380)
- https://github.com/WordPress/wordpress-develop/commit/409d03b70d5cb1176bf01c1eb65bb8ee5a1b7f53 (https://github.com/WordPress/wordpress-develop/pull/10395)

We need to verify the linting errors reported as some of them might be valid. For example:
- https://github.com/WordPress/abilities-api/pull/126/commits/0c7f6215800890578a14c5f6e0497ccda365dea5 – this fixes a potential issue reported where the instance could be `null`.
- https://github.com/WordPress/abilities-api/pull/126/commits/76f1fad51fb14aa4f8512d7bea2384732c3fe837 – PHPCS reported wrong order of types, as it should be `boolean|null`. It's worth updating in WordPress core, too.

There are several linting errors that are expected in WordPress core, so we have to disable them. However, the one about best practices for escaping translations was a bit surprising. It still has to be confirmed, as it isn't enforced in WP core and I simply followed feedback from review when updated `esc_html__()` to `__()`. It's mostly for `_doing_it_wrong()` calls, so maybe that's perfectly fine.

As part of the effort, I had to redo `WP_REST_Abilities_Init`, which is file-specific only to this repository. I added a check to ensure that the controllers are never re-declared or re-registered if they were previously registered in WordPress core.


justlevine:
@gziolo on the off chance I save you 30 seconds, here are the FQCN rules for phpcs that need to be removed . Considering the current plan is to archive, I see no point in conditionally disabling/allowlisting.

- https://github.com/WordPress/abilities-api/blob/9da857ea6c1141f6dca55f130614c61fb0c6f197/phpcs.xml.dist#L141-L154

includes/rest-api/class-wp-rest-abilities-init.php
```diff
@@ -22,17 +22,22 @@ class WP_REST_Abilities_Init {
...
 	 */
 	public static function register_routes(): void {
-		require_once __DIR__ . '/endpoints/class-wp-rest-abilities-run-controller.php';
-		require_once __DIR__ . '/endpoints/class-wp-rest-abilities-list-controller.php';
-		require_once __DIR__ . '/endpoints/class-wp-rest-abilities-categories-controller.php';
-
-		$categories_controller = new WP_REST_Abilities_Categories_Controller();
+		if ( ! class_exists( 'WP_REST_Abilities_V1_Categories_Controller' ) ) {
+			require_once __DIR__ . '/endpoints/class-wp-rest-abilities-v1-categories-controller.php';
+		}
```
JasonTheAdams: I'm curious why we're requiring these conditionally? Are we intending for someone to load their on controller first?

includes/abilities-api/class-wp-ability-categories-registry.php
```diff
@@ -0,0 +1,254 @@
...
+		}
+
+		if ( ! preg_match( '/^[a-z0-9]+(?:-[a-z0-9]+)*$/', $slug ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				__( 'Ability category slug must contain only lowercase alphanumeric characters and dashes.' ),
+				'6.9.0'
+			);
+			return null;
+		}
```
JasonTheAdams, line 75: Are we not checking for the category init hook anymore?

includes/abilities-api.php
```diff
@@ -6,47 +6,47 @@
...
+ *         @type array<string, null|bool> $annotations  Optional. Annotation metadata for the ability.
+ *         @type bool                     $show_in_rest Optional. Whether to expose this ability in the REST API. Default false.
+ *     }
+ *     @type string               $ability_class       Optional. Custom class to instantiate instead of WP_Ability.
+ * }
+ * @return WP_Ability|null An instance of registered ability on success, null on failure.
  */
 function wp_register_ability( string $name, array $args ): ?WP_Ability {
-	if ( ! did_action( 'abilities_api_init' ) ) {
+	if ( ! did_action( 'wp_abilities_api_init' ) ) {
```
JasonTheAdams, line 272: I'm curious why we're using `did_action()` and not `doing_action`? Is an action considered "done" immediately after `do_action()` is triggered? Or after the last hook function is fired?

JasonTheAdams (commented):
Just a few questions for you, @gziolo!

includes/rest-api/class-wp-rest-abilities-init.php
gziolo: Yes, this accounts for WP Core that loads these classes now in trunk. It’s the only place in code that didn’t guard against redeclaration correctly.

includes/abilities-api/class-wp-ability-categories-registry.php
gziolo: It was moved to the `wp_register_ability_category()` and we switched to `did_action( ‘wp_abilities_api_categories_init’ )`. It works exactly the same as for abilities now. 

I also raised concerns during the merge about the previous implementation. The issue is that `wp_register_ability_category()` internally calls `get_instance()`, which triggers the `wp_abilities_api_categories_init` action before the registry is created and returned on the first run.

This means that if someone tries to register an ability category directly—without using `add_action`—the code first needs to retrieve the shared registry before calling register and checking whether the action has been processed. As a result, the public method ends up unintentionally satisfying these conditions, creating unexpected behavior.

includes/abilities-api.php
gziolo: Yes, https://developer.wordpress.org/reference/functions/do_action/ does two things, increments the counter for how many times the action was fired, and sets the current action name. Then it runs all added actions. `did_action` ensures that registration doesn’t happen too early and only when we know the registry is necessary for a given HTTP request, but the point was raised we should not prevent registering more abilities and categories later in the flow, by using `doing_action`.

gziolo:
PHPCS and PHPStan are now good after not-matching rules were removed or ignored... One unit test to fix, and then we need to update the client to use the new REST API endpoint paths. Potentially, we also need to adjust the types to account that `meta.annotations` has slightly changed.

includes/rest-api/class-wp-rest-abilities-init.php
JasonTheAdams: Ohh, so it's not this way in WP Core, then? I was assuming this sync was verbatim. It would make sense for this not to be so it yields to Core when they're running side-by-side. Am I correct in my understanding?

includes/abilities-api/class-wp-ability-categories-registry.php
JasonTheAdams: Ok, I think I'm tracking. Let me say it back to make sure I'm following:
* Checking here requires that the registry is instantiated (obviously), meaning `get_instance()` already ran which triggered the hook, therefore it had always run before this function was called
* Moving the check to `wp_register_ability_category()` means the hook can be checked _before_ the registry is tentatively instantiated

In the "WP Way" this makes sense. It's admittedly why I'm not a fan of hooks being coupled to instantiation in some way, as it makes these sort of weird conditions. The only other way I could think to do this in WP would be to either:
1. Procedurally place the hooks somewhere in the WP bootstrapping process
2. Add a hook to `init` (or somewhere) that instantiates the registries and calls the hooks

includes/abilities-api.php
JasonTheAdams: Ahhh, thanks for clearing that up. I really should have looked up the code myself. Good to know how that works!

Personally, I'm not convinced on the merits of allowing folks to register at any point after this time, as those are the sort of moments in my WP dev history that are maddening – "wait, why isn't that available? Oh, they hook into _that_ hook later with a priority of 500? Why!?" Not being able to assume that _all_ abilities and categories have been initialized during... initialization is weird. Feels like an established WP quirk that we're perpetuating. 😆

includes/rest-api/class-wp-rest-abilities-init.php
gziolo: WP core will always load first, so it doesn't make sense to add any checks on whether the class is defined there. I'm not sure I fully understand the question.

JasonTheAdams: Hahah! It's not a question, really, just me making sure I'm following. I'm saying I thought this sync PR was a verbatim copy of what's been done in core — including this code. So I was imagining this exact code in core, which was confusing me. But this code is actually distinct to this repo, if I'm following, and is in place to yield to core when loaded together.

gziolo: Yes, this file is specific to the Composer package only. Maybe we should better separate what gets copied over with no changes between WP core and this repo.

includes/abilities-api/class-wp-ability-categories-registry.php
gziolo: Both proposed solutions remove most of the benefits of the current approach because it means all abilities will get registered on every page request when this init fires (no matter where and how it gets initiated). In particular, when rendering pages for visitors, you don't want to register all abilities if they will never be consumed, which I assume is going be the default case in WordPress core. If we don't care about the impact of that registration, then we could simply use `init` hook as the only mechanism for ensuring the abilities and their categories are registered at the right time.

includes/rest-api/class-wp-rest-abilities-init.php
JasonTheAdams: Not a bad idea, just to keep the purpose of PRs like this clear. Thanks for clarifying!

includes/abilities-api/class-wp-ability-categories-registry.php
JasonTheAdams: That's interesting! I didn't realize we were using subsequent hooks as a way of narrowing the registered abilities by context. I don't disagree, it just wasn't a means of "filtering" I'd considered, yet. My assumption so far was that, yes, _all_ abilities would be registered and then subsequent filtering would occur. There are certainly trade-offs of both approaches.

JasonTheAdams: Just to help me wrap my head around the merits, here. Can you give me an example of an ability that would be best registered in a specific hook context?

gziolo:
It should be mostly ready to review.

The only remaining task is to mirror the HTTP method selection when executing the ability, and cover that with unit tests

Client:

https://github.com/WordPress/abilities-api/blob/9da857ea6c1141f6dca55f130614c61fb0c6f197/packages/client/src/api.ts#L276

Server:

https://github.com/WordPress/abilities-api/blob/0a511e04d0fe031fb2eba0a840eb3595707b2cb3/includes/rest-api/endpoints/class-wp-rest-abilities-v1-run-controller.php#L111-L116

gziolo:
It tests correctly when using the client-side API with core abilities:

Screenshot 2025-10-23 at 20 42 57

gziolo:
It turned out to be a simple task to add support for `DELETE` HTTP method also on the client with https://github.com/WordPress/abilities-api/pull/126/commits/b70eec3bf3bcb133e1ddb77ba951517598b56b40.

includes/abilities-api/class-wp-ability-category.php
```diff
@@ -184,20 +195,22 @@ public function get_meta(): array {
...
 	 *
-	 * @since 0.3.0
-	 * @throws \LogicException If the category is serialized. This is a security hardening measure to prevent serialization of the category.
+	 * @since 6.9.0
+	 * @throws LogicException If the ability category object is serialized.
+	 *                        This is a security hardening measure to prevent serialization of the ability category.
 	 */
 	public function __sleep(): array {
-		throw new \LogicException( self::class . ' must not be serialized.' );
+		throw new LogicException( __CLASS__ . ' should never be serialized' );
```
jorgefilipecosta: I guess we should leave the "." here given that __wakeup has it.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -208,20 +231,20 @@ public function is_registered( string $name ): bool {
...
 	 * @param string $name The name of the registered ability, with its namespace.
-	 * @return ?\WP_Ability The registered ability instance, or null if it is not registered.
+	 * @return ?WP_Ability The registered ability instance, or null if it is not registered.
 	 */
 	public function get_registered( string $name ): ?WP_Ability {
 		if ( ! $this->is_registered( $name ) ) {
 			_doing_it_wrong(
 				__METHOD__,
 				/* translators: %s: Ability name. */
 				sprintf( esc_html__( 'Ability "%s" not found.' ), esc_attr( $name ) ),
```
jorgefilipecosta: Not related to this changes but I guess instead of esc_attr( $name ) we should have esc_html( $name ), as name is not used as an html attribute but as a child of an html element.

includes/abilities-api.php
```diff
@@ -55,116 +55,206 @@ function wp_register_ability( string $name, array $args ): ?WP_Ability {
 				'<code>abilities_api_init</code>',
```
jorgefilipecosta: The hook name here should be wp_abilities_api_init correct?

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -233,29 +256,40 @@ public function get_registered( string $name ): ?WP_Ability {
...
-	 * @return \WP_Abilities_Registry The main registry instance.
+	 * @return WP_Abilities_Registry|null The main registry instance, or null when `init` action has not fired.
 	 */
-	public static function get_instance(): self {
+	public static function get_instance(): ?self {
+		if ( ! did_action( 'init' ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				sprintf(
+					__( 'Ability API should not be initialized before the <code>init</code> action has fired' )
```
jorgefilipecosta: Nitpick: We are missing a period at the end of the sentence.

includes/abilities-api/class-wp-abilities-registry.php
```diff
@@ -208,20 +231,20 @@ public function is_registered( string $name ): bool {
...
 	 * Do not use this method directly. Instead, use the `wp_get_ability()` function.
 	 *
-	 * @since 0.1.0
+	 * @since 6.9.0
 	 *
 	 * @see wp_get_ability()
 	 *
 	 * @param string $name The name of the registered ability, with its namespace.
-	 * @return ?\WP_Ability The registered ability instance, or null if it is not registered.
+	 * @return ?WP_Ability The registered ability instance, or null if it is not registered.
```
jorgefilipecosta: In all the other cases on this file we are using WP_Ability|null.

includes/abilities-api.php
```diff
@@ -1,170 +1,607 @@
...
+ * Example:
+ *
+ *     // Prints information about all available ability categories.
+ *     $ability_categories = wp_get_ability_categories();
+ *     foreach ( $ability_categories as $ability_category ) {
+ *         echo $ability_category->get_label() . ': ' . $ability_category->get_description() . "\n";
+ *     }
  *
- * @see WP_Abilities_Category_Registry::get_all_registered()
+ * @since 6.9.0
```
jorgefilipecosta, line 592: On since tags we are using 6.9.0, everywhere, should we be using the plugin version and not the core version?

includes/abilities-api/class-wp-ability-categories-registry.php
```diff
@@ -0,0 +1,254 @@
...
+	 *
+	 * @since 6.9.0
+	 *
+	 * @return WP_Ability_Categories_Registry|null The main registry instance, or null when `init` action has not fired.
+	 */
+	public static function get_instance(): ?self {
+		if ( ! did_action( 'init' ) ) {
+			_doing_it_wrong(
+				__METHOD__,
+				sprintf(
```
jorgefilipecosta, line 207: This sprintf is unnecessary we are not using its functionality.

jorgefilipecosta (approved):
I left multiple comments, but all of them seem minor, and this is a sync so we may need to reflect what is in core and then update things on both places.
One thing I think we should consolidate is the since tags, on if we should use core version or plugin version.

Things worked well on the smoke tests I the code seems good 👍

includes/abilities-api.php
gziolo: I would keep files the same as in WP core, as we it's a direct copy. In the future, WP core becomes the source of truth and we would only continue keeping this repo up to date so folks can polyfill Abilities API for WP 6.8 and older if they want to start using it but their support policy includes a range of WP major versions. It's been the case for Woo that is actively testing usage of this API.

includes/abilities-api.php
gziolo: I fixed that in WP core today and copied here with https://github.com/WordPress/abilities-api/pull/126/commits/aa88f6a7597a062b7545d1cc436dd448e234ce2d.

gziolo:
@jorgefilipecosta, good catch with a few inconsistencies in WP core. I will update them here before landing, and then patch in WP core.

gziolo:
I updated a patch against WP core to address the feedback first:
- https://github.com/WordPress/wordpress-develop/pull/10424

includes/abilities-api.php
felixarntz: I did provide similar feedback somewhere else (I believe on the Core PR) - I think allowing registration at any arbitrary point means we're being "forgiving" to developers doing it wrong, which can then lead to far more obscure problems down the line.

I think clearly defined APIs and, given WordPress's procedural nature, clearly defined registration timing are crucial to avoid more complex problems. Better tell the developer they should fix something rather than them running into a problem later that they have no idea on why it's occurring.

gziolo: I’m happy to apply further updates. I was a bit on the clock when landing the entire API in WP core. So while I agreed with reasoning, I missed this important nuance at the time.

It’s a tiny change for production logic, but needs some solid tinkering for unit tests so they are in proper state while registering abilities and ability categories. I’ll try to sort out something tomorrow, so it’s ready before beta 3.

gziolo: PR should be ready for review:
- https://github.com/WordPress/wordpress-develop/pull/10452

Code changes as expected are trivial, but getting the unit tests passing was quite a journey.

PR #128, opened 2025-10-23
Update: Simplify core abilties registration in preparation for core patch.

jorgefilipecosta:
This PR applies feedback provided in https://github.com/WordPress/abilities-api/pull/108, by @aaronjorbin as a preparation for the core patch.

## Testing
Verify automated testing pass.
Oppenened the editor, and the browser console, made some test requests:
```
wp.apiFetch({ path: wp.url.addQueryArgs('/wp/v2/abilities/core/get-site-info/run', { fields:[ 'name' ] } ) } ).then(console.log);
wp.apiFetch({ path: wp.url.addQueryArgs('/wp/v2/abilities/core/get-site-info/run', {input: { fields:[ 'name' ] } } ) } ).then(console.log);
```
Verified things return as expected.


includes/bootstrap.php
```diff
@@ -42,7 +42,7 @@
 }
 
 // Load core abilities registration functions.
-if ( ! function_exists( 'wp_register_core_abilities' ) ) {
+if ( ! function_exists( 'register_core_abilities' ) ) {
```
gziolo: For consistency,  with the file name and other functions, and filters, the `wp_` prefix might be worth keeping.

jorgefilipecosta: Updated 👍

gziolo:
It looks like we need to land https://github.com/WordPress/abilities-api/pull/126 first so CI job running against WP core `trunk` don't fail anymore.

jorgefilipecosta:
> It looks like we need to land #126 first so CI job running against WP core `trunk` don't fail anymore.

Reviewing https://github.com/WordPress/abilities-api/pull/126, I think this and https://github.com/WordPress/abilities-api/pull/126 should be merged soon so then I can follow up with further syncs from core.

gziolo:
This will require a rebase with `trunk` now. We might need a better mechanism that detects whether these ability categories and abilities haven't been registered earlier, as now that happens automatically in WP core `trunk`.

includes/bootstrap.php
```diff
@@ -41,17 +41,17 @@
...
+	require_once __DIR__ . '/abilities/wp-core-abilities.php';
 }
 
 // Register core abilities category and abilities when requested via filter or when not in test environment.
 // phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedHooknameFound -- Plugin-specific hook for feature plugin context.
 if ( ! ( defined( 'WP_RUN_CORE_TESTS' ) || defined( 'WP_TESTS_CONFIG_FILE_PATH' ) || ( function_exists( 'getenv' ) && false !== getenv( 'WP_PHPUNIT__DIR' ) ) ) || apply_filters( 'abilities_api_register_core_abilities', false ) ) {
 	if ( function_exists( 'add_action' ) ) {
-		add_action( 'wp_abilities_api_categories_init', array( 'WP_Core_Abilities', 'register_category' ) );
-		add_action( 'wp_abilities_api_init', array( 'WP_Core_Abilities', 'register' ) );
+		add_action( 'wp_abilities_api_categories_init', 'wp_register_core_ability_categories' );
```
gziolo, line 53: It might be possible to simplify all the conditions used with a simple (and remove PHPCS exception):
```php
if ( has_action( 'wp_abilities_api_categories_init', 'wp_register_core_ability_categories' ) {
```
The same for another action.

includes/abilities/wp-core-abilities.php
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/**
+ * Registers the default core abilities.
+ *
+ * @since 0.3.0
+ *
+ * @return void
+ */
+// phpcs:ignore WordPress.NamingConventions.PrefixAllGlobals.NonPrefixedFunctionFound
```
gziolo, line 45: This is no longer necessary. I think we could copy the latest version directly from WP core.

includes/bootstrap.php
jorgefilipecosta: Hi @gziolo, you mean just something like as top level code without any other condition?
if ( ! has_action( 'wp_abilities_api_categories_init', 'wp_register_core_ability_categories' ) ) {
	add_action( 'wp_abilities_api_categories_init', 'wp_register_core_ability_categories' );
}

includes/abilities/wp-core-abilities.php
jorgefilipecosta: I copied the last version directly from core as suggested 👍

includes/bootstrap.php
gziolo: It's the common pattern used in the Gutenberg plugin. The wrapping:

```
if ( function_exists( 'has_action' ) ) {
```

might still be necessary to ensure it doesn't break when loading files too early before WP core, which happens in some scenarios.

gziolo (approved):
There is one feedback to consider, but otherwise this looks great. Thank you for taking care of it!

jorgefilipecosta:
Thank you a lot for the review @gziolo, if required I will follow up with enhancements 👍

PR #137, opened 2025-11-06
Sync changes applied to WordPress 6.9 beta 3

gziolo:
This PR copies all the recent changes applied in WordPress core in preparation for the 6.9 beta 3 / rc 1 releases. 

Related commits:
- [Abilities API: Use stricter doing_action check when registering](https://github.com/WordPress/wordpress-develop/commit/fd41ccd8b34b5af2d14688039bfda2718d882d3b)
- [Abilities API: Improve annotations documentation and add permission callback validation](https://github.com/WordPress/wordpress-develop/commit/828a3fe5e225dd0c3206f5b7ec2f5b7b1d9afb4e)


includes/abilities-api/class-wp-ability.php
```diff
@@ -498,6 +525,14 @@ protected function invoke_callback( callable $callback, $input = null ) {
 	 * @return bool|WP_Error Whether the ability has the necessary permission.
 	 */
 	public function check_permissions( $input = null ) {
+		if ( ! is_callable( $this->permission_callback ) ) {
```
justlevine, line 528: Is this redundant/contrary to the behavior in `::prepare_properties()`? 

( I see this was added in https://github.com/WordPress/wordpress-develop/pull/10431/files, but I can't find the discussion that prompted it. )

IMO the fatal error thrown from `::invoke_callback()` is desired behavior, instead of hiding it inside a `WP_Error`. This error can only happen if a user is extending this class and overloading both the `::$permission_callback` property type and either `::prepare_properties()` or the constructor to avoid calling `::prepare_proerties()` altogether, so the immediacy provides better DX.

gziolo: It might be redundant, but it doesn't break anything. Feel free to open a patch against WordPress `trunk` proposing the removal of both checks. There is still time to adjust that logic before the final release.

As for this PR, as noted in the description, the code is copied from WordPress core as is, without modifications.

jonathanbossenger (approved):
From the inline documentation side, looks good to me

PR #142, opened 2025-11-12
Update: Json schema version 

jorgefilipecosta:
We are currently using draft-04 of json schema on the validator, but the schemas we are using required at least the 2020 version because of having examples and defaults.

## Testing
Currently the ajv validation only happens for the output of client side abiltiies not the server ones so this issue is hidden but it will be exposed when hybrid abilities are implemented.

Paste the following code on the browser console:
```
const envInfoAbility = await wp.abilities.getAbility(
	'core/get-environment-info'
);
const envInfoResult = await wp.abilities.executeAbility(
	'core/get-environment-info'
);
await wp.abilities.registerAbility( {
	...envInfoAbility,
	name: 'core/get-environment-info-client',
	callback: () => envInfoResult,
} );
await wp.abilities.executeAbility( 'core/get-environment-info-client' );
```
It registers a core/get-environment-info-client client ability exactly the same as the server one.
Verify things work well.
Paste the same code on trunk and verify there is this error and the ability does not works:
```
Schema compilation error: Error: strict mode: unknown keyword: "examples"
    at checkStrictMode (util.js:174:1)
    at checkUnknownRules (util.js:32:1)
    at alwaysValidSchema (util.js:19:1)
    at properties.js:23:1
    at Array.filter (<anonymous>)
    at Object.code (properties.js:23:1)
    at keywordCode (index.js:464:1)
    at index.js:222:1
    at CodeGen.code (index.js:439:1)
    at CodeGen.block (index.js:568:1)
```

### DIFF
--- a/packages/client/src/validation.ts
+++ b/packages/client/src/validation.ts
@@ -7,7 +7,7 @@
  *
  * @see https://developer.wordpress.org/rest-api/extending-the-rest-api/schema/#json-schema-basics
  */
-import Ajv from 'ajv-draft-04';
+import Ajv from 'ajv/dist/2020';
 import addFormats from 'ajv-formats';
 import type { ValidationError } from './types';
 
@@ -169,7 +169,7 @@ export function validateValueFromSchema(
 
 	try {
 		const validate = ajv.compile( args );
-		const valid = validate( value );
+		const valid = validate( value || {} );
 
 		if ( valid ) {
 			return true;


justlevine:
(Related discussion re JSON draft-4 https://github.com/WordPress/abilities-api/pull/109#pullrequestreview-3330388191 - for the cross-link)

gziolo:
It looks like `examples` property is ignored by the custom WordPress validator for JSON schema. I couldn't find any usage in WP core that would use that property. In effect, I would recommend removing that line of code rather than trying to use the latest JSON schema without aligning with what is supported on the server. 

The other aspect is that we could offer in the future to opt-in for the most recent JSON schema by explicitly defining the version. I don't think we even do it now, and set:

```php
'$schema'    => 'http://json-schema.org/draft-04/schema#',
```

That would be helpful to be explicit.

jorgefilipecosta:
Closed in favor of https://github.com/WordPress/abilities-api/pull/144 and https://github.com/WordPress/wordpress-develop/pull/10510.

PR #144, opened 2025-11-12
Fix: Input Validation handle top level default.

jorgefilipecosta:
Apparently draft-04 of JSON schema allows the top level default, that we are using, but ajv validator errors if a top level default is used, it also error if null or undefined is passed as input.

This PR fixes the issue, we remove the default from the compile schema and use it as the input when it is defined.

## Testing
On core switch to PR (which has required core changes to be draft-04 complient and work correctly, like removing examples and return JSON containing a default of {} instead of []).
Paste the following on the browser console:
```
const envInfoAbility = await wp.abilities.getAbility(
	'core/get-environment-info'
);
const envInfoResult = await wp.abilities.executeAbility(
	'core/get-environment-info'
);
await wp.abilities.registerAbility( {
	...envInfoAbility,
	name: 'core/get-environment-info-client',
	callback: () => envInfoResult,
} );


const siteInfoAbility = await wp.abilities.getAbility(
	'core/get-site-info'
);
const siteInfoResult = await wp.abilities.executeAbility(
	'core/get-site-info'
);
await wp.abilities.registerAbility( {
	...siteInfoAbility,
	name: 'core/get-site-info-cli',
	callback: () => siteInfoResult,
} );
[await wp.abilities.executeAbility( 'core/get-site-info-cli' ), await wp.abilities.executeAbility( 'core/get-environment-info-client' ) ]
```

Verify both abilities execute with success.

### DIFF
--- a/packages/client/src/api.ts
+++ b/packages/client/src/api.ts
@@ -324,7 +324,7 @@ async function executeServerAbility(
  */
 export async function executeAbility(
 	name: string,
-	input: AbilityInput
+	input?: AbilityInput
 ): Promise< AbilityOutput > {
 	const ability = await getAbility( name );
 	if ( ! ability ) {
--- a/packages/client/src/api.ts
+++ b/packages/client/src/api.ts
@@ -324,7 +324,7 @@ async function executeServerAbility(
  */
 export async function executeAbility(
 	name: string,
-	input: AbilityInput = null
+	input: AbilityInput
 ): Promise< AbilityOutput > {
 	const ability = await getAbility( name );
 	if ( ! ability ) {
--- a/packages/client/src/validation.ts
+++ b/packages/client/src/validation.ts
@@ -170,7 +170,11 @@ export function validateValueFromSchema(
 	try {
 		const { default: defaultValue, ...schemaWithoutDefault } = args;
 		const validate = ajv.compile( schemaWithoutDefault );
-		const valid = validate( value || defaultValue || {} );
+		let valueToValidate = value;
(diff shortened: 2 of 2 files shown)


packages/client/src/api.ts
```diff
@@ -324,7 +324,7 @@ async function executeServerAbility(
  */
 export async function executeAbility(
 	name: string,
-	input: AbilityInput = null
+	input: AbilityInput
```
gziolo: This would make the `input` mandatory while it isn't. 
```suggestion
	input?: AbilityInput
```

I'm not an TS expert, but it feels like this would allow the value to be undefined.

packages/client/src/validation.ts
```diff
@@ -168,8 +168,9 @@ export function validateValueFromSchema(
 	}
 
 	try {
-		const validate = ajv.compile( args );
-		const valid = validate( value );
+		const { default: defaultValue, ...schemaWithoutDefault } = args;
+		const validate = ajv.compile( schemaWithoutDefault );
+		const valid = validate( value === undefined ? defaultValue || {} : value );
```
gziolo: These schemas can have simple types like `string`, so I'm curious what was the decision process for using `{}`. Should it keep it undefined when the default value is not found? I think it's what happens on the server:

https://github.com/WordPress/wordpress-develop/blob/2740a8ed55af56ccf8c5a548dd393bdd3e8472c8/src/wp-includes/abilities-api/class-wp-ability.php#L448-L450

jorgefilipecosta: Good point I removed the empty object default.

gziolo (approved):

PR #150, opened 2025-11-19
Add filters for input and output validation

### LINKED ISSUE #149: Proposal: add filters for input and output validation
The Abilities API currently validates ability input and output against JSON Schema using [WordPress's built-in](https://developer.wordpress.org/reference/functions/rest_validate_value_from_schema/) `rest_validate_value_from_schema()`, which supports only a subset of JSON Schema Draft 4 (aligning with WordPress core). This approach is reliable for core compatibility but limits extenders to dated JSON Schema features, missing newer ones like `$ref` references for composability and reusability, and `not` keyword to exclude patterns.

Developers extending the Abilities API that need more expressive schema validation have no way to override the default validator without forking or monkey-patching.

## Proposed Solution
Introduce two hooks to allow custom validation:
```
/**
 * Filter to override input validation for an ability.
 * 
 * @param true|WP_Error $is_valid      Validation result (true or WP_Error)
 * @param mixed         $input         The input being validated
 * @param array         $input_schema  The input schema definition
 * @param string        $ability_name  The ability name
 * @param WP_Ability    $ability       The ability instance
 */
apply_filters( 'wp_ability_validate_input', $is_valid, $input, $input_schema, $ability_name, $ability );

/**
 * Filter to override output validation for an ability.
 * 
 * @param true|WP_Error $is_valid       Validation result (true or WP_Error)
 * @param mixed         $output         The output being validated
 * @param array         $output_schema  The output schema definition
 * @param string        $ability_name   The ability name
 * @param WP_Ability    $ability        The ability instance
 */
apply_filters( 'wp_ability_validate_output', $is_valid, $output, $output_schema, $ability_name, $ability );
````

### PR DESCRIPTION
**Pull request by priethor**
## What
Closes #149 by adding validation filters to the Abilities API to allow custom validation strategies.

## Why
Current validation uses only WordPress REST's JSON Schema subset (Draft 4), whereas extenders might need support for modern JSON Schema features.

## How
- Added `wp_ability_validate_input` filter in `WP_Ability::validate_input()`
- Added `wp_ability_validate_output` filter in `WP_Ability::validate_output()`
- Both filters run AFTER default WordPress validation, allowing overrides
- Filters receive validation result, data, ability name, and ability instance
- Backward compatible—default behavior unchanged if no filters attached

## Testing Instructions
1. Register an ability with `input_schema` and `output_schema`
2. Add a filter to `wp_ability_validate_input` to use custom validation
3. Execute the ability with test data
4. Verify that the custom validator is called and validates correctly

Alternatively, run the new seven tests against `trunk` and verify they fail.

### DIFF
--- a/includes/abilities-api/class-wp-ability.php
+++ b/includes/abilities-api/class-wp-ability.php
@@ -479,7 +479,7 @@ class WP_Ability {
 
 		$valid_input = rest_validate_value_from_schema( $input, $input_schema, 'input' );
 		if ( is_wp_error( $valid_input ) ) {
-			return new WP_Error(
+			$is_valid = new WP_Error(
 				'ability_invalid_input',
 				sprintf(
 					/* translators: %1$s ability name, %2$s error message. */
@@ -488,9 +488,26 @@ class WP_Ability {
 					$valid_input->get_error_message()
 				)
 			);
+		} else {
+			$is_valid = true;
 		}
 
-		return true;
+		/**
+		 * Filters the input validation result for an ability.
+		 *
+		 * Allows developers to add custom validation logic on top of the default JSON Schema validation.
+		 * If the default validation already failed, the filter receives the WP_Error object and can
+		 * add additional error information or override it. If the default validation passed, the filter
+		 * can add additional validation checks and return a WP_Error if those checks fail.
+		 *
+		 * @since 7.0.0
+		 *
+		 * @param true|WP_Error $is_valid      The validation result from default validation.
+		 * @param mixed         $input         The input data being validated.
+		 * @param string        $ability_name  The name of the ability.
+		 * @param WP_Ability    $ability       The ability instance.
+		 */
+		return apply_filters( 'wp_ability_validate_input', $is_valid, $input, $this->name, $this );
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: includes/abilities-api/class-wp-ability.php [hunk 1]
```diff
@@ -488,9 +488,26 @@ public function validate_input( $input = null ) {
...
+		 * If the default validation already failed, the filter receives the WP_Error object and can
+		 * add additional error information or override it. If the default validation passed, the filter
+		 * can add additional validation checks and return a WP_Error if those checks fail.
+		 *
+		 * @since 7.0.0
+		 *
+		 * @param true|WP_Error $is_valid      The validation result from default validation.
+		 * @param mixed         $input         The input data being validated.
+		 * @param string        $ability_name  The name of the ability.
+		 * @param WP_Ability    $ability       The ability instance.
```
justlevine: I recommend we don't pass the class instance to the filter, so as to safeguard it's lifecycle/state. This would also align with the existing class actions (which IIRC were designed around the same consideration).

If someone (who knows what they're doing) wants to call  the class methods out of order, the intentionality of an `$ability = wp_get_ability( $ability_name );`first is IMO a desirable source of friction.

justlevine (commented):
thanks so much for this PR @priethor , 🙇‍♂️ I love the use case!

Only reviewed from mobile (I'm sick ATM) and had
a quick question:

> Both filters run AFTER default WordPress validation, allowing overrides

Can you share your reasoning here?
If the goal is to BYO validation, I would assume we'd want the filter to be a precheck and return early instead of running validation you know you're throwing out in favor of another parser/validation. I personally don't feel strongly either way, but since it was intentional enough for you to call out, I am curious.

(FWIW since it's a public method, a bypass hook could still be used to override/recover from a `WP_Error`, just with admittedly not too great ergonomics. Ironically my unrelated comment below recommends making those ergonomics slightly worse 😅)

includes/abilities-api/class-wp-ability.php
priethor: Good call, fixed it in 618d161. In most cases, the ability name should be enough to determine whether to alter the validation or not.

priethor:
Thanks, @justlevine !

> Can you share your reasoning here?

I'm glad you asked, since I gave this a few thoughts. I considered adding a pre-validation filter to short-circuit it entirely, similar to how the REST works, but I think there are cases where extenders might want to extend the validation output rather than completely replace it.

For example, I could see a case where extenders want to apply the core validation to verify the input structure and, when it succeeds, apply an extra dynamic filter (e.g., checking an API) to detect forbidden words.

I don't feel very strongly either way, but it seemed to me that the potential computational cost of applying the core validation first is worth the tradeoff for the flexibility it provides. Another option that could make sense is to provide both pre- and post-filters.

priethor:
Trac ticket, since core is the source of truth: [core.trac.wordpress.org/ticket/64311](https://core.trac.wordpress.org/ticket/64311)

PR #157, opened 2025-12-30
Fix incomplete translator comments in WP_Ability and WP_Ability_Category

igorradovanov:
The translator comments in `WP_Ability` and `WP_Ability_Category` only describe the first placeholder but the strings have three.

### WP_Ability (`class-wp-ability.php`)

Before:
```php
/* translators: %s: Property name. */
```

After:
```php
/* translators: %1$s: Property name, %2$s: Ability name, %3$s: Class name. */
```

### WP_Ability_Category (`class-wp-ability-category.php`)

Before:
```php
/* translators: %s: Property name. */
```

After:
```php
/* translators: %1$s: Property name, %2$s: Ability category slug, %3$s: Class name. */
```

Translators using tools like translate.wordpress.org don't see the source code, so having all placeholders described helps them know what each one is.

### DIFF
--- a/includes/abilities-api/class-wp-ability-category.php
+++ b/includes/abilities-api/class-wp-ability-category.php
@@ -88,7 +88,7 @@ final class WP_Ability_Category {
 				_doing_it_wrong(
 					__METHOD__,
 					sprintf(
-						/* translators: %s: Property name. */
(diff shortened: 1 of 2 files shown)


justlevine (approved):

justlevine:
(Clarifying - approving as this is indeed a docs bug + valid fix but I believe it needs to first be directly applied to `wordpress-develop` ).

